PR #12, opened 2015-05-21
Null constructors and global config check

### LINKED ISSUE #11: Empty Constructor Not Called
The empty constructor is not called, so tools are not set to nullptr.  If they are not instantiated in the code, then the job will crash on finalize() when these are deleted.

kratsg:
This resolves #11.

Global variables:
- m_name
- m_debug
- m_event
- m_store
- m_systName
- m_systVal

Protected variables:
- m_configName


jdandoy:
This update breaks previous versions as it forces the use of only a NULL constructor.  Can we continue to include the previous constructors for the time being?

gfacini:
Girodon,
Great stuff!  If we are going to break everything then we need to 
1. Make a stable tag people can use until they are ready to make the switch
2. Give a week warning before we make the switch in the trunk or whatever git calls it for those working on the cutting edge.  If not a week, at least a few days
Gabriel

kratsg:
@jdandoy - good point. I'll grab the regular constructor and add it back in. I can update the pull request soon. Maybe tomorrow.

@gfacini - sure. I can make the regular constructor print a warning so people know to switch.

kratsg:
Actually, @gfacini - it's not a huge change for folks. EG:

```
MuonCalibrator* muonCalib                     = new MuonCalibrator(       "muonCalib",                localDataDir+"muonCalib.config");
```

is replaced with

``` 
MuonCalibrator* muonCalib                     = new MuonCalibrator();
muonCalib->setName("muonCalib")->setConfig(localDataDir+"muonCalib.config");
```

we can make a release note telling people how to switch from 00-00-04 to 00-00-05.

Issue #13, opened 2015-05-21
Electron* revamping for Release 20


mmilesi:
I am currently working in:

-) Checking that the Electron* modules are up-to-date with 2.3.12+ ASG releases, and in case not, update them. One known issue is the major interface change for the Isolation Tool (now usable for both Electrons and Muons) .

-) For the ElectronEfficiencyCorrector module, decorate electrons with a vector<double> with nominal and systematically varied SFs, as it is done by @johnda102  in BTaggingSF module.

PR #14, opened 2015-05-22
Electron modules revamping for mc15

mmilesi:
-) Updated electron isolation tool interface to cope with 2.3.X CP tool version.
-) In ElectronEfficiencyCorrector, decorate with vector<double>  of SFs (nominal+syst) each electron
-) Create ParticlePIDManager.h, containing classes that manage multiple instances of CP tools for Particle ID (for now only electron LH tool). An instance of such classes becomes a data member of the Selectors, and allows particle selection for a given WP plus automatically provides decorations for WPs equal or tighter to the one selected. 
-) Using such decorations, electrons ID (and isolation) branches are now available from HelpTreeBase using switches.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -258,7 +258,7 @@ EL::StatusCode BJetEfficiencyCorrector :: execute ()
     }
 
     //
-    // Creat the name of the weight
+    // Create the name of the weight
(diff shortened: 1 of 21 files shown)


kratsg:
@mmilesi , run `git checkout master && git pull --rebase && git branch -d devbranch_mmilesi` to clean up.

kratsg:
Does this close #13 ?

PR #17, opened 2015-05-23
Remove dependency-related files.

kratsg:
This would close the milestone. (do it manually).


kratsg:
Remove scripts and diffs for 00-00-05 as we should release on 2.3.12 which works.

mmilesi:
Hi @kratsg ,
2.3.12 still needs to checkout a specific tag of ElectronEfficiencyCorrector, so we need the 
checkoutASGtags.py

In any case, I agree to remove the .sh and .diff, but I would leave the python script, as it is harmless and it is clear to everyone that 00-00-05 will be released for 2.3.12 (or 2.3.13), so nobody will ever try to use it. Also...what if we find other bugs of the ASG tools in the meantime?

kratsg:
Good point. The python script is remaining.

mmilesi:
Close #17

PR #23, opened 2015-05-26
Remove Unnecessary Printed Info with Debug String

jdandoy:
A debug string added to JetHists so less information is printed.  When making many JetHists for all the systematic variations, the Info printed is too much.  Also re-removing User functions from JetHists and HistogramManager.


kratsg:
This looks great! (thanks for doing the pull-request).

PR #25, opened 2015-05-26
ConfigName for GRID jobs.

kratsg:
Do not expand the configName for grid jobs. Add a getConfig(expand=false) option to allow folks to grab the configuration filename if they need it.

/cc James Saxon reported this issue and this is a commit to fix it up. I'll merge when Jamie confirms it works.


kratsg:
Jamie is running the job here: http://bigpanda.cern.ch/task/5535047/ . We'll see how it goes.

kratsg:
Jobs finished successfully!

PR #26, opened 2015-05-26
Remove histInitialize, Added jet cleaning decisions

jdandoy:
Everything moved out of histInitialize and into initialize to allow for m_isMC and other meta-data checking.  Nothing should be used in histInitialize from here on out.  Also added option to save all cleaning decisions into the output tree.

/cc @gfacini @mmilesi @kratsg

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -141,74 +141,7 @@ EL::StatusCode BasicEventSelection :: histInitialize ()
 
   // TODO why is histInitialize() called after fileExecute() ??
 
-  if ( this->configure() == EL::StatusCode::FAILURE ) {
-    Error("initialize()", "Failed to properly configure. Exiting." );
-    return EL::StatusCode::FAILURE;
-  }
-
   Info("histInitialize()", "Calling histInitialize");
-
-  // write the metadata hist to this file so algos downstream can pick up the pointer
-  TFile *fileMD = wk()->getOutputFile ("metadata");
-  fileMD->cd();
-  // event counts from meta data
-  m_histEventCount = new TH1D("MetaData_EventCount", "MetaData_EventCount", 6, 0.5, 6.5);
-  m_histEventCount -> GetXaxis() -> SetBinLabel(1, "nEvents initial");
-  m_histEventCount -> GetXaxis() -> SetBinLabel(2, "nEvents selected in");
-  m_histEventCount -> GetXaxis() -> SetBinLabel(3, "nEvents selected out");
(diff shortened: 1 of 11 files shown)


kratsg:
@jdandoy `histInitialize` is used to start histograms before the start of a file load. If you run over multiple datasets, then initialize() is called per file while histInitialize() is called per sample. Are we sure we want to move this all over?

mmilesi:
Hi @jdandoy , I am currently working on BasicEventSelection to sort out the CutBookkepers thing. I could be very sensitive to those changes, so I will wait a bit before merging this pull request.

gfacini:
@kratsg 
we call configure in initialize so we really reconfigure every algo for every file?

jdandoy:
@kratsg I don't think that's correct.  The skeleton files say:

Here you do everything that you need to do after the first input
  // file has been connected and before the first event is processed,
  // e.g. create additional histograms based on which variables are
  // available in the input files. 

 Also configure was already called in histInitialize in many place

mmilesi:
Hi @jdandoy 

Can you please explain why do we need something like this:

 // label the bins for the cutflow
+  m_cutflow_all  = m_cutflowHist->GetXaxis()->FindBin("all");
+  m_cutflowHistW->GetXaxis()->FindBin("all");
+  if(!m_isMC && m_applyGRL){
+    m_cutflow_grl  = m_cutflowHist->GetXaxis()->FindBin("GRL");
+    m_cutflowHistW->GetXaxis()->FindBin("GRL");
+  }

Do we really need to fork btwn data and MC? Why not just filling up with a "dummy" value the cutflow bin if the cut does not correspond to the data type under question?

But maybe I am overlooking something...

Marco

jdandoy:
Hi @mmilesi ,

I don't think we should be including cutflow selections for selections that don't happen.  This way it is obvious which selections you acutally called and their actual effect.  It also serves as a good double check, i.e. are you sure you turned of ApplyGRL?  Then it's not in the cutflow.  It is also nicer when presenting cutflow results to only have the selections you used.  For a few extra lines of code I think the functionality is worth it.

gfacini:
I tend to agree with Jeff on this but am interested to hear objections if you disagree @mmilesi

gfacini:
I made the default values -999 as other entries in the tree. Otherwise seems fine to me

jdandoy:
This is a plus one message.

+1

gfacini:
this is a question about what is a plus one message
what is a plus one message?

gfacini:
:+1:

kratsg:
I will merge, @mmilesi will handle his end with the Metadata tomorrow (CERN).

PR #28, opened 2015-05-26
Calo clusters in tree

gfacini:
Added back rapidity since somehow that was removed - now with a switch
Add PV selection for ghost associated tracks - taken for JVF/JVT TWiki (noted in code)
Adding calo clusters with a 2 GeV cut


gfacini:
Will merger after Jeff's latest changes go in concerning jet cleaning branches

PR #33, opened 2015-05-27
Read Cutbookkepers in BaseEventSelector

mmilesi:


kratsg:
We discussed this yesterday over gchat, this looks great. :gun:

PR #37, opened 2015-05-29
Trigger information

gfacini:
BasicEventSelection is the only place the trigger tools are used.
Information can be saved there to EventInfo to be used downstream.
This information can also make its way into the tree via HelpTreeBase Trigger functions.


mmilesi:
Looks good! Merge PR #37

gfacini:
Thanks!

PR #68, opened 2015-05-30
This closes #67.

### LINKED ISSUE #67: Get rid of ContainerRecording examples.
These should probably be deleted now. We don't test them as xAODRootAccess has updated to fix these issues.

kratsg:
Remove container recording examples. Clean up the Makefile so it looks nicer.


kratsg:
This is a trivial PR.

PR #70, opened 2015-05-31
Streamablizing it all

### LINKED ISSUE #32: Allowing for algorithms to be configurable from python
So we need to start moving away from our reliance on a TEnv config file. `xAH_run.py` will be incredibly useful and flexible if we can start moving towards that, but without stopping people from being old-fashioned and using TEnv config files as usual.

To do this, we need to get rid of the "//!" options in front of anything that is a public member variable that is meant to be configurable. This tells ROOT that the variable is streamable and configurable. This should also be set to a default value in the class constructor initialization.

kratsg:
This will close #32 when all is done.


mmilesi:
Everything looks fine. PR merged. Close #70

PR #73, opened 2015-05-31
Streamablize it all.

### LINKED ISSUE #32: Allowing for algorithms to be configurable from python

kratsg:
This closes #32. All algorithms with public members are meant to be configurable from python, a macro script, or a TEnv config file. This requires that all members are streamable, but also set to default values

Note that the `test_multiAlgo` does not work with the skimmed mc15 file provided. Need to run something like

```
test_multiAlgo submitDir /share/t3data3/kratsg/xAODs/ mc15_13TeV.410000.PowhegPythiaEvtGen_P2012_ttbar_hdamp172p5_nonallhad.merge.AOD.e3698_s2608_s2183_r6630_r6264_tid05419191_00/ AOD.05419191._000003.pool.root.1
```

where you provide your own file.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -44,6 +44,27 @@ BJetEfficiencyCorrector :: BJetEfficiencyCorrector () :
 
   Info("BJetEfficiencyCorrector()", "Calling constructor");
 
+  // read flags set from .config file
+  m_debug                   = false;
+  m_inContainerName         = "";
+  m_systName                = "";      // default: no syst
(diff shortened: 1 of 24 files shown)


mmilesi:
Merge #73

PR #79, opened 2015-06-02
flag to skip metadata stuff

gfacini:
There is an issue with MetaData corruption and CutBookkeepers in the Derivation Framework related to the move to release 20.1.5.  This flag is a temporary thing for those running on these derivations.


mmilesi:
Merge #79

PR #85, opened 2015-06-04
Update b jets

### LINKED ISSUE #74: make BTag cut configurable, and add decorations of btagging WP for ALL jets
I simply attach an old email thread we had.

1. At the moment, i see that if we specify in the configuration a certain Btag WP , say XXX,  (only for MV1 atm, which btw is not available in Dc14 as you most likely know...), then the selection decoration for a jet will become:

  ```
  passSelBTagXXX
  ```

  Wouldn't be better to have 2 separate flags: a 

  ```
  passSel
  ```

  flag, regardless of whether the jet is also b-tagged or not, and an additional flag

  ```
  passSelBTagXXX (or isBTagXXX)
  ```

  to identify jets that pass also the btag requirement?

  This is quite crucial: in fact, when the Overlap Removal tool is configured to look only at (pre)"selected" objects - as I am doing in my analysis - it assumes all objects: 

  - electrons
  - muons
  - jets
  - (taus)
  - (photons)

  share the same flag name for selected objects, i.e., `passSel`. If the jets have a different name for this flag, then it messes up the OR.

  > OK I can make it use `passSel` and `passSelBTagXXX`
 

2. Do we really wan to have a CUT on a jet if it doesn't pass a btag requirement ( as it is done now in the `PassCuts()` method )? Wouldn't the decoration just suffice?
  > At least, I will keep this cut configurable by the user.

johnda102:
Would like pull request for UpdateBJets:
  -) Moves baseline tagger to MV2c20 in JetSelector
  -) Applies btagging cuts if requested with DoBTagCut,
  -) Applies BTagXXX decoration in addition to passSel decoration for all jets if DoBTagCut is not set and only for passing jets if DoBTagCut is set.
  -) Move to Pre-Pre-Recomendation CDI file (inculded in xAH/data) 
  -) Update BJetEffCorr to use MV2c20 and new CDI file
  -) tested with multialgo example

closes #74


mmilesi:
Hi @johnda102 ,

looks good. Thanks a lot! So as for my understanding you suggest making multiple instances of the tool with different WPs if one wants to make separate jet containers. Can you confirm?

johnda102:
@mmilesi 

  Yes if you want separate containers with different WPs can you simply instantiate multiple JetSelectors 
   If you only care about the decorations and are not cutting on the btagging they will all be set. 

 There are a number of obvious improvements that we should put in later: 
     -) Tag veto selection
     -) Support for 0.3 track jets

but I guess we can add these at a later stage.

mmilesi:
Ok, I just wanted you to confirm what I guessed.
Merge #85

PR #87, opened 2015-06-05
protect trigger for truth only

gfacini:


gfacini:
ahh I guess this is okay to merge.

PR #89, opened 2015-06-08
Skeleton for MET recalculation, and order change in muon/electron selection

JamesSaxon:
This commit includes two distinct pieces: 

 1) METConstructor added to xAODAnaHelpers.  This code follows closely from [METMakerAlg.cxx](https://svnweb.cern.ch/trac/atlasoff/browser/Reconstruction/MET/METUtilities/trunk/src/METMakerAlg.cxx).  A test for it has been added in test_multiAlgo, with an example configuration in data/.  With Debug set to true, the stored MET values are printed and compared to the recalculated values and, for the most part, these agree well.  Further work is required to propagate the object systematics through this calculation.  The soft term uncertainty recommendations have not yet been released.

 2) I have changed the order of the muon and electron trackParticle access, so that it comes _after_ the pt and eta cuts.  This makes the cuts more compatible with derivations, where those collections may be slimmed, depending on the particles' pts and etas.  Since in those cases the track particles may not be available, this leads to a seg fault if the electrons or muons are not rejected before attempting access.  Hope this makes sense.

Thanks a lot for your consideration -- 

Jamie

### DIFF
--- a/Root/ElectronSelector.cxx
+++ b/Root/ElectronSelector.cxx
@@ -655,13 +655,6 @@ int ElectronSelector :: passCuts( const xAOD::Electron* electron, const xAOD::Ve
 
   int oq      = static_cast<int>( electron->auxdata<uint32_t>("OQ") & 1446 );
 
-  // https://twiki.cern.ch/twiki/bin/view/AtlasProtected/InDetTrackingDC14
-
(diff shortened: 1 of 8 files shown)


kratsg:
All good!

PR #90, opened 2015-06-09
Adding protection if trigger info not availible

johnda102:
Adding protection for samples without trigger information. 
(Current blackhole samples are truth-level with not TriggerMenu saved)


gfacini:
Yep - a dumb mistake on my part putting that outside the protection!
Thanks for catching this.

Issue #98, opened 2015-06-12
crash for missing track parrticle


JamesSaxon:
In the file listed below, I found an electron no track particle and pt = -nan.  This causes ElectronSelector.cxx to fail, because it can't find the tp->d0().

According to the official standard, I think nan should fail any comparison, so this should have been blocked from even reaching the track parameter checks.  But it's not (and this standard is apparently seldom respected).  I looked at this event, and it is strange, but the caloCluster() pt is nearly 0 so I think it is reasonable to toss that electron.  I would propose printing a warning, and doing the following before the pt comparison:

  if (std::isnan(et)) {
    Info("PassCuts()", "Electron pt is NaN: calo pt=%f", electron->caloCluster()->pt() );
    return 0;
  }

Alternatively, we could do a check on whether the trackParticle is found, before using it.

Could we include this?

Thanks -- 

Jamie

mc15_13TeV.301399.PowhegPythia8EvtGen_CT10_AZNLOCTEQ6L1_VBFH125_ZZ4nu.recon.AOD.e3909_s2608_s2183_r6630_tid05530075_00/AOD.05530075._000023.pool.root.1

PR #99, opened 2015-06-12
protection against missing track particles

JamesSaxon:
Protection against missing track particles, as discussed with Marco and Giordon.


kratsg:
This will close #98 when it is merged.

PR #101, opened 2015-06-13
Allowing the TrigDecTool to be used Downstream

johnda102:


kratsg:
Hi @johnda102 I don't think this is needed to explicitly put the tool in. When you initialize a new tool, it gets put into the toolstore. Downstream, you should just be able to retrieve it again.

johnda102:
So it is.

PR #103, opened 2015-06-15
Updated jet config

jdandoy:
JetCalibration config file for 2.3.12+

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -402,11 +402,11 @@ EL::StatusCode BasicEventSelection :: initialize ()
 
 
   // Trigger //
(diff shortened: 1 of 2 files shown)


mmilesi:
Hi @jdandoy ,

thanks a lot for this!

Just one question:

# MC 13 TeV
-CalibSequence		JetArea_Residual_EtaJES
+CalibSequence     	JetArea_Residual_Origin_EtaJES_GSC
 # Data
 #CalibSequence		JetArea_Residual_Origin_EtaJES_GSC_Insitu

Is CalibSequence the same for data and mc15?

jdandoy:
From the user's point of view the CalibSequence is the same in the config file.

In general the CalibSequence from data will be the same as MC, but with "_Insitu" appended.  To keep things simple for the user the Insitu is appended if they're running on Data and "Insitu" isn't already in the name.  Similarly, if "Insitu" is found in the name but you're running on MC, the code will send an error and exit.  You can use this config file safely for both MC and Data.

mmilesi:
Excellent, thanks for reminding. Merging #103

PR #106, opened 2015-06-16
Fix 'passSel' flag for jets in JetSelector.cxx

mmilesi:
-) In Root/HelpTreeBase.cxx, added a control on SV0_significance3D variable to check whether it's available before filling the tree.

-) In Root/JetSelector.cxx, make sure that the decoration for jets that pass the selection is 'passSel' ONLY (no appended strings).
   Jets that pass BTagging requirements will be decorated anyways later in the algorithm.
   This is needed to prevent crashes at the Overlap Removal stage.


mmilesi:
Hi @johnda102 ,

this PR should fix the issue with the wrong name for 'passSel' flag for jets in the overlap remover, plus one minor thing in the tree module that we already discussed. 
Please have a careful look at my changes ( it definitely works for me, but I don't want to screw up anything you might need for bjet studies! ), and then go ahead with the merging.

Thanks!

Marco

PR #111, opened 2015-06-17
Jet uncert and jvt

jdandoy:
Update to use new JESUncertainty Tool and JVT recalculation.  Will be included in a new tag for the Dijet analyses.

@gfacini


kratsg:
This looks ok on my end.

PR #113, opened 2015-06-17
Change eventWeight to mcEventWeight for correct MC weighting

lazovich:
BasicEventSelection sets the MC weight as "mcEventWeight", but subsequent stages access the property as "eventWeight". Fix this.


gfacini:
I do not think the histogram classes have been widely used so thank you for checking this.
Really the weight should be configurable incase one combines weights to make a total
weight with an arbitrary name.

Gabriel

PR #114, opened 2015-06-17
Make jet cleaning more robust to failures

lazovich:
The trimmed large-R jets (at least in EXOT8 derivation) don't have jet cleaning info saved. JetCleaningTool gives a segfault when this is the case. I have added code to try using the default and if it fails, catch the error and then try to clean the "Parent" jet (present for trimmed jets) if one is available. 

Also, Jet/Etmiss recommendation for AntiKt10LCTopoTrimmedPtFrac5SmallR20Jets does not include _Insitu in data. Currently the framework is appending that automatically to the calibration sequence if it is running over data. I've added a check so that it doesn't do it for the trimmed collection, because otherwise a segfault happens.


kratsg:
Hi @lazovich , thanks! A few suggestions since we cannot merge this into our code until some things are fixed (predominantly style-rules):

http://acode-browser.usatlas.bnl.gov/lxr/source/atlas/Event/xAOD/xAODJet/xAODJet/versions/Jet_v1.h#0254

- Please use `getAssociatedObjects` instead.
- Do not use `dynamic_cast` at all (`static_cast` is ok for primitive types in some cases)
- Try not to put a `try-catch` block in the code

Here's an example of a more robust code (I suggest not adding too many functions to `EL::Algorithm`) that can be placed inline.

```c++

  if(m_cleanParent){
    std::vector<xAOD::Jet*> parentJet;
    if(!jet_itr->getAssociatedObjects("Parent", parent)){
      Error("jetDecision()", "Could not make jet cleaning decision on the parent! It doesn't exist.");
    } else {
      isCleanDecor( *jet_itr ) = m_jetCleaning->accept( *(parent.at(0)) );
    }
  } else {
    isCleanDecor( *jet_itr ) = m_jetCleaning->accept( *(parent.at(0)) );
  }
```

lazovich:
Thanks for the feedback, @kratsg! I just committed a change to the code that gets rid of the function and adds the option. I tried to use the getAssociatedObject as you suggested, but ended up constantly getting runtime errors like   

` what():  SG::ExcAuxTypeMismatch: Type mismatch for aux variable '::Parent' (197); old type is 
ElementLink<DataVector<xAOD::Jet_v1, DataVector<xAOD::IParticle, DataModel_detail::NoBase> > > > new type is ElementLink<DataVector<xAOD::IParticle, DataModel_detail::NoBase> > `

even when experimenting with different possible types for the variable (besides the obvious xAOD::Jet). So, for now, I've used a working implementation with getAttribute. I know it's probably not ideal, so if you have any more recommentations please let me know!

Thanks,
Tomo

kratsg:
hi @lazovich -- we really don't want to static cast this. 

```
std::vector< const xAOD::Jet* > retrievedJets;
bool r = jet->getAssociatedObjects( "Parent", retrievedJets);
```

Are you sure the above does not work for you? It compiles for me. Also

```
const xAOD::Jet* parentJet = static_cast<const xAOD::Jet*>(*(jetToClean->getAttribute<ElementLink < xAOD::JetContainer > > ("Parent")));
```

is definitely not correct. You shouldn't be static casting a JetContainer to a Jet at all.

kratsg:
Also, looking at that error message -- it seems like the jets you run over do not have their parents set correctly. I would definitely ask on PATHelp about that. That's a run-time error associated with the file you're using, not with the code itself:

> ```
what(): SG::ExcAuxTypeMismatch: Type mismatch for aux variable '::Parent' (197); old type is ElementLink<DataVector<xAOD::Jet_v1, DataVector<xAOD::IParticle, DataModel_detail::NoBase> > > > new type is ElementLink<DataVector<xAOD::IParticle, DataModel_detail::NoBase> >
```

**Comment by kratsg** [comment]
Ahh, looks like I can confirm the error on my end:

```
terminate called after throwing an instance of 'SG::ExcAuxTypeMismatch'
  what():  SG::ExcAuxTypeMismatch: Type mismatch for aux variable `::Parent' (231); old type is ElementLink<DataVector<xAOD::Jet_v1, DataVector<xAOD::IParticle, DataModel_detail::NoBase> > > new type is std::vector<ElementLink<DataVector<xAOD::IParticle, DataModel_detail::NoBase> >, std::allocator<ElementLink<DataVector<xAOD::IParticle, DataModel_detail::NoBase> > > >
Aborted
```

This is actually really strange. Let me dump the file and get back to you.

**Comment by lazovich** [comment]
hi @kratsg, thanks for the reply. If it helps clarify at all, the code that sets the parent attribute is [here](http://acode-browser.usatlas.bnl.gov/lxr/source/atlas/Reconstruction/Jet/JetRec/Root/JetFromPseudojet.cxx#0218). It looks like they do something like this:

`ElementLink<JetContainer> el(*pcon, pparent->index());
pjet->setAttribute("Parent", el); `

So, the ElementLink has type JetContainer but it is actually a link to a specific jet within that container (the parent jet), which is specified by the second argument (the index). This is a really confusing way to do things but it is the way it is currently done. 

I don't know if this gives you any more insight on exactly how to access the parent using getAssociatedObject, but if it does please let me know. 

Thanks again!

**Comment by lazovich** [comment]
Hi again @kratsg...I actually just confirmed that the static cast is not necessary at all. If you simply do 

` const xAOD::Jet* parentJet = *(jetToClean->getAttribute<ElementLink < xAOD::JetContainer > > ("Parent")) `

the code compiles and runs successfully. So, I think this confirms that the ElementLink< xAOD::JetContainer> is actually a pointer to an xAOD::Jet* (i.e. a link to an element of the JetContainer). Not the most intuitively named but the types work out in the end.

**Comment by kratsg** [comment]
So I see here: https://gist.github.com/kratsg/a27b3fff266de0b88ec3#file-info-dump-L90

```
  |	  |	&->getAttribute<ElementLink<DataVector<xAOD::Jet> >>("Parent")
```

which looks like a better way may be... `getAssociatedObject` presumably because it expects a specific type. In fact, P-A seems to have hard-coded it so I need to email him. You also need to make sure you can retrieve it by protecting the code.

```
const ElementLink<xAOD::JetContainer> parentJetsLink;
if(!jetToClean->getAttribute("Parent", parentJetsLink)){
  Error("Could not find parent jet...");
} else if(!parentJetsLink.isValid()){
  Error("The link isn't valid! Must have been slimmed or thinned...");
} else {
  const xAOD::Jet* parent = parentJetsLink->getDataPtr()->at(0);
}
```

I need to email P-A now because this is broken and he's commented it out for some reason.

**Comment by kratsg** [comment]
I've emailed PATHelp, the current code you have "happens" to work only because dereferencing a pointer to an array of elements gives you the first element of the array. You should really be doing `foo->at(0)` here, but it's not even clear why the parent is being set as a JetContainer in the first place since there's no good reason.

**Comment by lazovich** [comment]
@kratsg : the * operator in this case, when applied to the ElementLink, is not actually accessing the first element of the JetContainer. They have overloaded the operator:

http://acode-browser.usatlas.bnl.gov/lxr/source/atlas/Control/AthLinks/AthLinks/ElementLink.icc#0262

It accesses the index of the JetContainer that the ElementLink was initialized with. So, doing at(0) as you suggest would not access the correct jet.

**Comment by kratsg** [comment]
So it turns out that it's even more convoluted! http://acode-browser.usatlas.bnl.gov/lxr/source/atlas/Control/AthLinks/AthLinks/ElementLink.icc#0230

`getDataPtr()` returns the raw pointer associated with the object you store which is a combination of the original container+key (index or element).

`cptr()` is a function that grabs `getDataPtr()` then applies that original container and key to dereference it for you and returns a pointer to it.

This is still very, very convoluted low-level code that I'm not sure we want to include. I'll wait for P-A to respond with his piece and we'll apply his suggestion. I'm still not sure why `setAssociatedObject` is not used at all. :cry:

**Comment by lazovich** [comment]
Ok, thanks for your help! Sorry for all the noise..tracking this down through the code has not been easy :)

**Comment by kratsg** [comment]
It's not noise! Much better than emails to be honest. This is still a feature we need, implementing it correctly is not trivial apparently.

**Comment by kratsg** [comment]
Here's P-A's reply, please update your code accordingly :)

> Hi Giordon,
>
> 'getAssociatedObject() ' is the recommended way to access particles (or
other objects such as MuonSegment) associated to jets.
> 
> The convention for jets is associated particles are internally stored as
IParticle and that getAssociatedObject() will know what to retrieve and
cast into the requested type (Jet, Muon, Tracks, ...)
> 
> However there's an inconsistency in JetFromPseudojet which breaks the
convention. As a result the recommended function fails.
> 
> A workaround should be to use low level code to access the parent :
> 
> ```
// untested but something like this should work :
ElementLink<JetContainer> & el = jet->auxdata<ElementLink<JetContainer>
 >("Parent") ;
id(el.isValid() ) const xAOD::Jet* jet = *el;
```
> I'll open a JIRA ticket to follow up on this issue.
> 
> Cheers,
> 
> P-A

kratsg:
Here is the JIRA issue: https://its.cern.ch/jira/browse/ATLJETMET-403 .

lazovich:
@kratsg thanks! I've fixed the merge conflicts and switched to P-A's recommended way of accessing. Let me know if there's anything else needed.

Tomo

kratsg:
:+1:  :gun:  :horse: :hospital:

Issue #102, opened 2015-06-15
Upgrade from 2.3.13 to 2.3.14 (w/ pre-recommendations) - Electron* and Muon* stuff 


mmilesi:
In 2.3.14 we are having some ASG tool interface changes:

See here: https://twiki.cern.ch/twiki/bin/view/AtlasProtected/AnalysisBase#Run2_EDM_xAOD_Rel20_i_e_mc15_and

kratsg:
Why do they keep constantly changing? Ugh.

mmilesi:
Closed through #118

PR #118, opened 2015-06-19
Egamma and Muon pre-recommendations AND Muon Trigger Matching

### LINKED ISSUE #110: Perform Trigger Matching for Muons
See [TWiki](https://twiki.cern.ch/twiki/bin/view/AtlasProtected/MCPAnalysisGuidelinesMC15#Muon_trigger_matching).

mmilesi:
This is an important update for muons and electrons.
All the changes were developed and tested in 2.3.14. For what I can see, it works fine.
however, I encourage people to have a careful look at the changes first.

This would close #110 and #102


kratsg:
@mmilesi - an extra question:

- why is AsgTools added as a dependency?

mmilesi:
In the Makefile.RootCore you mean?

kratsg:
See https://github.com/UCATLAS/xAODAnaHelpers/pull/118/files#diff-6297f591c84d349624bf656eef2d2b01

mmilesi:
Oh right, now I got your point.
Basically I have started implementing and testing in 2.3.13, where the asg::ToolStore::contains() method is available only when explicitly checking out a tag of AsgTools - hence the line in checkoutASGtags.py

In 2.3.14 that tag is in by default. The only issue I see is that this version of the code won't be backward compatible with 2.3.13 unless we keep the AsgTools tag in the list.

kratsg:
Ok, so I figured out why. I had removed everything prior to 2.3.14 since we bumped the official version of the script to 2.3.14+.

mmilesi:
Ciao @gfacini ,

did you get any time for looking at this? 

I understand this is a lot of new stuff pushed into xAH, but being high priority for my analysis (and not only for mine, I guess...) I'd really appreciate you having a "bird's-eye" look at it :)

Very sorry to put so much pressure! I'd go ahead and merge it myself, but this is really not how things should be done...

I extend the "invitation" to other analysts: @johnda102 , @JamesSaxon, @jdandoy .

Thanks very much in advance!

Marco

Root/ElectronSelector.cxx
```diff
@@ -343,31 +342,48 @@ EL::StatusCode ElectronSelector :: initialize ()
...
 
   // if not using LH PID, make sure all the decorations will be set ... by choosing the loosest WP!
+  //
   std::string likelihoodWP = ( m_doLHPIDcut ) ? m_LHOperatingPoint : "VeryLoose";
-  m_el_LH_PIDManager = new ElectronLHPIDManager( likelihoodWP );
+  m_el_LH_PIDManager = new ElectronLHPIDManager( likelihoodWP, m_debug );
+  
+  if  ( m_doLHPIDcut ) {
+    if ( m_debug ) { 
+       Info("initialize()", "Cutting on Electron Likelihood PID! \n ********************" );
```
gfacini: Nice! Can even be done without a check of the debug flag.

gfacini:
Looks good! Lots of nice work =)

mmilesi:
@gfacini  :heart:

PR #119, opened 2015-06-19
fix small bug in JetSelector.cxx - choice of BTagger algo

mmilesi:
Now decorating jets w/ BTag info using MV2c20, and not MV1


kratsg:
:hospital: :syringe:

PR #130, opened 2015-06-23
store fat jets 4-momenta

francescorubbo:
Store fat jets 4-momenta as for jets (switch `kinematics`)
Will add additional branches (e.g. substructure) before merging.

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -1854,10 +1854,55 @@ void HelpTreeBase::ClearJets() {
  ********************/
 
 void HelpTreeBase::AddFatJets(std::string detailStr) {
+
+  if(m_debug) Info("AddFatJets()", "Adding fat jet variables: %s", detailStr.c_str());
+
   m_fatJetInfoSwitch = new HelperClasses::JetInfoSwitch( detailStr );
+
+  // always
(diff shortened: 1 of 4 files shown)


mmilesi:
Ciao @francescorubbo ,

looks good. Do you want me to wait for you to add new stuff to this PR before merging?

francescorubbo:
Ciao @mmilesi ,

Yes, please. I expect to be done sometime tonight.

mmilesi:
:+1:

francescorubbo:
Good to merge. Thanks!

kratsg:
This looks great. Thanks @mmilesi for merging and @francescorubbo for putting the changes in :) Hope it wasn't too painful.

gfacini:
Awesome!! We have so much information include for the jets and not much for the fat jets....I wonder if there is a way of avoiding duplication of the code for the moments in common.  If one wants to write out only fat jet quantities then you can just FillJet( fatJets ) and FillFatJets( fatJets ) but if you want fat jets and non-fat jets in the same tree then to have the same moments, code might be duplicated.  Any clever solutions?

kratsg:
@gfacini I have to think hard about it. I do have a clever solution that you'll like, but we should discuss this in gchat (it's a little lengthy).

PR #137, opened 2015-06-26
Add uniform `isMC()` to xAH::Algorithm

kratsg:
This allows users to configure it consistently.

### DIFF
--- a/Root/Algorithm.cxx
+++ b/Root/Algorithm.cxx
@@ -6,6 +6,10 @@
 // RCU include for throwing an exception+message
 #include <RootCoreUtils/ThrowMsg.h>
 
+//  for isMC()
+#include <xAODAnaHelpers/HelperFunctions.h>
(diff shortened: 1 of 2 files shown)


mmilesi:
Hi @kratsg ,

I understand that this protects against a crash if the xAOD/DxAOD/mini-xAOD hasn't got the  eventInfo container and/or the eventType mask. But then my question is: do you know how to retrieve this metadata in those ill-defined cases? If I remember well, you mentioned this was an issue when trying to reprocess mini-xAODs produced outside Athena (correct me please if I am wrong...)

Btw, this looks perfectly fine to me, and a call to isMC() cleaner than having to get the Is_SIMULATION metadata explicitly every time from eventInfo.

Marco

kratsg:
Some cases can't retrieve the metadata via `eventType`. This method should work always.

PR #144, opened 2015-06-30
apply pileup weight then decorrate, protection for truth only

gfacini:
Was there a reason not to apply the PU weight to the event weight?
@jdandoy @mmilesi

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -156,6 +156,8 @@ EL::StatusCode BasicEventSelection :: configure ()
       Info("configure()", "Truth only! Turn off trigger stuff");
       m_triggerSelection = "";
       m_cutOnTrigger = m_storeTrigDecisions = m_storePassAny = m_storePassL1 = m_storePassHLT = m_storeTrigKeys = false;
+      Info("configure()", "Truth only! Turn off GRL");
+      m_applyGRL = false;
     }
 
     if( !m_triggerSelection.empty() )
@@ -358,7 +360,10 @@ EL::StatusCode BasicEventSelection :: initialize ()
   const xAOD::EventInfo* eventInfo(nullptr);
   RETURN_CHECK("BasicEventSelection::initialize()", HelperFunctions::retrieve(eventInfo, m_eventInfoContainerName, m_event, m_store, m_verbose) ,"");
 
-  m_isMC = eventInfo->eventType( xAOD::EventInfo::IS_SIMULATION );
+  if ( m_truthLevelOnly ) { m_isMC = true; }
+  else {
+    m_isMC = eventInfo->eventType( xAOD::EventInfo::IS_SIMULATION );
+  }
   if ( m_debug ) { Info("initialize()", "Is MC? %i", static_cast<int>(m_isMC) ); }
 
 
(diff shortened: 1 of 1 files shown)


mmilesi:
Hi @gfacini ,

I don't know, I think the best maybe would be to have a decoration for the PU weight separately and the full mc event weight which contains also the PU weight (as it was already).

BTW, I see that in this way #145 will be solved automatically.

gfacini:
Good that I fixed a bug! =)

For the weight, the comment above it was
"// decorate with pileup corrected mc event weight"
so I just make the comment true =)

I think it is a choice what to do.  Having the eventInfo decorated with the weights is key and a must - no controversy there.  I like the idea of a weight that has all corrections applied but it does not have to stay.  If we don't add more weights to this float, then we should not be saving it as a decoration to eventInfo.  Agreed?

Gabriel

mmilesi:
Hi @gfacini ,

I was lloking at what the ttH analysis had in the Run1 ntuples:

```
   Double_t        vertexEventWeight;
   Double_t        vertexEventWeight_Pu1;
   Double_t        vertexEventWeight_Pu088;
   Double_t        pileupEventWeight;
   Double_t        pileupEventWeight_094;
   Double_t        pileupEventWeight_080;
   Double_t        pileupEventWeight_090;
   Double_t        pileupEventWeight_088;
   Double_t        pileupEventWeight_075;
   Double_t        pileupEventWeight_103;
   Double_t        mcEventWeight;
   Double_t        mcWeightOrg;
   Double_t        lepTrigSFEventWeight;
   Double_t        lepTrigSFEventWeightUp;
   Double_t        lepTrigSFEventWeightDown;
   Double_t        MV1_70_EventWeight;
   Double_t        MV1_70_EventWeight_T;
   Double_t        MV1_70_BJetWeight;
   Double_t        MV1_70_BJetWeight_T;
   Double_t        MV1_70_CTJetWeight;
   Double_t        MV1_70_CTJetWeight_T;
   Double_t        MV1_70_MisTagWeight;
   Double_t        MV1_70_MisTagWeight_T;
   Double_t        MV1_70_BJetWeightUp;
   Double_t        MV1_70_BJetWeightUp_T;
   Double_t        MV1_70_CTJetWeightUp;
   Double_t        MV1_70_CTJetWeightUp_T;
   Double_t        MV1_70_MisTagWeightUp;
   Double_t        MV1_70_MisTagWeightUp_T;
   Double_t        MV1_70_BJetWeightDown;
   Double_t        MV1_70_BJetWeightDown_T;
   Double_t        MV1_70_CTJetWeightDown;
   Double_t        MV1_70_CTJetWeightDown_T;
   Double_t        MV1_70_MisTagWeightDown;
   Double_t        MV1_70_MisTagWeightDown_T;

```

I think we should make something similar:  

1. have an overall event weight (i.e., mcEventWeight) that accounts for ALL the weight contributions. So, the product of all of them. The user in theyr analysis code will take care of multiplying the proper number of object-level weights, depending on how many objects do they have in their SR/CRs.

2. Save separately the weight of the original MC, and the various contributions form PU, object and trigger SF. If there are systematics on the weights, let's do as we aere doing now: decorate with a vector<float> and eventually dump this in our final ntuples. It's just straightforward to pick the nominal at(0), and the systematic variations for the other components, since we have also a string with the sys names with a 1:1 correspondence to the vector components.

What do you think?

mmilesi:
Ok, I think that for the time being I can merge this PR (which closes #145 automatically)

I'll create an issue for how to deal with event weights in the FW, we can continue the discussion there!

jdandoy:
I agree that each weight should have it's own branch / auxdata (this has
been extremely useful for me in the past).  We can also decide a way to
combine them, I combine the ones I want to use manually at the end (i.e.
prescale is still not added to my total weight, but others are).
Regardless I don't think that "mcEventWeight" should be the name of the
total weight, because that is explicitly linked to the mcEventWeight single
weight.

jdandoy:
And Gabriel if you're now adding the pileupweight to mcEventWeight again,
then your "weight" branch will now have pileupweight included twice.

gfacini:
Hi Jeff,

Argh! That is not good and needs to be fixed asap.
Thanks for pointing that out.

Cheers,
Gabriel

Issue #84, opened 2015-06-04
Object cutflow


mmilesi:
Hi all!

I have been thinking about it for a while. Wouldn't be useful to have - along with the event cutflow - also an object cutflow? 

It can be useful to have both a  "cumulative" cutflow, i.e., a single histogram for every object (electrons, muons, jets,...) bookkeeping the total number of objects after each selection step of the cutflow, and a per-event cutflow. In the latter case, this is my idea:

1. make a set of **static** m_el_counter, m_mu_counter, etc. as public member variables of xAH::Algorithm, so that they can be visible and modifiable by all selectors (and by all custom classes that inherit from xAH::Algorithm). Maybe better to have a static vector of (static) counters.

2. initialise those counters somewhere ( in the steering macro maybe? ) to zero

3.  in the selector for a given object (say, electrons), set the value of the corresponding counter ( say, m_el_counter ) to the number of objects in the input xAOD container (the one you retrieve from TEvent/TStore . Usually it will be the calibrated container, whose size is by definition equal to the one of the original xAOD).  

4.  When applying the selection, increment the counter

5. Right after step 4, make an event decoration like this:

```
SG:AuxElement::Decorator< vector<vector<int> > object_cutflow("object_cutflow");
```
 Thus, every event will have a vector of vectors( --> one for each selector algo in the full chain ) attribute, something like:

```
// NB this is pseudo C++ code

// in event[i] : 

//   in electron preselector algo :  
object_cutflow.push_back ( [ nel,  nmu, njet ] );
// in electron selector:  
object_cutflow.push_back ( [ nel,  nmu, njet ] );
// in jet selector:
object_cutflow.push_back ( [ nel,  nmu, njet ] )

...

 ```

By construction, an electron selector will update the counter for electrons only, and so on...

6.  After the last selector has been called, we should have the full final number of objects in the last element of the vector ( of vectors ).

This solution might be cumbersome, but then it will be very easy to check, for a given event, the full breakdown of an object cutflow in the output ntuple, even with a direct TTree::Draw, by accessing the right components of the vector with [ ]. 

Of course we have to give  convention for the elements in the vector of counters. We can have   [ nel,  nmu, njet ] for instance...

This is no more than thinking out loud.

kratsg:
How would this work if there were multiple kinds of jets being used? (eg: R=0.4 and R=0.8 and R=1.0?)

mmilesi:
Hi @kratsg ,

[...]

PR #151, opened 2015-07-01
Add global object cutflow histogram for ElectronSelector and MuonSelector

mmilesi:
This partially fixes issue #84


mmilesi:
Good job bot

PR #153, opened 2015-07-01
protection from massive powheg weight

gfacini:


gfacini:
@kratsg @mmilesi 
There is a huge weight in this sample that needs to be removed like it never existed.
Any strong objections?

kratsg:
Why does it have a huge weight in the first place? Do we really want to hardcode that? That seems out of the context of xAH :(

gfacini:
Generator flaw.  It is not analysis specific - no one wants this event.  See slide 2:
http://hep.uchicago.edu/~johnda/work/PowhegVsPythia.pdf

kratsg:
Can we detect the sample being used automatically from the dataset ID and remove it, rather than a configurable variable?

Root/BasicEventSelection.cxx
```diff
@@ -374,6 +377,12 @@ EL::StatusCode BasicEventSelection :: initialize ()
     }
   }
 
+  m_cleanPowheg = false;
```
gfacini: @kratsg

gfacini:
L380!
If the runnumber matches then the check is done

kratsg:
The runNumber is unique within a sample, but not within multiple samples. I'm saying, we should have the script grab the DID automatically and apply the filter if the DID matches. Rather than having a configurable check. Not everyone will know to configure it.

gfacini:
runNumber is DID!

kratsg:
I'm stupid. This looks fine. @johnda102 will be pleased.

gfacini:
1) False
2) Who cares!? =p

PR #165, opened 2015-07-07
small fixes - one tmp for trigger matching

gfacini:


gfacini:
@mmilesi 
Temp fix so if one does not make a selected container in muon selector that trigger matching does not crash but is just skipped

Issue #120, opened 2015-06-19
Perform Trigger Matching for Electrons


mmilesi:

mmilesi:
Hi all,

this is on hold, since we might have the trigger matching already implemented in the next coming derivations with the Trigger Navigation Smart Slimming (or whatever it's called). I'll keep you posted if something changes.

Marco

mmilesi:
Hi again,

I need to amend the previous message :disappointed: 

By looking at this [JIRA issue](https://its.cern.ch/jira/browse/ATLASG-133), it seems that trigger matching in the derivations won't be available soon. Thus, I'll try to get the electron trigger matching up and running in xAH, as already is for the muons

mmilesi:
Closed via #179

PR #179, opened 2015-07-13
SKELETON for electron trigger matching

mmilesi:
Hi @kratsg, @johnda102, @jdandoy, @gfacini ,

this PR contains mostly the code to perform the electron trigger matching and to conveniently store the matching info on the final tree.

Please, have a detailed look at the commit message and at the code to get an idea of what I have added. In the config file for the ElectronSelection.cxx module you can find some detailed explanations on how to configure the matching.

Sadly, at this stage the ASG TrigEgammaMatchingTool - which is included in the release and is supposed to perform the matching - does **not** work outside Athena due to some not-well-defined issues with parts of the TrigDecisionTool which are used by the tool internally.

This does not mean however that the code will crash: simply, we do not see any trigger-matched electron whatsoever.

Ulla confirmed however that a working version of the tool will be available (hopefully) soon for RootCore based analyses too, and I do believe the interface won't be much different to the one we have atm. 
Also, my additions do not cause any problem with the rest of the code, so should be safe to incorporate them.

**CAVEAT**

I have added a pointer to  xAOD::TStore (in addition to the one to xAOD::Event, that was already in place) as an additional parameter to HelpTreeBase constructor:

https://github.com/UCATLAS/xAODAnaHelpers/blob/EgammaTrigMatching/xAODAnaHelpers/HelpTreeBase.h#L49

I did this because I needed it in a temporary implementation of the matching. If people are not happy with it, I can edit the PR and go back to the previous implementation, but I still believe it could be useful to have access to TStore in the tree in the future.

Sorry for the super long message


kratsg:
Looked it over, I agree with the current implementation. It would definitely be useful later.

mmilesi:
Thanks @kratsg, appreciate that.

I'd like to hear the opinion of @jdandoy, @gfacini, @johnda102 too. Especially to make sure none of the change I made here will negatively affect people.

mmilesi:
Ok, I interpret this as tacit consent. Merge this and close #120

jdandoy:
Yeah definitely negatively affect us.  Sorry for not paying closer
attention but there are a ton of things to do right now and I'm paying 0
attention to xAODAnaHelpers development.  Right now we need to not make any
changes to xAODAnaHelpers until after EPS that would break anyones code.
All changes should be postponed until afterwards.   Can we make that a
policy for the time being? Can we briefly revert this or is it needed?

Jeff

mmilesi:
Hi @jdandoy 

I have waited 2 days before merging this. I am sorry but having explicitly tagged all the people that might be involved by the changes I was expecting them to pay attention, at least at the careful description I gave in the PR. 

If you want to make a statement in the README to freeze the PRs, for me it's just fine.

Also, please bear in mind that maybe not everyone knows what others are up to at anytime, so if there are some delicate things ongoing please make an announcement so that everyone knows. Perhaps the good old mailing list is the best place to go.

And of course you can go ahead and revert this PR if it's causing issues.

Cheers
Marco

jdandoy:
Sorry I had assumed you were aware about EPS!!  Right now most every
analysis that is using this package is rushing to freeze results (by
friday) on brand new data we're actively collecting in time for EPS next
week.

I didn't know the README was a place for technical development
requirements,  I don't think we've listed any rules for this sort of
situation as it hasn't come up before, maybe we should.  So there's no
blame to go here, just miscommunication.  But neither myself, John, or
Gabriel noticed this would happen.  The email subject should say something
about breaking packages, and if you don't get any response maybe it would
be best to message us.   Again this is literally the busiest time of my
entire life.

I'm not as proficient in git as you and Giordon so I'm going to try to just
make a reverted branch for our EPS analysis now with my recent additions.

Jeff

gfacini:
Bottom line....we need to communicate better about rush
periods so that we avoid snafus in the future.  No fingers
to point here - we just do it better the next time.

Marco, I appreciate you flexibility and patience!

I promise, as soon as I can, Ill go through your changes
and try to get us adjusted to them so you are not in
a further delay.

Thanks again for the help and patience!

Cheers,
Gabriel

jdandoy:
Yes thanks a ton for the new PR!  It really helps us out alot now.

PR #194, opened 2015-07-17
GRL for EPS (period C3 added, 38.2 pb-1)

mmilesi:
Just added to data/ folder


mmilesi:
@jdandoy , @gfacini 

YOU DO THIS :kissing_heart:

jdandoy:
Looks like you've got it under control

PR #195, opened 2015-07-20
Overload HelpTreeBase constructor 

mmilesi:
This should make everybody happy.

@jdandoy ,@gfacini

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -509,24 +509,25 @@ EL::StatusCode BasicEventSelection :: initialize ()
     std::string tmp_PRWFileNames = m_PRWFileNames;
 
     // Parse all comma seperated files
-    while( tmp_PRWFileNames.size() > 0){
-      unsigned int pos = tmp_PRWFileNames.find_first_of(',');
-      if( pos == std::string::npos){
+    //
+    while ( tmp_PRWFileNames.size() > 0) {
+      size_t pos = tmp_PRWFileNames.find_first_of(',');
+      if ( pos == std::string::npos ) {
         pos = tmp_PRWFileNames.size();
         PRWFiles.push_back(tmp_PRWFileNames.substr(0, pos));
         tmp_PRWFileNames.erase(0, pos);
-      }else{
+      } else {
         PRWFiles.push_back(tmp_PRWFileNames.substr(0, pos));
         tmp_PRWFileNames.erase(0, pos+1);
       }
     }
-    while( tmp_lumiCalcFileNames.size() > 0){
-      unsigned int pos = tmp_lumiCalcFileNames.find_first_of(',');
-      if( pos == std::string::npos){
+    while ( tmp_lumiCalcFileNames.size() > 0) {
+      size_t pos = tmp_lumiCalcFileNames.find_first_of(',');
+      if ( pos == std::string::npos ) {
         pos = tmp_lumiCalcFileNames.size();
         lumiCalcFiles.push_back(tmp_lumiCalcFileNames.substr(0, pos));
         tmp_lumiCalcFileNames.erase(0, pos);
-      }else{
+      } else {
         lumiCalcFiles.push_back(tmp_lumiCalcFileNames.substr(0, pos));
         tmp_lumiCalcFileNames.erase(0, pos+1);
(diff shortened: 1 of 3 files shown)


mmilesi:
If no one is afraid of potential problems with this, could we merge?

jdandoy:
The warning below appears for every event and may slow alot of things down and really increase the log if we're running on the grid.  Maybe it can somehow just be mentioned once or not at all?

 ******************************** 
Warning in <FillEvent()>: 
 ******************************** 
 Passing the TEvent pointer to this function is deprecated and will be phased out soon! (you have already set it via HelpTreeBase constructor...) 

If we want to keep things backwards compatible then we can't phase this out.

mmilesi:
One option could be to avoid passing the TEvent object, as the warning suggests, since there's no need for it. Is it really a huge issue for you to do that in your analysis code?

If yes, I will make sure the warning appears only once

Btw, with the new changes in master I'll need to check the cnflicts before merging this

kratsg:
@mmilesi you'll need to rebase this branch ;)

jdandoy:
It's not a huge issue for me.  But it's not just my analysis code, it's a
host of people who use xAODAnaHelpers but aren't plugged into any changes,
or who are unable to edit the code to account for this themselves.   Is it
a huge issue to just have a second function to ensure backwards
compatibility?  That's something we've promised we'd preserved, and I think
Gabriel and John agree that we should maintain it.  If they disagree then
we can fine a different compromise.  @johnda102 @gfacini

kratsg:
@jdandoy - I think if people are plugged into the code, they're using a specific version (and not updating as often). For those who are always updating - I think they have to accept that these changes happen from time to time (not a HUGE interface change) but minor ones that more or less require updating one line. As long as we advertise these interface changes temporarily in the README and via the email-list -- I think we would be ok?

mmilesi:
I agree @kratsg , I think that xAH users will all receive emails, so we can advertise any interface change in there.

Anyway, I will make sure the Warning message above get printed only once (it's basically encouraging people how to use FillEvent() the proper way!), fix the conflicts and let you make the final check.

mmilesi:
@jdandoy , I think everything is ok now.

jdandoy:
I disagree.  We are trying to make a product and we want people on ATLAS to
use it.  These are extremely busy people who get hundreds of emails a day
and are swamped with many projects.   They need their code to continue to
just work with the latest and greatest features, and they don't have time
to stay plugged in or make small changes to all their algorithms just
because we feel it's not clean to pass a pointer too often.  They don't
have to accept random changes that break their code, they can just find
another framework to use that promises backwards compatibility, which I
thought would be one of our tenants.  @gfacini and @johnda102 were very
adamant about this a week ago, what do you guys think?

It's just a few lines of code, we can stick them in the bottom of the file
under
///// backwards compatibility /////
and forget about them.  To help keep things even cleaner, would we be able
to use a delegating constructor for this?

Jeff

mmilesi:
@jdandoy,

with the current implementation of the thing, everything will be just fine and people will not notice that we have a different constructor. Plus, I implemented the FillEvent() method right in order to let people decide whether to pass the pointer or not.

If you don't find any issue with this, can we go ahead and merge?

gfacini:
Hi Marco,

Can we change the constructor to
HelpTreeBase(xAOD::TEvent *event, TTree* tree, TFile* file, const float units = 1e3, bool debug = false, bool DC14 = false, xAOD::TStore* store = nullptr );

So that we have 1 constructor and backwards compatibility is not broken?

Cheers,
Gabriel

mmilesi:
Hi @gfacini ,

in principle yes, we can.

However, my ttH analysis package - which makes use of a class that derives from HelpTreeBase - already makes use of the "new" constructor in the way it's implemented now.

If you do not have strong arguments against having an overloaded constructor, I'd prefer to stay as we are now in this PR.

Let me know, thanks!

Marco

gfacini:
I added TStore to the original constructor as the last item and gave it a default value.  I then took the constructor Marco wants to use and just called the original constructor from that one. Now everything goes through the one constructor.

If everyone agrees, we can merge this and get it over with already.

mmilesi:
I am happy with this.

PR #198, opened 2015-07-24
Update to AB 2.3.21, + Changes

jdandoy:
Update to latest Analysis Base 2.3.21 .  A change to config file for JER tool because of the update.

Get latests GRL for Run II using checkoutASGTags.py.  During data taking these files change rapidly, so it is useful for the user to know they have an updated version. Removed default version of these from data directory.  

Added bcid to HelpTreeBase

@gfacini @mmilesi


kratsg:
Thanks for updating the `.travis.yml`!

mmilesi:
Hi @jdandoy ,

the GRL update looks nice!

So, whenever a new GRL is released we should update the python script, correct?

jdandoy:
Hi Marco,

No, in general when a new GRL is released it overwrites the previous one.
So if there is an update to the GRL we should only need to rerun
checkoutASGTags.py again!

mmilesi:
Ok, good to know. Merge this

PR #199, opened 2015-07-25
Revisiting how to store passed trigger info in NTuple

mmilesi:
(I like to stir things up lately :smile: )

**No interface change in here** , but I'd like to change slightly the logic of how trigger items/chains that were fired in the event are stored for later use in the output ntuple.

Please have a look at the commit message and at the code changes: basically, I'd get rid of `passAny`, and `passHLT/L1` would now be set to `true` depending on the trigger list/regexp passed in input by the user, not for ALL the HLT/L1 items.

This - along with being more natural imho - would help reducing the size of the output branches to the necessary info only, if one wishes to apply trigger event selection at plotting level (as I would like to do in my analysis).

Please let me know what do you think.

Cheers

Marco


kratsg:
@mmilesi One of the best ways to do this is to store a list of all triggers per file, and then just store binary data representing which triggers fired (010101110110 - for example).

mmilesi:
Hi @kratsg ,

we already store a branch `passedTriggers ` with the names of the triggers from the input list that were fired in the event.

The `passL1/HLT` flags will be true if at least one of the input L1/HLT triggers were fired, and can be used at plotting level to reject uninteresting events - i.e., the ones where none of the triggers of interest have been fired.

In my rationale,
If one would like to plot events where - say - `HLT_e24_lhmedium_L1EM18VH` or `HLT_mu40` were fired, now he/she should simply do in a `TTree::Draw()`:

```
physics->Draw("MY_VAR", "passHLT==1 && (passedTriggers =="HLT_e24_lhmedium_L1EM18VH" || passedTriggers =="HLT_mu40" )")
```

Does this make sense?

Marco

gfacini:
Looks good to me! I am happy to merge.  However this changes in Root/MuonEfficiencyCorrector.cxx should be revisited since that file has undergone some changes and is not relevant to this change.

@mmilesi can you double check the change there (i.e. is it already in the master?) then merge when you are ready.

Thanks for removing this redundancy.  I would also like to change this to pick up the trigger tool directly...for another day.

mmilesi:
Hi @gfacini ,

Waiting for xAHBot to resurrect. 

If the testing goes well, this can be merged safely.

gfacini:
Nice!

PR #200, opened 2015-07-27
minor changes, GhostTrack helperfunction

gfacini:


kratsg:
Looks good. The only thing I'd worry about is the `dynamic_cast` which might give us issues later, but that's it.

PR #202, opened 2015-07-28
B tag sf to tree

gfacini:


gfacini:
@jdandoy @johnda102 @kratsg 
Some dirt in the code for sure....
I centralized the MV2c20 cut value locations
The b-tagging group uses the efficiency value as the name so this is what I used for the tree but in the code and in the configs it is still Tight, etc...a minor point as of now

Using JetSelector and not cutting on the b-tag output but asking for any OP will decorate the jet with decisions ( someone else did this )
Running BJetEffCorr for a given OP then decorates with the scale factors ( also already done )
Now both can be put into the tree - and can be done for all 4 OPs

PR #206, opened 2015-08-03
Fix Warning when compiling BasicEventSelection.cxx

mmilesi:
A small fix to get rid of the following warning:

```
warning: comparison is always false due to limited range of data type [-Wtype-limits]       
 if( pos == std::string::npos){
```

See [here](http://stackoverflow.com/questions/131803/unsigned-int-vs-size-t) for more info.

Marco


mmilesi:
Hi @kratsg,

Waiting for xAHBot to resurrect. 

If the testing goes well, this can be merged safely.

gfacini:
We should go through all the warnings!!

PR #207, opened 2015-08-03
Event CutBookkeepers MetaData - This is it (?)

### LINKED ISSUE #15: Metadata - EventBookkeepers for mc15
Read metadata and lumiblocks in mc15.

mmilesi:
Hi @gfacini , @kratsg , 

I dug into this email thread on PATHelp:

https://groups.cern.ch/group/hn-atlas-PATHelp/Lists/Archive/DispForm.aspx?ID=14739

and it seems like we needed to add a safety check where we retrieve the event bookkeepers for the initial number of events. 

Basically we need to make sure that we do not skip files where there are Incomplete CutBookkeepers (these files in principle *should* be skipped...), but ALL of them have  unknown input stream. 
This issue has been observed in some DAOD flavours, so it's good to make sure we handle it properly.

Please read through the email thread for the technicalities.

Another little thing I have added: now the metadata histogram is fillled with meaningful values also when processing **full** xAODs. In this case, the numbers are taken directly from the `CollectionTree` TTree (there is a TEvent::getEntries() method, but it cannot access the tree weight, so I got the info straight from the Tree)

This would close #15 

Marco

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -3,8 +3,8 @@
  * Basic event selection. Performs general simple cuts
  * (GRL, Event Cleaning, Min nr. Tracks for PV candidate)
  *
- * G. Facini, M. Milesi (marco.milesi@cern.ch)
- *
+ * G. Facini (gabriel.facini@cern.ch)
+ * M. Milesi (marco.milesi@cern.ch)
  *
  *******************************************************/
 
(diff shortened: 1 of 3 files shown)


gfacini:
Hi Marco,

Looks good! Thanks for keeping a close eye on this MetaData business.

Cheers,
Gabriel

PR #214, opened 2015-08-05
B tag changes

gfacini:
Tested and ready to go!!

The user now uses on the names provided by the flavor tagging group as the operating point.  Further information has been added to the TWiki.

Please have a look - I think this is ready to be merged.


gfacini:
@kratsg, @johnda102, @jdandoy, @alexandertuna @mmilesi, @AndreaCoccaro 
its a decent sized change - have a look.

kratsg:
Looks good. I worry about the ugliness with the tree maker section. Is there really no clever way?

gfacini:
It is super ugly but I did not have the time (or a really good idea) to make it elegant.  If you have ideas, I am happy to change it.  Can we merge then beautify since making it less ugly should not lead to any functional changes?

kratsg:
Yes, I will be fine with refactoring. We should definitely encourage folks to clean things up as we go :)

gfacini:
And documentation!! I at least wrote what the tools is supposed to do on the TWiki - got to document the header / config options next.

gfacini:
If no one comments or has objections I will merge this soon.

mmilesi:
Hi @gfacini 

I had a quick look and looks fine! However, I would like to have a deeper look into it. Would it be ok for you to wait until monday? Unfortunately these last days I am overwhelmed with stuff to do :(

gfacini:
Hi @mmilesi,
Monday is a bit too long away.  I have tested it so I know it works.  Are you in the middle of developing something and therefore need the master a certain way?  If not, Ill merge this now.
Gabriel

mmilesi:
Ok, then you can merge :+1:

PR #224, opened 2015-08-10
Minor fixes and some improvements on new BTagging changes

mmilesi:
Some fixups for latest changes in BTagging handling:

-) Set a default value for the BTagging algorithm OperatingPoint (natuarally overrridden in case the user passes a different one from config/python macro)
   For now, use "FixedCutBEff_70" as default
-) Update ALL the relevant config files 
-) Update checkoutASGtags.py in order to automatically copy CDI file in xAH/data folder as soon as the CDI files directory is checked out from SVN
   I think this is the best solution since we don't want to put in version control too many 'big' ( > 1MB ) files: the config files will look for the CDI file into xAH/data (to be grid-safe)
-) In JetSelector.cxx and BJetEfficiencyCorrector.cxx, check if the BTagging tools are already in asg::ToolStore before allocating new heap memory for them

@gfacini, if I misunderstood some of your changes and accidentally overwritten them, please let me know and I'll revert back.

**An important side comment**

In addition to the changes in our personal analysis config files, I seriously think everybody should make an effort at updating the xAH default config files (and/or the python steering macro) when configuration of the modules is changed. This morning I had to struggle a bit before making the latest changes working on  "stand-alone" xAH.
I might have missed new guidelines, but I remember we said once that it's crucial to deliver a working xAH stand-alone package everytime we commit new stuff. People should be able to compile and run it out of the box every time they upgrade to newer versions.

Of course, I might be the one that screwed up in this regard in the past, so, my apologies for that! 

Cheers :smiley: 

Marco

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,9 @@ The xAOD analysis framework, born out of ProofAna...or not.
   - [Previous](#previous)
 - [Common Issues](#common-issues)
   - [Missing Metadata Information](#missing-metadata-information)
-- [Who uses us?](#who-uses-us)
+- [Community](#community)
+  - [Tutorial](#tutorial)
+  - [Who uses us?](#who-uses-us)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -53,7 +55,13 @@ UseMetadata False
 
 and try again.
 
-## Who uses us?
+## Community
+
+### Tutorial
+
+[Brian Tuan](https://github.com/btuan) made a [tutorial](http://twiki.mwt2.org/bin/view/UCATLAS/AnaHelpersExample) for those starting with xAOD analysis for the first time.
+
+### Who uses us?
 
 The following list are packages / analyses searches that depend on xAH. We thank them for their hard work and hope they continue to use us!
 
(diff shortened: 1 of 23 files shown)


gfacini:
Hi @mmilesi 

A lot of white space changes...1 comment:

The initialization of the BTaggingEfficiencyTool and BTaggingSelectionTool in BJetEfficiencyCorrector and JetSelector needs to be done with the name included as it was before:

-  std::string sel_tool_name = std::string("BJetSelectionTool_") + m_name;
-  m_BJetSelectTool= new BTaggingSelectionTool( sel_tool_name );

This is so anyone who wants to select b-jets at multiple operating points can do so.  Otherwise it will complain that the tools have already been created.

Cheers,
Gabriel

mmilesi:
Hi @gfacini 

I have reverted the changes you requested. You can merge anytime

gfacini:
@mmilesi 
I un-reverted the revert ;-)
I think you were correct to check if the tool exists but the name still had to be unique for the different b-tagging tools.  I added back that check. It might never get used but it is good practice.

PR #226, opened 2015-08-10
B tag sf bug fix

gfacini:


gfacini:
@mmilesi 
here is your bug fix

PR #228, opened 2015-08-10
Add analysis base files

johnda102:
Adding job submission files used in common by di-jet/Multi-jet and BJet trigger code

### DIFF
--- a/scripts/downloadAndMerge.py
+++ b/scripts/downloadAndMerge.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python
+
+##******************************************
+#downloadAndMerge.py
+#download and merge output datasets produced by dijet analysis code on the GRID
+
+#NOTE before starting, set the variables below
+#NOTE just run the script without any parameter to get instruction
+#HOW TO time python -u downloadAndMerge.py dataset
+#NOTE 'dataset' can have wildcards, the matching datasets will be treated independently
+#EXAMPLE time python -u downloadAndMerge.py user.gfacini.*.v06_*.root/
+##******************************************
+
+#------------------------------------------
+#import
+import os, sys, subprocess, glob, shutil
+import argparse
+parser = argparse.ArgumentParser(description="%prog [options]", formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+parser.add_argument("--container", dest='container', default="None",
+     help="Name of dataset to be downloaded, may include wildcards")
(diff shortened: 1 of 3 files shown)


kratsg:
@johnda102 why do we want to add the header files?

I'm also a little confused about why you need to merge your files manually here. I'll wait for @gfacini to comment first.

johnda102:
@kratsg there is only one header file.  In includes functions that are used in job submission for several analysis.  Better implement it once in the base code that have duplication for each package. 

Not sure what you're referring to with the manual merge.  I pushed a branch to orgin and requested a pull (didnt I?)

kratsg:
> Not sure what you're referring to with the manual merge. I pushed a branch to orgin and requested a pull (didnt I?)

I meant the 'downloadAndMerge.py' script. Do people not use xAODMerge on the grid (should be working as of AB2.3.23) and use TChain / hadd?

johnda102:
Nope all the EPS code was pre AB2.3.23 (Im actually on 2.3.21 for the latest development) 
Never heard of xAODMerge. The script that was included hadd files after a dq2-get

mmilesi:
Hi @kratsg ,

isn't `xAODMerge` specifically designed to merge `xAODs `- and not just simple `TTrees`?
I know of its existence, but I never really checked that out.

Marco

kratsg:
@mmilesi it merges xAODs and histograms. xAODs are just giant ntuples. I'm still not a fan of adding these header files because it increases the compilation time for anyone using xAH -- and that all of this code is already in xAH_run.py which nobody uses :( (except David and a few undergrads here).

johnda102:
@kratsg:  Im sure xAH_run is great.  However the bottom line is that it came along after a lot of code was already developed and there is overhead to change. These existing packages have a wide user base beyond the people active on here on git. Introducing a structural change now to how the end-user interacts with the code could jeopardize our ability to turn around results in order of week. 

Can you point us to instructions on how to:
    - use xAH_run to run jobs locally 
    - submit to the grid
    - fetch/hadd output from the grid 
    - to migrate existing utils to xAH_run
And ill see what needs to be done to make xAH_run usable on our end.

kratsg:
Hi @johnda102, unfortunately, `xAH_run.py` has been around since roughly March. I just never got around to advertising it since there were lots of other issues at the time. It will be getting a major update soon to incorporate a lot of new features that will support most folks.

PR #229, opened 2015-08-11
Final Shape for Muon & Electron Efficiency Correctors

mmilesi:
Hi @gfacini, @kratsg , @johnda102 , @jdandoy 

this is a pretty big PR, but also quite relevant imo.

I have reviewed all the Electron* and Muon* algorithms, fixed some bugs related to the systematics treatment, cross-checked them against the CP groups recommendations and, most of all, enabled all muon and electron efficiency SFs  and dumped them in the output tree.

The branches are `vector<double>`, with the first component representing the nominal case, and the other components each systematics.
To know "which component corresponds to which syst", one should dump a vector<string> to the output tree that at present is only saved in TStore. I can add this in a future PR if you all agree.

A nota-bene:

1. Trigger SFs: 
these are EVENT-LEVEL weights (i.e., to each event, its SF vector): people should multiply the weight at plotting level only once per event

2. Reco/Isolation efficiency (for muons) & Reco/ID efficiency (for electrons): 
these are OBJECT-LEVEL weights, meaning that people should multiply the object-SF depending on how many objects are being selected  - i.e, supposing we want to plot a-la-TTree::Draw(),  if we have el0 and el1, the event needs to receive a weight corresponding to` el_SF[0][0] * el_SF[1][0] ` (second component means "take the nominal")

TO DO (for a next PR):

1. Save different SFs depending on the isolation/ID WP (as done by @gfacini for Bjets)
2. Dump `vector<string>` w/ syst names for each SF (see above)

Cheers

Marco


gfacini:
Hi @mmilesi,

A big PR indeed!  Nice work!!! Thanks for making the names of the scale factors more meaningful in the for the decoration.

Below I have some comments on what I see.  Please let me know what you agree with and what I can help you implement.

1.
When working on the BJet stuff recently I noted what John had done which was to call CP::make_systematics_vector.  This returns a list that includes the an empty SystematicSet for the nominal.  In fact, I took advantage of that here:

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/BJetEfficiencyCorrector.cxx#L332

to get only 1 SystematicSet in the case where one was not running on systematics.  In this PR you use this call and also add by hand an empty SystematicSet.  I think we can drop this adding by hand and just use the CP function.  Do you agree?

2.
If someone sets up 2 efficiency corrections to get two working points say to do an optimization study or whatever, the CP tools i.e. AsgElectronEfficiencyCorrectionTool will already exist when the second algo is created.  To avoid this, we should do here what we did for the b-tagging and add m_name to the name of the tool.  Do you agree?

3. 
Why doubles?  Do you really need that much precision on a scale factor that is probably only known to 3 significant digits.

4.
Curious: Why prepend with the syst names
https://github.com/UCATLAS/xAODAnaHelpers/pull/229/files#diff-7915aad1fab9644670d64c9d65b28e2bR565
ElPIDEff_SFBLAH seems more organized than BLAHElPIDEff_SF as one can sort the strings and find the ones of the same type very easily.  Say you use these in some statistical machinery and sort the Nuisance Parameters after a fit, having ElPIDEff_SF as the start of the string will keep all those NPs in the same place in diagnostic plots etc. 

5.
Sort is not initialized in the MuonCalibrator:
https://github.com/UCATLAS/xAODAnaHelpers/pull/229/files#diff-76bd3f85a73152f1f60bf2ddfa6266c8L71

6.
Missing initialization:
https://github.com/UCATLAS/xAODAnaHelpers/pull/229/files#diff-d15015d45879aed5430a665acf4b889eL67
Unless it was wrong "Data15" seems like a good option.

7.
What is the plan for the name of the systematics?  This vector of names is only needed for 1 event since it going to be the same for all.  It is a good point you raised and we should do something similar for the BJets.  Can we write something to the output file when the tool is initialized?  We don't really need a vector of strings...do we?

Again - some really good work.

Cheers,
Gabriel

mmilesi:
Hi @gfacini 

thanks very much for spending time to go through this, and for the comments too!

Inline replies:

> 1.
> When working on the BJet stuff recently I noted what John had done which was to call CP::make_systematics_vector. This returns a list that includes the an empty SystematicSet for the nominal. In fact, I took advantage of that here:
> 
> https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/BJetEfficiencyCorrector.cxx#L332
> 
> to get only 1 SystematicSet in the case where one was not running on systematics. In this PR you use this call and also add by hand an empty SystematicSet. I think we can drop this adding by hand and just use the CP function. Do you agree?

I am indeed calling `CP::make_systematics_vector`, but only inside a debug block, where I print out the list of the systematics affecting this tool (i.e., ALL of the possible systematics):

```
  if ( m_debug ) {
  
    // Get a list of affecting systematics
   //  
    CP::SystematicSet affectSystsReco = m_asgMuonEffCorrTool_muSF_Reco->affectingSystematics();
    //
    // Convert into a simple list
    //
    std::vector<CP::SystematicSet> affectSystRecoList = CP::make_systematics_vector(affectSystsReco);
    for ( const auto& syst_it : affectSystsReco ) { Info("initialize()","MuonEfficiencyScaleFactors tool can be affected by reco efficiency systematic: %s", (syst_it.name()).c_str()); }
  }
```

The list of the actual systematics we are going to run over is the one returned by `HelperFunctions::getListofSystematics( recSystsReco, m_systNameReco, m_systValReco )`, which internally resizes the list based on the systematic name chosen by the user. This makes sense, since we might be interested in looking at only one systematic at a time (i.e, only the nominal "empty" sys) and so there's no need to loop over a full list. (btw...didn't you write this function? Why aren't you using it in BJetEfficiencyCorrector?)

However,  `HelperFunctions::getListofSystematics` in the current implementation does not add an empty `CP::SystematicVariation` at the top of the list, as it should. That's why I added manually the empty set to the list.

Agreed that this is pretty ugly, I have slightly modified `HelperFunctions::getListofSystematics` in order to have this done internally. Also, I realised that even in the case we are passing a single sytematic variation as input, we still want to have the nominal case! The new implementation of  `HelperFunctions::getListofSystematics`  takes this into account.

Btw, in this way the flags `m_runAllSyst*` become useless, so I removed them

> 2.
> If someone sets up 2 efficiency corrections to get two working points say to do an optimization study or whatever, the CP tools i.e. AsgElectronEfficiencyCorrectionTool will already exist when the second algo is created. To avoid this, we should do here what we did for the b-tagging and add m_name to the name of the tool. Do you agree?

Mmmhh....I have a similar situation in my analysis but with selectors: I call them to preselect objects with loose cuts, then I apply the OR, and then I call again the selectors but tightening the cuts.
By putting the check on `asg::ToolStore`, the second algoriithm will take the same instance of the CP tool previously allocated, and simply re-configure it via `setProperty`: it seems to work indeed, and does not imply having to instantiate multiple tools with different names.

Have you ever tried to do like this? I guess one can re-configure the CP corrector tools as happens with the CP selector tools, but we should try it out.

> Why doubles? Do you really need that much precision on a scale factor that is probably only known to 3 significant digits.

Good point: can you check everything is fine by making the switch?

> 4.
> Curious: Why prepend with the syst names
> https://github.com/UCATLAS/xAODAnaHelpers/pull/229/files#diff-7915aad1fab9644670d64c9d65b28e2bR565
> ElPIDEff_SFBLAH seems more organized than BLAHElPIDEff_SF as one can sort the strings and find the ones of the same type very easily. Say you use these in some statistical machinery and sort the Nuisance Parameters after a fit, having ElPIDEff_SF as the start of the string will keep all those NPs in the same place in diagnostic plots etc.

I agree with you in principle....however I wonder why I did this in the past, there might have been a reason... I am trying to remember. For the time being, can we stick to that and then change later if that doesn't make sense for any reason? 

> 5.
> Sort is not initialized in the MuonCalibrator:
> https://github.com/UCATLAS/xAODAnaHelpers/pull/229/files#diff-76bd3f85a73152f1f60bf2ddfa6266c8L71
> 
> 6.
> Missing initialization:
> https://github.com/UCATLAS/xAODAnaHelpers/pull/229/files#diff-d15015d45879aed5430a665acf4b889eL67
> Unless it was wrong "Data15" seems like a good option.

Should be ok now. `m_DataPeriod` is useless now and has been removed

> 7.
> What is the plan for the name of the systematics? This vector of names is only needed for 1 event since it going to be the same for all. It is a good point you raised and we should do something similar for the BJets. Can we write something to the output file when the tool is initialized? We don't really need a vector of strings...do we?

I had a thought about this a while ago when I added the electron trigger matching: unfortunately I haven't found a way to save a vector in a `ROOT::TTree` only for one event, so I ended up saving a vector<string> branch with the same content for all the N events. That's clearly a waste of space, but I couldn't find any other way. Any ROOT experts out there?

**************

If you see some minor things that still need to be fixed, just go ahead.

Cheers

Marco

mmilesi:
Hi @gfacini ,

I noticed that after your last change, we end up with zero events in the output tree.

In the selector at line https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/MuonSelector.cxx#L534:

```
    if ( m_debug ) { Info("execute()", " input list of syst size: %i ", static_cast<int>(systNames->size()) ); }
```

now prints out a size of 0

The problem seems to stem already at the calibrator stage: without the empty syst addedby default, now `m_outputAlgoSystNames ` for the calibrator has a size of 0 , hence `m_inputAlgoSystNames ` for the selector has a size of 0, and the algorithm does nothing.

If there is no strong reason against having always the nominal case, I'd keep it

gfacini:
Hi @mmilesi 

Lets take a step back

There are two ways to run systematics:
1. The user gives the name of the systematic they want and only that one is run.  This is if you want to submit say 1 job per syst
2. The user gives "All" and all systematics and the nominal are run.  The change I added was only to add the nominal to the list of systematics if the user was asking for all. 

I think the code follows this logic.  Can you confirm that you are giving "All" to the tool?

As for your previous comment on the selector being reconfigured....really!? That is not what I was expecting at all.  Ill try to double check that today.

Cheers,
Gabriel

mmilesi:
Hi @gfacini 


> There are two ways to run systematics:
> 1. The user gives the name of the systematic they want and only that one is run. This is if you want to submit say 1 job per syst
> 2. The user gives "All" and all systematics and the nominal are run. The change I added was only to add the nominal to the list of systematics if the user was asking for all.
> 

Okay, I got your point.
If then we don't want to have always the nominal, I think you should change the `if` statement in the `HelperFunctions::getListofSystematics()` function to look like:

```
iif ( systName.find("All") != std::string::npos || systName.empty() ) {
    outSystList.insert( outSystList.begin(), CP::SystematicSet() );
    const CP::SystematicVariation nullVar = CP::SystematicVariation(""); 
    outSystList.back().insert(nullVar);
 }
```

In that case, if the input syst name is an empty string (this is what happens if you leave a "blank" option in the config file), the empty syst variation is added as well, so we will end up with a systematic list that is not an empty vector, rather has 1 element which is an empty string.

As I said, not doing this leads to meaningless results with the current implementation.

Would you mind giving it a go?
I won't be able to do it tonight...I'd be really tahnkful if you could do that :-)

Marco

kratsg:
@gfacini One point to make is that if we start doing mini-xAOD outputs, we will want to copy over the nominal container and shallow copy (and record with `shallowIO=true` the systematic container). This means we're not recording the full shallow copy container, but only the portion that changes -- to be efficient in disk usage.

gfacini:
Hi Giordon,

mini-xAODs would be great! I am not sure that I can switch
to them for this analysis very easily since we have a lot of
machinery built on the tree but in general the idea is sound.
Is that ready?

Gabriel

kratsg:
@gfacini There's some testing and issues with systematics that I've emailed out about a month ago. More groups are starting to report the same issue, so I have to wait until critical mass before it gets fixed. I'm still working on workarounds.

PR #234, opened 2015-08-17
Add branches with number of BTagged jets PLUS bug(?)fix

mmilesi:
Hi @gfacini ,

this adds a branch with the number of jets fulfilling some BTag requirement (which I think is very useful, at least for my analysis purposes is :blush: ) plus...a fix to what I suspect is a bug in BJetEfficiencyCorrector.cxx.

Basically I have noticed that, with such an implementation of the BTag decoration:

```
      //
      // Add decorator for decision
      // 
      SG::AuxElement::Decorator< int > isBTag( m_decor );
      if( !isBTag.isAvailable( *jet_itr ) ) {
        if( m_BJetSelectTool->accept( *jet_itr ) ) { 
          isBTag( *jet_itr ) = 1; 
          tagged = true;
        }
        else { 
          isBTag( *jet_itr ) = 0; 
          tagged = false;
        }
      }
``` 

, an event with >=2 jets that pass the BTag requirement will end up having only the first one decorated, but not the others because of the `isBTag.isAvailable( *jet_itr )` check. The final outcome is that I do not see events with more than 1 bjet, even when checking a ttbar sample!

To be honest, I haven't really understood why the decoration availability check is working that way: my guess was that SG::Decorator was a property of a single xAOD object, but apparently is not.

@kratsg, can you shed some light on this maybe?

Cheers

Marco

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -402,17 +402,15 @@ EL::StatusCode BJetEfficiencyCorrector :: execute ()
       // Add decorator for decision
       // 
       SG::AuxElement::Decorator< int > isBTag( m_decor );
-      if( !isBTag.isAvailable( *jet_itr ) ) {
-        if( m_BJetSelectTool->accept( *jet_itr ) ) { 
-          isBTag( *jet_itr ) = 1; 
-          tagged = true;
-        }
-        else { 
-          isBTag( *jet_itr ) = 0; 
-          tagged = false;
-        }
+      if( m_BJetSelectTool->accept( *jet_itr ) ) { 
+        isBTag( *jet_itr ) = 1; 
+        tagged = true;
       }
-
+      else { 
+        isBTag( *jet_itr ) = 0; 
+        tagged = false;
+      }
+ 
       // if only decorator with decision because OP is not calibrated, go to next jet
       if(!m_getScaleFactors) { continue; }
 
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -855,7 +855,6 @@ void HelpTreeBase::FillElectrons( const xAOD::ElectronContainer* electrons, cons
 
     if ( m_elInfoSwitch->m_isolation ) {
 
-      
       static SG::AuxElement::Accessor<char> isIsoLooseTrackOnlyAcc ("isIsolated_LooseTrackOnly");
       static SG::AuxElement::Accessor<char> isIsoLooseAcc ("isIsolated_Loose");
       static SG::AuxElement::Accessor<char> isIsoTightAcc ("isIsolated_Tight");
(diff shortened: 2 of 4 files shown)


mmilesi:
@jdandoy , @johnda102 

Could you also please have a look at this? I need to submit an ntuple production and I want to make sure everything will be fixed on this side. Thanks for that!

mmilesi:
Hey all,

I have unearthed an old PAT help topic, where Attila and TJ explained why doing as above is a (nasty!) mistake:

https://groups.cern.ch/group/hn-atlas-PATHelp/Lists/Archive/Flat.aspx?RootFolder=%2fgroup%2fhn-atlas-PATHelp%2fLists%2fArchive%2fIParticle%2eisAvailable%28%29%20on%20i%2b1th%20object&FolderCTID=0x0120020084D7E80CB0C3394A8D54EF07C309B044

So, the bottom line is: **never ever use isAvailable() for decorators, i.e., decorate selectively!**.

We should all check if any of such bugs is somewhere else in our code ASAP!!

Cheers

Marco 

@gfacini @kratsg @jdandoy @johnda102

jdandoy:
Great catch!  I'll search for it on my end.

johnda102:
hmm...I think we have quite a few of these: 

eg: in jet selector 

for ( auto jet_itr : *inJets ) { // duplicated of basic loop                                                                                        

    // if only looking at a subset of jets make sure all are decorated                                                                                
    if ( m_nToProcess > 0 && nObj >= m_nToProcess ) {
      if ( m_decorateSelectedObjects ) {
        passSelDecor( *jet_itr ) = -1;
      } else {
        break;
      }
      continue;
    }

mmilesi:
Hi @johnda102 ,

I have just realised that :cry: 

Also, the SF decorations might be affected...

jdandoy:
But to be clear, this only affects instances where we make a decision based on whether or not an object (jet, electron, etc.) is decorated. 

For the instance you showed John, what is the issue?  Here it looks like we decorate all the objects with a value.  Even ones beyond m_nToProcess are still sure to get the decoration of -1.

Jeff

johnda102:
Hi Jeff, 

  From reading the thread that @mmilesi  linked it sounded to me like this was a problem anytime you only decorate a sub-set of the objects in a container.  But maybe im wrong

jdandoy:
But we're not decorating a sub-set of objects in the container in that instance, we're decorating all objects in the container.  The code right after that snippet decorates the rest.  That's why the commented line says:

// if only looking at a subset of jets make sure all are decorated

johnda102:
ok yes. makes sense

kratsg:
The rule of thumb is to do the following

```
someDecor(*object) = 0;
for(auto obj: *objects){
  if(someCondition) continue;
  someDecor(*object) = 1;
}
```

One of the reasons why decorating selectively fails is because decorating a single object in a group does not set the default for other objects anymore. You now have to decorate every single object in a container to make sure they all have a value -- because `isAvailable` asks the container if it knows of a decoration of the given name and type, without actually checking if all of them had the right value set (eg: a runtime crash will occur).

PR #238, opened 2015-08-21
Fix bug in jet systematics handling

mmilesi:
Do not add an empty systematic more than once!

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -433,16 +433,6 @@ EL::StatusCode JetCalibrator :: initialize ()
     RETURN_CHECK("JetCalibrator::initialize()", m_JVTTool->initialize(), "");
   }
 
-  // if not running systematics, need the nominal
-  // if running systematics, and running them all, need the nominal
-  // add it to the front!
-  if ( m_systList.empty() || (!m_systList.empty() && m_systName == "All") ) {
-    m_systList.insert( m_systList.begin(), CP::SystematicSet() );
-    const CP::SystematicVariation nullVar = CP::SystematicVariation(""); // blank = nominal
-    m_systList.begin()->insert(nullVar);
-    m_systType.insert(m_systType.begin(), 0);
-  }
-
   for ( const auto& syst_it : m_systList ){
     Info("initialize()"," Running with systematic : %s", (syst_it.name()).c_str());
   }
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -342,12 +342,13 @@ EL::StatusCode JetCalibrator :: initialize ()
 
     //If just one systVal, then push it to the vector
     if( m_systValVector.size() == 0) {
+      if ( m_debug ){ Info("initialize()", "Pushing the following systVal to m_systValVector: %f", m_systVal ); }
       m_systValVector.push_back(m_systVal);
     }
 
     for(unsigned int iSyst=0; iSyst < m_systValVector.size(); ++iSyst){
       m_systVal = m_systValVector.at(iSyst);
-      std::vector<CP::SystematicSet> JESSysList = HelperFunctions::getListofSystematics( recSysts, m_systName, m_systVal );
+      std::vector<CP::SystematicSet> JESSysList = HelperFunctions::getListofSystematics( recSysts, m_systName, m_systVal, m_debug );
 
       for(unsigned int i=0; i < JESSysList.size(); ++i){
         m_systList.push_back(  JESSysList.at(i) );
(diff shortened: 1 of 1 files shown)


johnda102:
Merging for now. Will test later with Multijet code.  Maybe over the weekend.

btuan:
@mmilesi 

I ran into some issues running with the head of xAH on JetCalibrator as well. It seems the latest `fixJetSyst` branch can have the case where the nominal is never run -- which caused some issues for me downstream. I have my changes in the following branch, that seems to work.
https://github.com/UCATLAS/xAODAnaHelpers/tree/IncrementalFix

Best,

Brian

mmilesi:
Hi @btuan ,

I think we want to have a case when the nominal is never run: that is, when you run with a single systematic variation. This is going to be the best solution in case one wants to run multiple jobs, one per syst, hence does not need the nominal to be run on more than once.

Perhaps @gfacini and @johnda102  want to chime in...

Btw, can you point me at the relevant changes you made in your branch?

Cheers

Marco

mswiatlo:
Hi all,

I'm seeing some errors which may be related to this bug, unfortunately. 

When I run (on the master) with the following run algorithm set up in my run script:

`systName = "None"
systVal = 0.
jetCalib = ROOT.JetCalibrator()
jetCalib.setLevel(3)
jetCalib.setName( "jetCalib" ).setConfig( "$ROOTCOREBIN/data/TaggingStudies/jetCalib_AntiKt4EMTopo.config").setSyst( systName, systVal )`

I get nothing in the TStore afterwards (no jets at all). Brian's branch at least gives me jets within the TStore (JetCalib, JetCalibShallowCopy), but then crashes with:

`JetSelector::execute()    ERROR   /share/home/swiatlow/analysis/TaggingStudies/xAODAnaHelpers/Root/JetSelector.cxx:465 Failed to execute: HelperFunctions::retrieve(systNames, m_inputAlgo, 0, m_store, m_verbose)`

when we get to the JetSelector.

This was working one week ago, so I'm guessing some of these changes have caused the issues.

btuan:
@mmilesi 

The relevant changes are around [L:333-350](https://github.com/UCATLAS/xAODAnaHelpers/compare/IncrementalFix#diff-d78c6b23d6f4693e87363ade6518a855R337) where the handling of the systematic is more like the other calibrator classes. I believe if one specifies `None` or a specific systematic in the config, the nominal case should be passed over -- the `getListofSystematics` [function](https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelperFunctions.cxx#L390) only appends the nominal case when the systName variable is empty or "All". So, this should work as you would want.

In addition, I [add a criteria](https://github.com/UCATLAS/xAODAnaHelpers/compare/IncrementalFix#diff-d78c6b23d6f4693e87363ade6518a855R444) to the JER systematic handling -- the check for the nominal systematic with an empty systematic set does not work with the `.empty()` function as expected, so I add a comparison to a bare `CP::SystematicSet()`.

@mswiatlo 

Try leaving the systematic name as an empty string, i.e. `systName = ""` to run over the nominal case only -- as I suppose is your intention. As you see in the [`HelperFunctions::getListofSystematics`](https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelperFunctions.cxx#L390) that is called by the JetCalibrator, no systematic is appended in the "None" case, not even an empty systematic set.  

Best,
Brian

mswiatlo:
Same issue @btuan when I use an empty string for the `systname`.

For what it's worth, I'm using configs very similar to what you have in your tutorial.

I also upped the verbosity on the JetSelector, and this is the result:

```HelperFunctions::retri... INFO          Attempting to retrieve Jets_Calib_Algo of type std::vector<std::string>
HelperFunctions::retri... INFO                  Looking inside: xAOD::TStore
HelperFunctions::retri... INFO                  Not found at all
JetSelector::execute()    ERROR   /share/home/swiatlow/analysis/TaggingStudies/xAODAnaHelpers/Root/JetSelector.cxx:465 Failed to execute: HelperFunctions::retrieve(systNames, m_inputAlgo, 0, m_store, m_verbose)
```

kratsg:
@mswiatlo it looks like JetSelector is trying to run over systematics. Maybe that configuration is faulty.

btuan:
@mswiatlo

Ah! In my branch I have the output algo in the config file configured as
`OutputAlgoSystNames` to harmonize more with the other calibrator classes
(previously just OutputAlgo). My tutorial wasn't adapted to the unofficial
xAH/IncrementalFix branch quite yet. Perhaps this?

Best,
Brian

mswiatlo:
Yeah, it was this. I just removed the specification of the `Algo` parameters, since I don't need systematics, and things work again.

Thanks very much!

Issue #237, opened 2015-08-20
Uniforming systematics configuration between calibrators


mmilesi:
It seems JetCalibrator is quite different form the others. @gfacini could you please have a look at this?

Issue #236, opened 2015-08-19
Head of xAH crashes when running jet systematics


johnda102:
Head of xAH is broken.

The list of systemaitcs that it tries to run over is given below.. 
Looks like the Nominal is included more than once. So the code crashes. 

There were a lot of changes recently to getListOfSystematics I suspect this is where the problem is coming from.  Were these changes tested on jets ?  

JetCalibrator::initial... INFO     Running with systematic : 
JetCalibrator::initial... INFO     Running with systematic : 
JetCalibrator::initial... INFO     Running with systematic : JET_GroupedNP_1__1up
JetCalibrator::initial... INFO     Running with systematic : JET_GroupedNP_1__1down
JetCalibrator::initial... INFO     Running with systematic : JET_GroupedNP_2__1up
JetCalibrator::initial... INFO     Running with systematic : JET_GroupedNP_2__1down
JetCalibrator::initial... INFO     Running with systematic : JET_GroupedNP_3__1up
JetCalibrator::initial... INFO     Running with systematic : JET_GroupedNP_3__1down
JetCalibrator::initial... INFO     Running with systematic : 
JetCalibrator::initial... INFO     Running with systematic : JET_JER_SINGLE_NP__1up

mmilesi:
Hi @johnda102 ,

I thought the full jet part regarding syst treatment had been updated by @gfacini, but I guess he checked only the BTag section so far. I have created a remote branch:

**fixJetSyst**

Which *might* fix the problem. However, is definitely unclear to me how to configure the JetCalibrator to run on systematics...can you please let me know how you did?

Btw, it looks like the systematic settings in JetCalibrator is quite different than in the other calibrators. Is there a way to make the whole thing more uniform? E.g., where is 'm_sysName` set?

Marco

johnda102:
Hi Marco, 

   To get the systemaitcs I run with the config file given below with:

and calling 
setSyst( "All", 1.0 ); 
on the JetCalibrator

m_sysName gets set in the base class from the setSyst call above. 
Im not sure why the treatment is different here than in other classes. 
I agree we should harmonize in the future. 

config file:
InputContainer    AntiKt4EMTopoJets
JetAlgorithm      AntiKt4EMTopo
OutputContainer   AntiKt4EMTopoJets_Calib
OutputAlgo        AntiKt4EMTopoJets_Calib_Algo
Sort              true
SaveAllCleanDecisions true
# 13 TeV                                                                                                                                              
[...]

PR #241, opened 2015-09-02
Fix Jet Calibrator Systematics Handling 

btuan:
Hi All,

Just a few changes to fix the jet calibration step.
* JetCalibrator
  * Call `HelperFunctions::getListofSystematics()` like the other calibrators do.
  * Run nominal if `m_systName` is empty. If configured to be `None` or a specific systematic, do not run nominal.
  * Improve guards against duplicating nominal case.
  * **N.B. `outputAlgo` configuration changed to `outputAlgoSystNames` for uniformity. May break prior analyses if config file is not changed accordingly.**
* Other Changes
  * Add 25ns GRL to the `checkoutASGtags.py` script.

This addresses issues #236 and #237; is relevant to discussion in PR #238.

Best,

Brian

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -393,7 +393,7 @@ void HelpTreeBase::AddMuons(const std::string detailStr) {
   if ( m_debug )  Info("AddMuons()", "Adding muon variables: %s", detailStr.c_str());
 
   m_muInfoSwitch = new HelperClasses::MuonInfoSwitch( detailStr );
-
+  std::cout << "Muon info switch detail string. "<< detailStr << std::endl;
   // always
   m_tree->Branch("nmuon",   &m_nmuon, "nmuon/I");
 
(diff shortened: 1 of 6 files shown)


kratsg:
@mswiatlo - can you sign off on this?

mswiatlo:
Will look at this tomorrow

mmilesi:
Hi @btuan ,

it looks okay to me. Only a question: what's the point of having a `None` systematic value?
Shouldn't it be a redundancy for the nominal case?

Marco

btuan:
Hi @mmilesi ,

It can be easily implemented as a redundancy for the nominal case, but at present the `None` case is not being specifically handled. I think that this is likely a non-issue, since the config file recommends passing an empty string to the `OutputAlgoSystNames` parameter when running nominal only.

Best,
Brian

mmilesi:
Hi @btuan,

yes, I agree with you. Personally, I'd eliminate `None` as a possible option as there's no need to have it, but it's certianly harmless to keep it around for now. Up to you.

For me, this can be merged, so...go ahead after you have changed that little thing (if you plan to)

mswiatlo:
(Just to confirm: this did look good to me. Thanks!)

PR #246, opened 2015-09-04
fix for JetCalibration, minor change to downloader

johnda102:


gfacini:
How does that fix the problem?  Isn't the default value for m_systName already ""?

gfacini:
How does that fix the problem? Isn't "" already the defualt value of m_systName?

johnda102:
No. The m_systName gets written in the config script with setConfig method of the base class. 
This happens after the constructor, but before the config method.

Without this change we iniit systName to "", then set it, then set it back to "" .

kratsg:
I agree. :+1: Good catch John!

kratsg:
This closes #236

PR #250, opened 2015-09-07
Add protection for old btagging vars

johnda102:
Add protection for some btagging vars

PR #251, opened 2015-09-07
Add mv2xx hists

johnda102:
Adding you guessed it...Mv2xx hists


kratsg:
This will close #250 because it contains that commit.

PR #255, opened 2015-09-17
Reverting outputAlgoSystNames back to outputAlgo, and allowing for ba…

jdandoy:
See thread: https://github.com/UCATLAS/xAODAnaHelpers/issues/236


mmilesi:
All good

PR #256, opened 2015-09-18
Merging version of py config

johnda102:


kratsg:
Two main comments:

1) we can write `xAH_config` a lot nicer if we create a class instead. This way, we can actually introspect the local objects created and only select the ones of the specific class: http://stackoverflow.com/a/4458733/1532974

```python
class xAH_config(object):
  def __init__(self, job):
    self.__job = job
    self._algorithms = []

  def __setattr__(self, alg_name, configs):
    alg = reduce(lambda x,y: getattr(x, y, None), alg_name.split('.'), ROOT)
    if alg is None:
      raise AttributeError(item)
    alg = alg()
    for k,v in configs.iteritems():
      setattr(alg, k, v)
```

Now, at this point, all you have to write in your python file is something a little more equivalent to this:

```
c = xAH_config(Job)
c.BasicEventSelection = {"debug": True, ....}
c.JetSelector = {"debug": False, ...}
```

johnda102:
Hi Giordon, 

    for 1) The python is very dense so Im not 100% sure what the code does.  Is there any fundamental difference to what is there now ? or is this just a rewite to avoid the config_* wrappers. 
  If so I agree we should move to it but with a more human readable version.

   for 2) a crude version of this is already being done in setMemberFunctions. 
      ```python
      print "Setting" , user_opt, "to", config[user_opt]
     ```
  of course this could be beautified but the functionality is in principle there.

kratsg:
> For 1, the python is very dense so Im not 100% sure what the code does. Is there any fundamental difference to what is there now ? or is this just a rewite to avoid the config_* wrappers. 
> If so, I agree we should move to it but maybe with a more human readable version.

Right, so this is just a general class that allows the user to, for example, specify multiple sets of configs/algorithms. This can end up being quite flexible because we can automatically create multiple algorithms from the same file (for example). One of the reasons for doing it this way is that it avoids, as you said, hardcoding the config wrappers - so if we added a new algorithm or changed the name, or the user wants to use one of their algorithms -- they can do so without having to write a similar custom function for just that algorithm. This will automatically find the algorithm and spin it up. For the user, it probably looks a bit cleaner.

Secondly, because of the `setattr()` call example, you can now make a list of algorithms and automatically loop over them like so:

```python
c = xAH_config(job)
for jet in ['AntiKt4LCTopoJets', 'AntiKt10LCTopoJets', 'AntiKt4TruthJets']:
  for alg in ['JetSelector', 'JetCalibrator', 'JetHistsAlgo']:
    setattr(c, alg) = {..., 'InputContainer': jet, ...}
```

which creates 9 algorithms in a relatively clean code as well. You can't very easily call custom function names (you can, but the syntax in python is a bit more confusing than the above).

Third, it does look a little dense in python, but it's only particularly overloading some special function calls: `c.attribute` automatically calls the underlying `__setattr__` function after calling the `__getattr__` function to make sure the attribute exists. In this case, the `__getattr__` will check to see if the attribute you're looking up exists on the class first, before checking if it's in the `ROOT` namespace (eg: algorithms). The `__setattr__` will check to see which attribute you're trying to set, and all other ones it will append to a list of configured algorithms (a large chunk of that code already existed in `xAH_run.py`).

That whole file definition automatically does all the grunt work and dynamic calls for you, so all the user has to worry about is actually writing code like

```
c = xAH_config(Job)
c.BasicEventSelection = {"debug": True, ....}
c.JetSelector = {"debug": False, ...}
c.AnotherCustomAlgorithm = {... other configs ...}
setattr(c, "Namespace.Algorithm") = {... more configs ... }
```

and the `xAH_run.py` can introspect and find out all the `xAH_config`s being used, and just... use them.

kratsg:
> for 2) a crude version of this is already being done in setMemberFunctions. 
> python print "Setting" , user_opt, "to", config[user_opt] 
> of course this could be beautified but the functionality is in principle there.

I'll try to sync it up later today after I land in Boston.

johnda102:
ok sounds good

johnda102:
I propose we merge and have a bug-report to update the xAH_config to the new version.

kratsg:
Feel free to file an issue, label as "enhancement", reference #256 and then make it my responsibility perhaps. If it works as is, great!

kratsg:
We just need to be careful to make the changes so we nail down the format before too many people start getting adjusted.

johnda102:
ok ill try to migrate tonight.

johnda102:
Why is 

```python 
 if name in ['_job', '_algorithms']:
      super(xAH_config, self).__setattr__(name, val)
```

needed?

kratsg:
Because of:

```python
  def __init__(self, job):
    self._job = job 
    self._algorithms = []
```

Those are calls to `setattr` -- and they're special. Meaning calls with those names `_job` and `_algorithms` should use the default call `super(xAH_config, self).__setattr__(name, val)` which invokes the parent method (default method) rather than the custom method for adding algorithms.

johnda102:
OK another one I dont understand:

```python 
      alg = reduce(lambda x,y: getattr(x, y, None), name.split('.'), ROOT)
``` 
whats going on here ?

kratsg:
Namespaces. C++ handles namespaces well, but python doesn't have an equivalent. The problem ends up when you try to do something like:

`ROOT.MyNamespace.MyAlgorithm` -- Python will attempt to chain it from left to right, looking for `MyNamespace` under `ROOT` and then `MyAlgorithm` under the returned object. In order to grab the algorithm correctly in a dynamic way, it's easier if you do:

`setattr(c, "MyNamespace.MyAlgorithm")` which will grab `ROOT.MyNamespace.MyAlgorithm` for you. That's a really magical one-liner I've done for years. It's equivalent to:

`getattr(...getattr(getattr(getattr(ROOT, x1), x2), x3), ...)`

where `x1 = MyNamespace` and `x2 = MyAlgorithm` and so on. We're just splitting up the periods so we decompose the chain correctly if a user wraps their code in a namespace. It works in JSON as well.

I have a brief blurb here: https://github.com/UCATLAS/xAODAnaHelpers/wiki/xAH_run.py#getting-started

> so that one could create an algorithm by something like `ROOT.AlgorithmName()` and then start configuring it. In fact, this is how one normally does it within python. Namespaces are automatically linked up by something like `ROOT.Namespace.AlgorithmName()` in case you wrapped the entire algorithm in a namespace.

kratsg:
Note: a future version of the code that I'll write will probably allow the user to write `c.MyNamespace.MyAlgorithm` instead. I have a way to do it with meta-classing which basically overrides the type of object being used to have a custom `setattr` that gets used. I'll add it when I fix some recursion bugs.

johnda102:
ok fine, 

but after you initilaize alg you override it 3 lines later.  So is this only needed to test that the Class is in the ROOT namespace ?

kratsg:
Oh, so that's what is called a "method-proxy". EG:

```
def add(x, y):
  return x+y

def subtract(x, y):
  return x-y

def multiply(x, y):
  return x*y;

for alg in [add, subtract, multiply]:
  print alg(10, 9)
```

I can use variable-named functions. When I write the call `alg = alg()` - this creates an algorithm object of the given algorithm passed in, initialize it with all the parameters/configurations, and add it to the (ordered) list of algorithms for the `xAH_config` object.

kratsg:
Here's an updated version of the code without the `job` and also an easy way to access the algorithms

```python
import ROOT

ROOT.gROOT.Macro("$ROOTCOREDIR/scripts/load_packages.C")

class xAH_config(object):
  def __init__(self):
    self._algorithms = []

  def __setattr__(self, name, val):
    if name in ['_job', '_algorithms']:
      super(xAH_config, self).__setattr__(name, val)
    else:
      alg = reduce(lambda x,y: getattr(x, y, None), name.split('.'), ROOT)
      if alg is None:
        raise AttributeError(name)
      alg = alg()
      for k,v in val.iteritems():
        if not hasattr(alg, k):
          raise AttributeError(k)
        setattr(alg, k, v)
      self._algorithms.append(alg)

  def __getattr__(self, name):
    if hasattr(self, name):
      return super(xAH_config, self).__getattr__(self, name)
    else:
      return getattr(ROOT, name)

  @property
  def algorithms(self):
    return self._algorithms

# c = xAH_config()
```


which can be used for example like the following:

```
kratsg@tier3:/share/home/kratsg/RestFramesAnalysis$ python
Python 2.7.4 (default, May  5 2013, 17:33:10) 
[GCC 4.8.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from test import *

RestFrames v1.0.0 -- Developed by Christopher Rogan (crogan@cern.ch)
                     Copyright (c) 2014-2015, Christopher Rogan
                     http://RestFrames.com

Info in <xAOD::Init>: Environment initialised for data access
>>> c = xAH_config()
>>> c.JetSelector = {'m_debug': True}
Info in <JetSelector::JetSelector()>: Calling constructor
>>> c.algorithms
[<ROOT.JetSelector object at 0xcca34a0>]
```

johnda102:
After a profanity laden discussion with goirdon offline we have come to a consensus on what needs done. 

Implemented now.  More details when tested and ready.

johnda102:
ok @kratsg Ive put in a first pass Let me know if you see something I missed. 
(Namespaces are not working right)

johnda102:
Pulling into master. 

Bottomline: 
  xAH_run now supports both json files and pyconfig files. 

python config files create an object of type xAH_config and add algorithms to it via the member function 
setalg.  

This takes two arguements the first is the name of the c++ class being configured (nested namespaces are with a "." eg: XhhResolved.SetXhhEventCuts) and the second argument is a dictionary specifing the configuration. 

This is run with xAH_run as you would do with the json files but giving the python config instead. 
A few example scripts from Multijets and X->hh->4b are given below.

MultiJets:

```python
import ROOT
from xAH_config import xAH_config

c = xAH_config()

#
#  Data Config
#
if not args.is_MC:
    leadingJetPtCut    = 10*1000
    TruthContainerName = "None"
    doPUreweighting    = False

#
#  MC Config
#
else:
    leadingJetPtCut    = 40*1000
    TruthContainerName = "AntiKt4TruthJets"
    doPUreweighting    = True        

if args.is_AFII:
    setAFII         = True
    JESUncertMCType = "AFII"
else:
    setAFII         = False
    JESUncertMCType = "MC15"

#
# Basic Setup 
#
triggers = "HLT_j.*|HLT_ht.*|L1_J.*|L1_HT.*|HLT_noalg_L1J.*"
c.setalg("BasicEventSelection",  { "m_name"                : "basicEventSel",
                                   "m_applyGRLCut"         : False,
                                   "m_derivationName"      : "EXOT1", 
                                   "m_storePassL1"         : True,
                                   "m_storePassHLT"        : True,
                                   "m_storeTrigDecisions"  : True,
                                   "m_storeTrigKeys"       : True,
                                   "m_applyTriggerCut"     : True,
                                   "m_triggerSelection"    : triggers, 
                                   "m_doPUreweighting"     : doPUreweighting, 
                                   "m_lumiCalcFileNames"   : "$ROOTCOREBIN/data/MultijetAlgo/ilumicalc_histograms_None_267073-271048.root",
                                   #"m_PRWFileNames"        : "$ROOTCOREBIN/data/MultijetAlgo/Multijet_PRW_25ns.root",
                                   "m_PRWFileNames"        : "$ROOTCOREBIN/data/MultijetAlgo/Multijet_PRW.root",
                                   "m_PU_default_channel"  : 302590
                                   }
         )
#
#  Jet Calibration
#
c.setalg("JetCalibrator", { "m_name"                   : "jetCalib_AntiKt4TopoEM", 
                            "m_inContainerName"        : "AntiKt4EMTopoJets",
                            "m_jetAlgo"                : "AntiKt4EMTopo",
                            "m_outContainerName"       : "AntiKt4EMTopoJets_Calib", 
                            "m_outputAlgo"             : "AntiKt4EMTopoJets_Calib_Algo",
                            "m_sort"                   : True,
                            "m_saveAllCleanDecisions"  : True,
                            "m_calibConfigAFII"        : "JES_MC15Prerecommendation_AFII_June2015.config",
                            "m_calibConfigFullSim"     : "JES_MC15Prerecommendation_April2015.config",
                            "m_calibConfigData"        : "JES_MC15Prerecommendation_April2015.config",
                            "m_calibSequence"          : "JetArea_Residual_Origin_EtaJES_GSC",
                            "m_JESUncertConfig"        : "$ROOTCOREBIN/data/JetUncertainties/JES_2015/Prerec/PrerecJES2015_3NP_Scenario1_25ns.config",
                            "m_JESUncertMCType"        : JESUncertMCType,
                            "m_setAFII"                : setAFII,
                            "m_JERUncertConfig"        : "JetResolution/Prerec2015_xCalib_2012JER_ReducedTo9NP_Plots_v2.root",
                            "m_JERApplyNominal"        : False,
                            "m_redoJVT"                : False,
                            "m_systName"               : "All", 
                            "m_systVal"                : 1,
                            } )


#
#  Jet Selection
#
c.setalg("JetSelector", { "m_name"                    :  "JetSelector",
                          "m_inContainerName"         :  "AntiKt4EMTopoJets_Calib",
                          "m_inputAlgo"               :  "AntiKt4EMTopoJets_Calib_Algo",
                          "m_outContainerName"        :  "SignalJets",
                          "m_outputAlgo"              :  "SignalJets_Algo",
                          "m_decorateSelectedObjects" :  True, 
                          "m_createSelectedContainer" :  True, 
                          "m_cleanJets"               :  False, 
                          "m_pT_min"                  :  50e3,
                          "m_eta_max"                 :  2.8,
                          "m_useCutFlow"              :  True,
                          "m_doJVF"                   :  False,
                          } )


#
# Ntuples
#
c.setalg("MultijetAlgorithm", { "m_name"                   : "MultijetAlgo",
                                "m_debug"                  : False, 
                                "m_truthLevelOnly"         : False, 
                                "m_isMC"                   : args.is_MC,
                                "m_writeTree"              : True, 
                                "m_inContainerName"        : "SignalJets",
                                "m_inputAlgo"              : "SignalJets_Algo",
                                "m_leadingJetPtCut"        : 40000, 
                                "m_MCPileupCheckContainer" : TruthContainerName, 
                                "m_eventDetailStr"         : "pileup", 
                                "m_jetDetailStr"           : "kinematic rapidity clean energy  layer scales trackPV",
                                "m_trigDetailStr"          : "passTriggers basic menuKeys", 
                                } )
```

From X->hh->4b 
``` python 
import ROOT
from xAH_config import xAH_config
import sys, os

sys.path.insert(0, os.environ['ROOTCOREBIN']+"/user_scripts/XhhCommon/")

from XhhCommon_config   import *
from XhhResolved_config import *
from XhhBoosted_config  import *

c = xAH_config()


#
# Basic Setup 
#
triggers = "L1_4J20,L1_J100,HLT_j400,HLT_4j100,HLT_ht1000,HLT_j75_bmedium_3j75,HLT_j175_bmedium_j60_bmedium,HLT_j300_bloose,HLT_2j55_bmedium_2j55,HLT_2j45_btight_2j45,HLT_2j75_bmedium_j75,HLT_j100_2j55_bmedium,HLT_j460_a10_lcw_sub,HLT_e24_lhtight_iloose,HLT_mu26_imedium,L1_3J25.0ETA23,L1_4J15.0ETA25"
c.setalg("BasicEventSelection", { "m_name"                  : "basicEventSel",
                                  "m_applyGRLCut"           : False,
                                  "m_doPUreweighting"       : False, 
                                  "m_vertexContainerName"   : "PrimaryVertices",
                                  "m_PVNTrack"              : 2,
                                  "m_truthLevelOnly"        : False,
                                  "m_applyPrimaryVertexCut" : True,
                                  "m_derivationName"        : "EXOT8", 
                                  "m_triggerSelection"      : triggers, 
                                  "m_storeTrigDecisions"    : True,
                                  "m_useMetaData"           : False,
                                  "m_applyEventCleaningCut" : True,
                                  "m_applyTriggerCut"       : False,
                                  "m_storePassL1"           : True,
                                  "m_storePassHLT"          : True,
                                  "m_storeTrigKeys"         : True,
                                  } )



c.setalg("MuonCalibrator", { "m_name"                : "Muons", 
                             "m_inContainerName"     : "Muons", 
                             "m_outContainerName"    : "Muons_Calib", 
                             "m_outputAlgoSystNames" : "MuonCalibrator_Syst"
                             } )

#
#  Set up resolved Analysis
#
config_ResolvedAnalysis(c, args)

#
#  Set up Boosted Analysis
#
config_BoostedAnalysis(c, args)

#
# Ntuples
#
c.setalg("XhhMiniNtuple", { "m_name"                : "MiniNTuple",
                            "m_inCaloJetName"       : "selCaloJets",
                            "m_inTrackJetName"      : "GhostAntiKt2TrackJet",
                            "m_inJetName"           : "AntiKt4EMTopoJets_Calib_preSel",
                            "m_inBJetName"          : "AntiKt4EMTopoJets_Calib_btagged",
                            "m_inDiJetName"         : "PreDiJets",
                            "m_jetDetailStr"        : "kinematic truth flavorTag",
                            "m_trigDetailStr"       : "passTriggers",
                            "m_inTruthParticleName" : "TruthParticles",
                            } )
```

PR #261, opened 2015-09-22
Skip empty lines in the input file

johnda102:


kratsg:
@johnda102, that won't work. Some systems put a carriage return and some don't:

```
if not line.strip():
  # do something
```

or

```
if not line in ['\r', '\r\n']:
  # do something
```

alexandertuna:
+1 if not line.strip():

johnda102:
done

PR #265, opened 2015-09-23
Don't require inputDQ2 when using condor

### LINKED ISSUE #262: xAH_run Assumes DQ2 With Condor
The xAH_run.py scripts assume that the input files come from DQ2 when running on condor. This is not always true. For example, I have a set of private samples on faxbox which manage using txt filelists (ie --inputList option).

kratsg:
This closes #262. Can @kkrizka :+1: it?


kkrizka:
Seems good.

PR #272, opened 2015-09-30
Make single event weight for object SFs

mmilesi:
Before merging this, I'd like to discuss with you about one point.

This PR adds the following feature:

```
Save an **event** weight in the output tree which represents, for a given SF type (e.g., BTag efficiency SF, electron reconstruction efficiency SF...) , the product of the SFs for each object in the event.
```

Th reason why I wish to add this is b/c when plotting with a TTree::Draw() like cut string, it can be bothersome to include the SF weight for each object, e.g. for the BTag SF:

```
physics->Draw("my_var", " MY_OTHER_EVENT_WEIGHTS * jet_MV2c20_SFFix70[0][0] * jet_MV2c20_SFFix70[1][0] * jet_MV2c20_SFFix70[2][0] * (njets == 3 && MY_OTHER_CUTS)")
```
or even impossible, in the case where you don't know exactly the number of objects you are going to have in your plot such as for an inclusive jet selection:

```
physics->Draw("my_var", " ??? * (njets > 2 && MY_OTHER_CUTS)")
```

At the moment, I am calculating the "global" product of the SFs in each `*EfficiencyCorrector.cxx` module, which of course is the most natural choice, save it as an **event** decoration, and finally retrieve it in `HelpTreeBase.cxx` and dump it in the ntuple along with the per-object weights as always done.

However, there's a caveat here: if the user calls for the  `*EfficiencyCorrector.cxx`  algorithms **AFTER** the object has been defined - that is, after all the selectors AND the overlap removal - then the global weight will be correct. On the contrary, if the user redefines the objects after having called the corrector algorithms, then the global weight will be wrong. 

One option could be to calculate the product weight in  `HelpTreeBase.cxx`, since it's by all means the last algorithm to be called in the chain. However, I am not very happy with that, since it will bloat the tree code additionally, and I think it safer just to tell the user to correct the objects with SFs always **after** the object final definition. Something like this:

```
    job.algsAdd( baseEventSel );
    job.algsAdd( jetCalib );
    job.algsAdd( jetSelect );
    job.algsAdd( muonCalib ); 
    job.algsAdd( muonSelect);
    job.algsAdd( electronCalib );
    job.algsAdd( electronSelect );
    job.algsAdd( overlapRemoval );
    job.algsAdd( bjetEffCorr_BTag_MV2c20_Fix70 );
    job.algsAdd( muonEffCorr );
    job.algsAdd( electronEffCorr );
    job.algsAdd( met );
    job.algsAdd( analysis );
    job.algsAdd( out_tree );

```

What do you think?

@jdandoy @johnda102 , @gfacini 

Cheers

Marco


johnda102:
Hi Marco, 

  Looks fine to me.  

On the issue of global event weights, given that it so analysis dependent I always assumed the best place to calculate and write that out would be in the derived Ntuple class specific to the analysis. 

That said I havent really had to deal with SF in xAH yet.

mmilesi:
Hi @johnda102 ,

thanks for the comment! 

Ok, let's wait also for @gfacini  and @jdandoy to comment on this.

In any case, I just stress that my addition does not interfere with the current way we store the SFs whatsoever (it's just a more handy way of doing it :wink: , and provided that the users do not correct efficiencies for the objects before the final selection - which makes sense to me - it should be also independent on the analysis, thanks to the high modularity of the xAH algorithms.)

Marco

jdandoy:
Hi Marco,

Thanks for starting to look into this, it may be useful to have a smart way
to calculate a global weight that can be propagated through the
algorithms.  I think there are merits to this strategy, but I can't recall
them at the moment.

As John says, many of the jet analyses calculate their global weight in
their specific algorithm, for example DijetResonanceAlgo defines the event
decoration "weight" as the combination of all relevant weights, and adds it
to the tree.  This strategy allows each analysis to define exactly which
weights they want to apply, and when they want to choose them (usually at
the end of all the selections, as you mentioned).  For this it is essential
to save individual SFs for each object upstream (as is done in b-tagging).


Jeff

gfacini:
Hi Marco, John,

I agree to some degree with John that the user
should just build this weight themselves.

The idea of building the event weight in each
efficiency corrector is nice. If things are setup
properly, that works just fine as you point out.

I can be messed up, but in something as flexible
as this FW, not everything can be fool proof.

Cheers,
Gabriel

mmilesi:
Thanks @jdandoy for the comment!

Do you agree with me that the **efficiency** corrections should be applied **after* the final object selection? Is there any use case where this will not happen? If it's not the case, then the way I am doing should be safe.

Marco

mmilesi:
Thanks @gfacini too!

I think that if users want to re-derive their combined weight, then no problem.
But I do think that having xAH already doing that for you (provided you use it wisely, but that holds in any case!) is a good plus

mmilesi:
Hi @johnda102 , @jdandoy , @gfacini 

If there is no *strong* argument against this PR, I'd like to go ahead and merge. 

This in any case should not break what other users might be already doing (the only potential issue I see is if the names of the tree branches that I chose are the same assigned by the users...can you guys check please?), and I also plan to add some lines in the documentation explicitly saying that - in order for the global weights to make sense - the SF modules of xAH must be called after the final selection.

If we want to change/improve things, it might be done at a later stage.

Thanks!

Marco

mmilesi:
Hi @gfacini , @johnda102 , @jdandoy 

is it ok to merge this?

Issue #273, opened 2015-10-02
overlap removal: don't run nominal of nominal's not there.


JamesSaxon:
If one runs 

jcal->setName("jetCal")->setConfig(rcdir + "VBFInv/jetCal.config" )->setSyst( "SingleSyst", 1 );

the algo just provides the positive and negative variations of the jet calibration.

One then threads the systematics through to OverlapRemover.cxx, where at L325 nominal runs regardless.  This causes a crash if one tries to run a single systematic.  Presumably if one does this, only the variations are desired.  Could we make so that it only runs if it can (i.e., nominal exists)?  

Thanks -- 

Jamie

mmilesi:
Hi @JamesSaxon ,

I am having a look at this. Will get back to you soon!

mmilesi:
I have opened a PR #275 for this

mmilesi:
@kratsg, this one refers to #275 (not #274). So this is not to be closed

mmilesi:
This now is covered by #287

PR #274, opened 2015-10-04
updates to METConstructor for Alex and Marco

JamesSaxon:
Hi Marco, all -- 

Per private mail, I've checked a few things in the METConstructor.

  - The JVT treatment seems to be correct: it is false by default, but I have made it configurable (DoJVFCut).  The name of the decoration, "Jvt," matches what is done elsewhere in xAH.
  - I've added the possibility to specify the MuonEloss options (DoMuonEloss, DoIsolMuonEloss), though there seems to be a bug in METMaker when MuonEloss is true and IsolMuonEloss is false.  Rui and I are in touch with TJ et al.
  - In the [Twiki](https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/METUtilities#METMaker_For_MC15_Data15), it says that it "may be wiser" to only run one type of soft term, but the [example](https://svnweb.cern.ch/trac/atlasoff/browser/Reconstruction/MET/METUtilities/trunk/util/example_METMaker_advanced.cxx#L205) does both and this seems more straightforward to me.
  - Default options have been set so that the configuration only *updates* based on the variables, as is done for the other modules.

Marta (in the email) has said she is working on the systematics (fixing was a grid crash), so we can follow up with her in that thread.

Let me know if you have any other questions or if this gives you any trouble -- 

Jamie

### DIFF
--- a/Root/METConstructor.cxx
+++ b/Root/METConstructor.cxx
@@ -51,7 +51,31 @@ using std::cout; using std::cerr; using std::endl;
 ClassImp(METConstructor)
 
 
-METConstructor :: METConstructor () : m_metmaker(0) {}
+METConstructor :: METConstructor () : m_metmaker(0) {
+
+  m_debug           = false;
+  m_referenceMETContainer = "MET_Reference_AntiKt4LCTopo";
+
+  m_mapName         = "METAssoc_AntiKt4LCTopo";
+  m_coreName        = "MET_Core_AntiKt4LCTopo";
+  m_outputContainer = "NewRefFinal";
+
+  m_inputJets       = ""; 
+  m_inputElectrons  = ""; 
+  m_inputPhotons    = ""; 
+  m_inputTaus       = ""; 
+  m_inputMuons      = ""; 
+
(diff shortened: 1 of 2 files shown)


kratsg:
go ahead

mmilesi:
Hi @kratsg 
is there any way I can add commits to this?

kratsg:
@mmilesi The easiest way in this case, is to let @JamesSaxon do a `git cherry git-sha-id` of the commits that need to be applied from another branch on here, to the master of his fork.

Otherwise, you have to make a new PR and do it yourself:

```
git fetch origin pull/274/head:pr274
git cherry commit1 commit2 ...
git push -u origin pr274
```

like so. The first command takes the pull request 274 (this one) and puts it into a branch named `pr274` locally.

mmilesi:
Thanks @kratsg, maybe the best solution is to merge this one, and then I'll create a new PR with some additions.

kratsg:
Does this close #273 completely? (regardless of your other changes to be made?)

PR #275, opened 2015-10-05
Fix for mssing nominal container in OverlapRemoval

mmilesi:
Hi @JamesSaxon,

please test this PR (just fetch the branch locally and test it with your configuration). If everything is fine for you, just go ahead and merge it :wink: 

Marco

### DIFF
--- a/Root/OverlapRemover.cxx
+++ b/Root/OverlapRemover.cxx
@@ -293,7 +293,7 @@ EL::StatusCode OverlapRemover :: initialize ()
   //
   const std::string selected_label = ( m_useSelected ) ? "passSel" : "";  // set with decoration flag you use for selected objects if want to consider only selected objects in OR, otherwise it will perform OR on all objects
   RETURN_CHECK( "OverlapRemover::initialize()", m_overlapRemovalTool->setProperty("InputLabel",  selected_label), "");
-  //RETURN_CHECK( "OverlapRemover::initialize()", m_overlapRemovalTool->setProperty("OverlapLabel", "overlaps"), ""); // tool will decorate objects with 'overlaps' boolean if they overlap (not possible to customise name atm!)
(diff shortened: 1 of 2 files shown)


mmilesi:
@JamesSaxon

to fetch the branch on your local area:

```
git fetch origin
git checkout --track origin/FixNominalSysOverlapRemoval
```

After that, you can add commits to it! Once you're set, do:

```
git push -f origin FixNominalSysOverlapRemoval
```

and this PR will be automagically updated

mmilesi:
Ok, I think I have seen where the problems stem from. I am making some changes now

mmilesi:
This is strictly related to #287

mmilesi:
Hi @JamesSaxon

do you think we can merge this? Did you have any chance to make a final test on it?

Cheers

Marco

JamesSaxon:
Hi @mmilesi  -- 

This works as expected for me.  I continue to find it convoluted that _every_ downstream algorithm should take in three sets of systematics and figure out the overlapping set.  

To me a simple example of this is that if I run a single electron systematic (jets and muons nominal), then the nominal jet container no longer exists post OR.  So when I retrieve the jet syst vector from the selector tool, I have to check that muons and electrons have "nominal" defined.

And in principle I have to do that (or use a helper) for _every_ algorithm post overlap.  But, I think we both understand the sides of this at this point.  If anyone else has an opinion here, it would be great to chime in.

Thanks so much Marco, for doing all of the work on this.

Best -- 

Jamie

mmilesi:
Hi @JamesSaxon ,

thanks for the comments, I definitely agree with you that there's plenty of room for improvement of the logic of the overlap removal machinery. For this reason, I'd encourage to make a new PR soon with updates on this!

For the time being, given this PR - although non optimal -is  fixing some pretty serious bugs, I'd go ahead and merge it. Do you agree?

Marco

JamesSaxon:
Yep, sounds great!!

mmilesi:
Merging this PR. This closes #287

PR #279, opened 2015-10-06
Always include lumiblock in ntuples

alexandertuna:
This is a small proposal to always include the lumiBlock in HelpTreeBase, same as runNumber or eventNumber. It's useful for stuff beyond pileup reweighting, like comparing against GRLs, for example.


kratsg:
go ahead

PR #281, opened 2015-10-12
METConstructor Update

JamesSaxon:
Tau default cuts updated and return successfully if no config given.


kratsg:
go ahead

JamesSaxon:
```git tag -a 00-03-21 -m "METConstructor: improved tau 'courtesy' function and better python matching"```

PR #295, opened 2015-10-20
TruthSelector now uses cutflow_truths_1 for object cutflow.

kkrizka:
The TruthSelector was using the cutflow_jets_1 histogram, which is also used by JetSelector. I changed it to use cutflow_truths_1, since it is a cutflow of a different collection.

There is still a problem with two separate TruthSelector algorithms using the same cutflow histogram, but I'll make another bug report for it.


gfacini:
ok to test

PR #299, opened 2015-10-22
Make sure we don't always print out the TStore!

mmilesi:


kratsg:
:laughing:

mmilesi:
@kratsg , would you like to have the honour of signing this milestone PR off?

kratsg:
:beer:

PR #301, opened 2015-10-24
make bjet operating point public (accessible in python)

JamesSaxon:
Title says it all, one line change works fine.


kratsg:
go ahead

PR #302, opened 2015-10-25
Set Muon quality independent of config file

johnda102:


johnda102:
@mmilesi Have a look at this PR.  Its a fix to use the quality enum function outside of the confuguration from .config file. This is needed for us are we are using the python config which sets the member variables directly.

mmilesi:
Looks good! I am also going to use python to configure the algos, so thanks for checking this.

PR #307, opened 2015-10-27
Small fix in reading of MetaData

mmilesi:
Just a minor fix to make sure we cope with:

https://twiki.cern.ch/twiki/bin/view/AtlasProtected/AnalysisMetadata

and 

https://its.cern.ch/jira/browse/ATLASG-242


kratsg:
Hi @mmilesi can you explain this fix?

PR #314, opened 2015-10-29
Change type of SG::Accessor "PileupWeight" from double to float

mmilesi:
Why the hell did they change it in 2.3.32??

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -185,7 +185,7 @@ void HelpTreeBase::FillEvent( const xAOD::EventInfo* eventInfo, xAOD::TEvent* /*
 
     if ( m_isMC ) {
 
-      static SG::AuxElement::ConstAccessor< double > weight_pileup ("PileupWeight");
+      static SG::AuxElement::ConstAccessor< float > weight_pileup ("PileupWeight");
       static SG::AuxElement::ConstAccessor< float >  correct_mu("corrected_averageInteractionsPerCrossing");
       static SG::AuxElement::ConstAccessor< unsigned int > rand_run_nr("RandomRunNumber");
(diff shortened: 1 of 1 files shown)


johnda102:
does this break compatibility before 2.3.32 ?

mmilesi:
Hi John,

Most likely, yes. However, the problem seems to be with the
PileupReweightingTool that comes with 2.3.32 which changed something
internally.

Maybe a problem to raise at PAThelp...

mmilesi:
Hi @johnda102 

Look at here:

https://twiki.cern.ch/twiki/bin/view/AtlasProtected/AnalysisBaseReleaseNotes23#ReleaseNotes32

Checking out an older tag of the PRW tool via the python script is dodgy to me, since there might have been important changes in the latest version.

The only option to preserve backwards compatibility would be to add the latest tag of PRW tool in the list of packages to checkout for the previous releases (I guess no one is using releases older than 2.3.28, so that would be just three lines of code to add to the python script)

What shall we do?

Marco

johnda102:
im still in 2.3.25. I say lets merge. its a simple patch if it breaks prior to 2.3.28. 

All agree ?

mmilesi:
Hi @johnda102 ,

I can put the latest tag of PRW in the check out list for all the releases, that's no problem at all. Of course assuming the latest tag of PRW is backward compatible with all the older packages in the releases...

@gfacini @jdandoy @kratsg

kratsg:
@johnda102 I would suggest bumping up to the latest version xAH supports. We have the condor driver for tier3 and connect supported pretty much as well as major changes to shallow copies that allow for mini-xAODs to be produced correctly.

johnda102:
ok so we agree to merge?

@kratsg updating is a question of priority. Stability is paramount given that the code is used in an analysis that is in collaboration review now

Issue #330, opened 2015-11-05
Migrate the wiki over to code documentation


kratsg:
I'll take care of this - it won't be restrictively difficult and we already have a halfway decent wiki full of documentation. It will be great to move it all into one location :)

jdandoy:
@kratsg happy with this?

kratsg:
I think so -- https://github.com/UCATLAS/xAODAnaHelpers/wiki I don't see anything on there that needs to move over?

PR #335, opened 2015-11-09
Docs/update

kratsg:
A ton of changes for #330. Most of the wiki is now migrated except for a few files. Will migrate these over into code rather than specific pages most likely.

What do we do about the ASG-Releases page? It's not very well-maintained and hasn't been updated in quite a while...


kratsg:
HUUUUUGE update. Hopefully docs should generate a.ok.

PR #342, opened 2015-11-10
Info Switch Matching needs to be tighter.

### LINKED ISSUE #341: Tighten Info switch matching.
Im starting to add HLT Bjet support to xAH and ran into a subtle issue with the info switch parsing that others may also come across. 

Accessing the BTagging info is different for HLT jets than for offline jets.
I added a flag to info switch to tell the Jet plot maker which acessor to use. 

I tried naming it "flavorTagHLT" and "HLTflavorTag"  but it turns out that with our current parse function: 

```c++
bool InfoSwitch::parse(const std::string flag)
  {
    return m_configStr.find(flag) != std::string::npos; // not equal to npos means it is found                                                             
  }
```

either of these will also match the offline switch "flavorTag" .

Seems to me we should tighten the matching criteria in parse, so the full space or comma separated token should match. 

What do others think ?

This could break backward compatibility if you're using a silly detail string.

kratsg:
Should resolve #341.

### DIFF
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -85,11 +85,6 @@ namespace HelperClasses{
    *  turn a set on
    *
    **************************************/
-  bool InfoSwitch::parse(const std::string flag)
-  {
-    return m_configStr.find(flag) != std::string::npos; // not equal to npos means it is found
-  }
-
   void EventInfoSwitch::initialize(){
     m_pileup        = parse("pileup");
     m_shapeEM       = parse("shapeEM");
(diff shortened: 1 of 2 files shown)


kratsg:
@johnda102 go ahead and test this version :)

johnda102:
on it

On Tue, Nov 10, 2015 at 7:26 PM, Giordon Stark <notifications@github.com>
wrote:

> Assigned #342 <https://github.com/UCATLAS/xAODAnaHelpers/pull/342> to
> @johnda102 <https://github.com/johnda102>.
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/342#event-460428479>.
>



-- 
University of Chicago

johnda102:
looks good

PR #344, opened 2015-11-11
Revert "Add hlt jets"

jdandoy:
Reverts UCATLAS/xAODAnaHelpers#343

@johnda102  @kratsg This update breaks the trunk for standard ("jet") running.  No jet_pt is saved to the TTree output with this commit.


johnda102:
This commit doesnt touch those files.
I guess its either an earlier problem or a later one.  
Ill what things look like from teh multijet side

jdandoy:
Seems to work fine for John, but caused Dijet errors.  Will investigate this further from my side soon, and check with John before any action is taken.

kratsg:
Hi @jdandoy -- before we start reverting, can you do the following?

```
git reset --hard dccd2af2
```

and see if it still works. (You will need an `rc clean` since it removes the header files). I'm also confused about why `jet_pt` seems missing since I don't see any of John's changes touching it.

mmilesi:
@johnda, @jdandoy 

I am also observing something like what Jeff says when looking at master. I am looking into this as well

kratsg:
Oh you know what? It might be related to the change in #342 -- but that only makes things tighter while preserving backwards compatibility. My next question would be -- what is your detail string when trying to do the kinematics?

mmilesi:
This is what I am using:

"kinematic energy flavorTag sfFTagFix77 truth"
and atm only the branches corresponding to:

"energy flavorTag"

gets dumped

kratsg:
Closing this because this should have been filed as an issue and the revert will not change or fix anything.

jdandoy:
Hi @mmilesi,

Did you try running with commit 2706ff921e (right before this one?).  I went back one commit and found everything to work properly.

PR #346, opened 2015-11-12
ROOT breaks std::cout (or rather overloads it in a strange way).

kratsg:
This totally works fine either way, however the original way just seems to break code that uses ROOT. I'm using a different way now with the sstream and getline which should work as we've used it before.

This resolves the issue that was discussed in #344 .

/cc @mmilesi @jdandoy @johnda102 and catdog


kratsg:
@mmilesi reports that this doesn't fix it. But it does fix a current major bug.

PR #347, opened 2015-11-12
has_exact, has_match, and parse

kratsg:
Adding in two new functions:

```
InfoSwitch::has_exact(flag)
InfoSwitch::has_match(flag)
```

and deprecating `InfoSwitch::parse`. From now on, you can chose whether you want a partial match or an exact match. Using `parse` will spit out a lot of warning messages that it is being deprecated but the original behavior is maintained to make the transition a little less abrupt.

### DIFF
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -86,76 +86,76 @@ namespace HelperClasses{
    *
    **************************************/
   void EventInfoSwitch::initialize(){
-    m_pileup        = parse("pileup");
-    m_shapeEM       = parse("shapeEM");
-    m_shapeLC       = parse("shapeLC");
-    m_truth         = parse("truth");
-    m_caloClus      = parse("caloClusters");
-    m_muonSF        = parse("muonSF");
-    m_electronSF    = parse("electronSF");
+    m_pileup        = has_exact("pileup");
+    m_shapeEM       = has_exact("shapeEM");
+    m_shapeLC       = has_exact("shapeLC");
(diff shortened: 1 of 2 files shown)


kratsg:
@johnda102 -- to make this work for you, you'll want to use something like

```
m_flavorTag = has_exact("flavorTag");
m_flavorTagHTL = has_exact("flavorTagHLT");
```

mmilesi:
Ok, I confirm this works at parsing BTag SF switches correctly.

johnda102:
Looks fine to me.  Im still confused as to why that jeffs code was not running before

kratsg:
@johnda102 The main point is that if you remove the quotes in your config files -- it will work. Jeff and Marco are using TEnv which changed the way they parsed the config files so that exact matches do not match `"kinematic` anymore.

johnda102:
got it thanks.

TEnv is garbage

On Thu, Nov 12, 2015 at 2:32 PM, Giordon Stark <notifications@github.com>
wrote:

> @johnda102 <https://github.com/johnda102> The main point is that if you
> remove the quotes in your config files -- it will work. Jeff and Marco are
> using TEnv which changed the way they parsed the config files so that exact
> matches do not match "kinematic anymore.
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/347#issuecomment-156102863>
> .
>



-- 
University of Chicago

kratsg:
> TEnv is garbage

:+1:

PR #353, opened 2015-11-16
Adding flags to optionally turn off elec and muon overlap removal

johnda102:


gfacini:
Did you mean to make those changes to JetSelector as well?

johnda102:
yeah, the jet selector is just adding debug output and fixing a minor bug.

gfacini:
Okay - it seems fine from looking at the code.

Do you need to call "new" every event - isn't that unnecessarily slow?
Also, who is deleting this object?

kratsg:
Calling `new` on every event is not a good idea at all. This has a memory leak right now.

johnda102:
I can set it up to not call new on every event, but the new objects are
being deleted.

On Mon, Nov 16, 2015 at 5:02 PM, Giordon Stark <notifications@github.com>
wrote:

> Calling new on every event is not a good idea at all. This has a memory
> leak right now.
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/353#issuecomment-157079935>
> .
>



-- 
University of Chicago

kratsg:
> I can set it up to not call new on every event, but the new objects are
> being deleted.

Are you sure? I quickly profiled and I see a memory leak.

gfacini:
If I remember correctly after you add it to the
TStore then TStore takes ownership and will
delete it appropriately...

johnda102:
ok only new once now on init. and delete in finalize. 
(Im not adding these to teh TStore so I think the clean up needs to be done on our end)

PR #357, opened 2015-11-17
Add Tau utilities

### LINKED ISSUE #311: Adding Tau Selector
@johnda, @kkrizka 

I remember you said you were implementing the photon selector taking advantage of a base selector class. Can we have a common starting point to implement the tau selector as well? I will take care of this.

mmilesi:
This would close #311


johnda102:
lot of changes.  

Looks good to me.

One line that caught my attention:
```c++
 static SG::AuxElement::ConstAccessor<char> myAccessor(flagSelect);
```
Seems counterintuitive as the first call with flagSelect will create the object and subsequent calls with different flagSelect values will do nothing.  Do we want the static here ?

mmilesi:
Hi John,

actually `flagSelect` will be the same for all objects in the container passed to `makeSubsetCont` (in the overlap removal case, it is always = "overlaps"). So that should not be harmful. As for the static, I honestly don't remember. @kratsg , do you know what's the rule of thumb for usage of static/non-static accessors?

MArco

johnda102:
ok sounds good.
do we want to hard code the static overlap removal to have "overlaps"
instead of  flagSelect to avoid confusion ?

On Tue, Nov 17, 2015 at 2:05 PM, Marco Milesi <notifications@github.com>
wrote:

> Hi John,
>
> actually flagSelect will be the same for all objects in the container
> passed to makeSubsetCont (in the overlap removal case, it is always =
> "overlaps"). So that should not be harmful. As for the static, I honestly
> don't remember. @kratsg <https://github.com/kratsg> , do you know what's
> the rule of thumb for usage of static/non-static accessors?
>
> MArco
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/357#issuecomment-157365546>
> .
>



-- 
University of Chicago

mmilesi:
Actually, `makeSubsetCont` might be used also outside the overlap removal. For instance, in a selector, we might want to resize a container to contain only the objects that have "passSel" decoration `true`.

That's why I made the `passSel` flag configurable.

johnda102:
On Tue, Nov 17, 2015 at 2:18 PM, Marco Milesi <notifications@github.com>
wrote:

> Actually, makeSubsetCont might be used also outside the overlap removal.
> For instance, in a selector, we might want to resize a container to contain
> only the objects that have "passSel" decoration true.
>
ok then I think we dont want to make the myAccessor static.

if we call it first with "objects" and then with "passSel"
The second call will not create a new object that accesses "passSel" it
will use the already created static object which uses "objects"


> That's why I made the passSel flag configurable.
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/357#issuecomment-157368546>
> .
>



-- 
University of Chicago

kratsg:
> if we call it first with "objects" and then with "passSel"
> The second call will not create a new object that accesses "passSel" it
> will use the already created static object which uses "objects"

John is right. Static should be used when you know the input value is fixed, in this case, it isn't.

mmilesi:
Ok thanks very much for clarifying that.

Let me update this.

kratsg:
Are you also going to make the updates to add the documentation so it's at least included as well?

mmilesi:
Hi Giordon, 

it seems I have to learn how to make a doxygen documentation first. And I couldn't find any example in what is now in docs/ 
If you have some tips, I'll be happy to do it tomorrow.

PR #361, opened 2015-11-18
Change OLR tool instance in xAH::OverlapRemover from 'private' to 'protected'

mmilesi:
This change should not negatively affect the users.

This is needed as in HTop analysis we have developed a wrapper to a 'custom' Overlap Removal tool which would inherit from `xAH::OverlapRemover`, and it needs to have direct access to the OLR tool instance.


kratsg:
My only comment would be that `setCutFlowHist` should return a `StatusCode` rather than an `EL::StatusCode` since it's not called by EventLoop... but that's pretty much it.

mmilesi:
That seems to be something we should change quite across the board then :wink: 
Ok to merge?

PR #363, opened 2015-11-19
xAH_run Condor AdditionsFixed problem with filelists and condor in xAH_run. Added option --op…

kkrizka:
Fixed problem with filelists and condor in xAH_run. Previously xAH_run assumed that all running on condor is with DQ2.

Added option --optCondorWait to submit to condor using the submit() command. This causes the code to wait until all jobs are finished and then merge all of the outputs automatically.

### DIFF
--- a/scripts/xAH_run.py
+++ b/scripts/xAH_run.py
@@ -165,6 +165,7 @@ prun.add_argument('--optGridOutputSampleName', metavar='', type=str, required=Fa
 
 # define arguments for condor driver
 condor.add_argument('--optCondorConf', metavar='', type=str, required=False, default='stream_output = true')
+condor.add_argument('--optCondorWait', action='store_true' , required=False)
 
 # define arguments for lsf driver
 lsf.add_argument('--optLSFConf', metavar='', type=str, required=False, default='-q short')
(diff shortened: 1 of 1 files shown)


kratsg:
Hi @kkrizka -- will this merge output by sample or ALL output?

kkrizka:
By sample.

kratsg:
Oh, awesome!

kratsg:
The only other commend I have is you have the `needXRD` but then use `args.use_scanDQ2` and I can't see a logical difference in the script as it worked back then (note that a shared-fs condor doesn't want scanDQ2 automatically).

kkrizka:
Previously xAH_run assumed that condor means you are running on a DQ2 sample (ie: called scanDQ2). However the input can also be list of ROOT files on an xrd server, listed in a filelist. The `needXRD` now checks if xrdclients are present (useful for DQ2 and filelists). Then `args.use_scanDQ2` runs scanDQ2 on if the input is from DQ2.

kratsg:
> Previously xAH_run assumed that condor means you are running on a DQ2 sample (ie: called scanDQ2).

How do you see that? The old code was

```
   use_scanDQ2 = (args.use_scanDQ2)|(args.driver in ['prun', 'lsf'])
```

This now means if someone uses `prun`, the script doesn't automatically use `scanDQ2` but will require the user to add that option.

kkrizka:
Ah right, I forogt also added 'condor' to the list of drivers for this check. 
Since the condor driver checks if all files can be opened before submission, I 
wanted to make sure xrd is setup properly.

After this change it reduced to what I described previously.

kratsg:
Right, but you'll notice that you changed `use_scanDQ2` to `args.use_scanDQ2` so it won't automatically switch on with specific drivers.

PR #367, opened 2015-11-21
Debugging Muons

johnda102:
@mmilesi  have a look ive made a few minor changes to the muons selection to support more than one instance with different quality cuts and to calibration to avoid a bunch of error print out

### DIFF
--- a/Root/LinkDef.h
+++ b/Root/LinkDef.h
@@ -30,6 +30,7 @@
 
 /* Plotting Tools */
 #include <xAODAnaHelpers/JetHistsAlgo.h>
+#include <xAODAnaHelpers/MuonHistsAlgo.h>
 #include <xAODAnaHelpers/TrackHistsAlgo.h>
 
 /* Tree */
@@ -73,6 +74,7 @@
 #pragma link C++ class BJetEfficiencyCorrector+;
 
 #pragma link C++ class JetHistsAlgo+;
+#pragma link C++ class MuonHistsAlgo+;
 #pragma link C++ class TrackHistsAlgo+;
 
 #pragma link C++ class TreeAlgo+;
--- a/Root/MuonCalibrator.cxx
+++ b/Root/MuonCalibrator.cxx
@@ -303,9 +303,10 @@ EL::StatusCode MuonCalibrator :: execute ()
       for ( auto muSC_itr : *(calibMuonsSC.first) ) {
 
 	if ( m_debug ) { Info("execute()", "  uncailbrated muon %i, pt = %.2f GeV", idx, (muSC_itr->pt() * 1e-3)); }
-	
-	if ( m_muonCalibrationAndSmearingTool->applyCorrection(*muSC_itr) == CP::CorrectionCode::Error ) {  // Can have CorrectionCode values of Ok, OutOfValidityRange, or Error. Here only checking for Error.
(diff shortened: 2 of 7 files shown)


mmilesi:
Hi @johnda102 

this looks good to me. I agree that the `ToolStore` retrieval should be removed if one wants to have multiple selectors with different configurations.

Just one question:

https://github.com/UCATLAS/xAODAnaHelpers/pull/367/files#diff-76bd3f85a73152f1f60bf2ddfa6266c8R306

is this b/c of derivations thinning out tracks for (I guess) very low pT muons? So far I never encountered crashes, so I thought the check on the existence of the track particle for a muon candidate was done internally by the calibration tool.

Marco

johnda102:
On Sat, Nov 21, 2015 at 8:25 AM, Marco Milesi <notifications@github.com>
wrote:

> Hi @johnda102 <https://github.com/johnda102>
>
> this looks good to me. I agree that the ToolStore retrieval should be
> removed if one wants to have multiple selectors with different
> configurations.
>
> Just one question:
>
>
> https://github.com/UCATLAS/xAODAnaHelpers/pull/367/files#diff-76bd3f85a73152f1f60bf2ddfa6266c8R306
>
> is this b/c of derivations thinning out tracks for (I guess) very low pT
> muons? So far I never encountered crashes, so I thought the check on the
> existence of the track particle for a muon candidate was done internally by
> the calibration tool.
>

yes I think so.  I dont see any crashes either, but I see an error message
each time a muon without a track particle gets passed.  (Using EXOT8). When
running the ttbar MC they are very many messages.

john

> Marco
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/367#issuecomment-158604021>
> .
>



-- 
University of Chicago

mmilesi:
Hi @johnda102 ,

I had updated my comment with another question you probably missed :wink:  I copy it here:


> However, I think we should do something like this:
> 
> https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetSelector.cxx#L383
> 
> since I suspect it could be a potential issue to have multiple instances of the same CP tool with the same name on asg::toolStore. Maybe we could still drop the attempt to retrieve from ToolStore...
> 
> I will take care of fixing this also in the other selectors etc. in a different PR.

Marco

johnda102:
yup missed it.

Yes that sounds good to me

On Sat, Nov 21, 2015 at 1:26 PM, Marco Milesi <notifications@github.com>
wrote:

> Hi @johnda102 <https://github.com/johnda102> ,
>
> I had updated my comment with another question you probably missed [image:
> :wink:] I copy it here:
>
> However, I think we should do something like this:
>
>
> https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetSelector.cxx#L383
>
> since I suspect it could be a potential issue to have multiple instances
> of the same CP tool with the same name on asg::toolStore. Maybe we could
> still drop the attempt to retrieve from ToolStore...
>
> I will take care of fixing this also in the other selectors etc. in a
> different PR.
>
> Marco
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/367#issuecomment-158634954>
> .
>



-- 
University of Chicago

mmilesi:
Ok, it will be fixed in my upcoming PR!

PR #369, opened 2015-11-21
Allow for multiple instances of xAH::Selector (Muon and Electron)

mmilesi:
This follows what was started with #367


johnda102:
Looks good

PR #384, opened 2015-11-27
Leakage corrections for electron isolation variables

mmilesi:
This needs to be included, see:

https://twiki.cern.ch/twiki/bin/view/AtlasProtected/IsolationSelectionTool

and

https://twiki.cern.ch/twiki/bin/view/AtlasProtected/IsolationLeakageCorrections


kratsg:
Add those links here: https://github.com/UCATLAS/xAODAnaHelpers/blob/master/docs/ToolsUsed.rst

PR #388, opened 2015-12-02
Fix for BTag decoration name

mmilesi:
Otherwise all BTag weight branches get filled with junk unless using `TEnv`.

@johnda102 can you check this is ok for you? Cheers


johnda102:
looks good

Issue #379, opened 2015-11-25
For SkimmyDumping


kratsg:
Copying Cutbookeepers

```c++
    #include "xAODCutFlow/CutBookkeeper.h"
    #include "xAODCutFlow/CutBookkeeperContainer.h"
    #include "xAODCutFlow/CutBookkeeperAuxContainer.h"
    
    // Retrieve the input container:
    const xAOD::CutBookkeeperContainer* input = 0;
    event->retrieveMetaInput(input, "CutBookkeepers"); 

    // Create an output container
    xAOD::CutBookkeeperContainer* output =  new xAOD::CutBookkeeperContainer();
    xAOD::CutBookkeeperAuxContainer* output_aux =  new xAOD::CutBookkeeperAuxContainer();
    output->setStore( output_aux );

    
    // Copy input to output
    for( const xAOD::CutBookkeeper* cutbookkeeper : *input ) {
      xAOD::CutBookkeeper* out = new xAOD::CutBookkeeper();
	output->push_back( out );
	*out = *cutbookkeeper;
    }
      
   
    event->recordMeta(output,"CutBookkeepers");
    event->recordMeta(output_aux,"CutBookkeepersAux.");
    event->fill();
```

kratsg:
For checking which type we dump -- we just need to go over all common container types and make it easy to add support for others.

```c++
if(NewType* v = dynamic_cast<NewType*>(old)) { 
    // old was safely casted to NewType 
    v->doSomething(); 
}
```

as an example.

PR #398, opened 2015-12-10
Skimmy dumping

### LINKED ISSUE #4: SkimmyDumping
Event Skimming and xAOD Outputs

kratsg:
This will close #4, #379 when merged.

I need to add CutBookkeepers so that it copies it over. It currently doesn't. Please do not merge this. Just simply bask in the glory of the beautiful script.

We also need to add all the various containers and type-checking.


kratsg:
Just need to add CutBookkeeper metadata copying and we are golden.

kratsg:
I added CutBookkeepers -- I am going to merge this after it finishes the check since I am currently satisfied with how it works. Documentation is added in the header file for it (so read it online!)

https://gist.github.com/f68ed0e958e61415cc60

Will update with more fixes for some other stuff before ready to merge.

kratsg:
The only thing breaking the code right now

```

terminate called after throwing an instance of 'SG::ExcStoreLocked'
  what():  SG::ExcStoreLocked: Attempted to modify auxiliary data in a locked store: `::weight_prescale' (261)
Aborted
```

and this occurs because when copying objects, decorations become AuxData items and therefore become locked and can't be changed! Frick. I don't see a way around it. One can protect against it by only decorating a decoration if it does not exist... otherwise, you cannot do anything.

kratsg:
@mmilesi CutBookkeepers is confirmed to be ok - the corresponding DataHeaders do not work (which is why `checkxAOD.py` is breaking but `checkFile.py` seems fine). Others have used the same code in their analysis (preprocessing files basically) and noticed that everything is fine.

The LockedStore is somewhat problematic.

kratsg:
@mmilesi I am going to merge this in an hour or so. There is an issue that I will need to fix later with xAH in general which I'll elaborate in a minute.

To produce a mini-xAOD, I used the following configs: https://gist.github.com/005f666d7605ddfaf039

 `jetSelect.json` is the first run through xAH, and `nextSelect.json` is the second run through xAH using commands:

`xAH_run.py --files /share/t3data3/kratsg/xAODs/mc15_13TeV.410000.PowhegPythiaEvtGen_P2012_ttbar_hdamp172p5_nonallhad.merge.AOD.e3698_s2608_s2183_r6630_r6264_tid05419191_00/AOD.05419191._000003.pool.root.1 --config jetSelect.json --nevents=10 --force -vv direct`

and then

`xAH_run.py --files submitDir/data-out_miniXAOD/mc15_13TeV.410000.PowhegPythiaEvtGen_P2012_ttbar_hdamp172p5_nonallhad.merge.AOD.e3698_s2608_s2183_r6630_r6264_tid05419191_00.root --submitDir=nextSubmitDir --config nextSelect.json --nevents=10 --force -vv direct`

For issues, see #400 and #401.

PR #429, opened 2016-01-04
Change on BJetEfficiencyCorrector and xAH_run

zqhsfz:
1. BJetEfficiencyCorrector

Add support of using WP in CDI file different from the WP in BJetSelector. This would allow user to use a WP that does not have official calibration yet. For example, for antikt2 track-jet, one can cut on 77% WP but use 70% WP calibration as an approximation (before 77% WP calibration is available). By default, WP used in CDI file is the same as the WP used by BJetSelector.

2. xAH_run.py

Add option to specify number of files per worker for LSF driver


kratsg:
go ahead

PR #288, opened 2015-10-16
Add jet collections

johnda102:

Issue #172, opened 2015-07-10
Adding truth / Mulitple collection to ntuple in HelpTreeBase


johnda102:
Issue here is we would like to write out jets and truth jets to an ntuple. 
In the past we've either done one or the other. I can think of two ways forward

a) void HelpTreeBase::AddTruthJets(const std::string detailStr)

or 

b) void HelpTreeBase::AddJets(const std::string detailStr, const std::string jetColName="")

In a) we would call AddJets to add the jets and AddTruthJets to add truth jets

In b) we would call AddJets as usual to add jets. then we would call AddJets("kinematic ...", "truth") 
where the second string specifies a postfix for branches in the ntuple. 

What do people think ?  I can see b) being useful if people want to write out multiple jet collections of any type...eg: trigger jets and offline jet ect. 

john

gfacini:
Hi John,

I like the idea of adding more collections vis AddJets with a ColName.  I have a 2 requests:
1. jetColName = "" gives what we have now (no ones code breaks)
2. The names of the collections are forced to some degree to agree with the naming conventions in the xAODs.  For example, all of the jet variables in the tree have the same name as the corresponding xAOD jet moment as it is not our job to remake or rename the EDM but to make using it easier.  I would like to avoid one person having EMTopo and someone else having TopoEM.  We quickly start to get into a configuration where every tree coming out of this code has unique names - sounds like the 100 flavors for D3PDs....

Cheers,
Gabriel

mmilesi:
Hi @johnda102 , @gfacini

sorry for hijacking this conversation, but it's probably the right place for me to go.

A guy that shares the framework with me would like to perform jet constituents truth matching, as well as get somehow the truth jet origin information (e.g., jet is from a W, from a top quark, etc.).

I guess you are familiar with this and have already implemented it in xAH somewhere. Could you kindly point me to the parts of the code where this truth matching for jets is actually implemented, and how the relevant info gets dumped in the ntuple?

Thanks!

Marco

johnda102:
Hi Marco, 

  Sorry just saw this now.  I havent done any truth info at this level.  Ive only used the standard bjet truth matching in btagging.

PR #431, opened 2016-01-11
Pull request for multiple jet collections including trigger jets.  

urania277:
Here's a rather big pull request for some changes that are needed for the Trigger-Level Analysis. 

Here's a summary of the changes:
   * Reco, trigger and truth jets can be in the same tree
   * Added HLTJetGetter class (+some initial docs) to retrieve trigger jets out of the TriggerDecisionTool
   * TriggerDecisionTool initialized in BasicEventSelection only if it's not already in the store
   * Some memory leaks and minor bugs fixed 
   * Allow HelpTreeBase to save the event-level cleaning flags 

The main change of this PR is that one can now extend the number of jet collections by having a different prefix for reco jets, trigger jets and truth jets in the same tree. Nothing will change if only one jet collection is requested, regardless of the type: the prefix in the tree will still be jet_[variable]. If trigger jets are requested in the config file, the prefix will be trigjet_[variable], while if truth jets are requested in the config file, the prefix will be truthjet_[variable]. This mirrors what's done for fat jets, but since the EDM is the same for all these collections the same function is used (just some variables won't be filled or throw an error if they are completely unavailable). 

Everything should be backwards compatible, meaning that the current configuration people use without asking for anything new should still work in the exact same way. In any case, since this is a large chunk of changes, I would appreciate it if other jetty developers also had a quick test within their analysis code before we merge it all in. 

The next PR will fill a few more small bugs and add a few features (e.g. allowing for an 'mjj' variable for various jet collections to be calculated in algorithms inheriting from TreeAlgo, making JetSelector work for trigger jets as well...) but I preferred to get rid of this big chunk first and make smaller incremental changes from now on.  

Thanks!
Caterina, for the TLA team

### DIFF
--- a/Root/HLTJetGetter.cxx
+++ b/Root/HLTJetGetter.cxx
@@ -1,0 +1,223 @@
+/******************************************
+ *
+ * This class gets HLT jets from the TDT and can be expanded to get other features
+ *
+ * Merlin Davies (merlin.davies@cern.ch)
+ * Caterina Doglioni (caterina.doglioni@cern.ch)
+ * John Alison (john.alison@cern.ch)
+ *
+ *
+ ******************************************/
+
+// c++ include(s):
+#include <iostream>
+#include <vector>
+
+// EL include(s):
+#include <EventLoop/Job.h>
+#include <EventLoop/StatusCode.h>
+#include <EventLoop/Worker.h>
+
+// EDM include(s):
+//CD: do we need all this stuff?
+#include "xAODEventInfo/EventInfo.h"
+#include "xAODJet/JetContainer.h"
+#include "xAODJet/JetAuxContainer.h"
+#include "xAODJet/Jet.h"
+#include "xAODBase/IParticleHelpers.h"
+#include "xAODBase/IParticleContainer.h"
+#include "xAODBase/IParticle.h"
+#include "AthContainers/ConstDataVector.h"
+#include "AthContainers/DataVector.h"
+#include "xAODCore/ShallowCopy.h"
+
+// package include(s):
+#include "xAODAnaHelpers/HelperFunctions.h"
+#include "xAODAnaHelpers/HLTJetGetter.h"
+#include <xAODAnaHelpers/tools/ReturnCheck.h>
+
+#include "TrigConfxAOD/xAODConfigTool.h"
+#include "TrigDecisionTool/TrigDecisionTool.h"
+
+// ROOT include(s):
+#include "TEnv.h"
+#include "TSystem.h"
+
+using std::cout;  using std::endl;
+using std::vector;
+
+// this is needed to distribute the algorithm to the workers
+ClassImp(HLTJetGetter)
+
+HLTJetGetter :: HLTJetGetter (std::string className) :
+    Algorithm(className),
+    m_trigItem(""),
+    m_outContainerName(""),
+    m_trigDecTool(nullptr)
+{
+  Info("HLTJetGetter()", "Calling constructor");
+
+  // read debug flag from .config file
+  m_debug                   = false;
+
+  m_sort                    = true;
+
+}
+
+
+EL::StatusCode HLTJetGetter :: setupJob (EL::Job& job)
+{
+  Info("setupJob()", "Calling setupJob");
+  job.useXAOD ();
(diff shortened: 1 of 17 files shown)


kratsg:
Hi Catdog -- this is excellent. Can you list the issue numbers (format: #X) [should auto-pop up as you type] for which this pull request is related to, if any?

urania277:
Hmm, not many afaik, there were #288 and #172 that are both closed and did some of the groundwork. I can make new issues if needed?

Root/HLTJetGetter.cxx
```diff
@@ -0,0 +1,225 @@
...
+
+// this is needed to distribute the algorithm to the workers
+ClassImp(HLTJetGetter)
+
+HLTJetGetter :: HLTJetGetter (std::string className) :
+Algorithm(className),
+m_triggerList(".*"),
+m_inContainerName(""),
+m_outContainerName(""),
+m_trigDecTool(nullptr)
```
kratsg: The lines above that are being configured (except for `Algorithm`) need to be placed inside the enclosure below (see other algorithms for an example of how it's done).

Root/HLTJetGetter.cxx
```diff
@@ -0,0 +1,225 @@
...
+{
+    Info("HLTJetGetter()", "Calling constructor");
+    
+    // read debug flag from .config file
+    m_debug                   = false;
```
kratsg: This does not need to be specified. `m_debug` is in `xAH::Algorithm` and is set to `false` by default.

Root/HLTJetGetter.cxx
```diff
@@ -0,0 +1,225 @@
...
+
+#include "TrigConfxAOD/xAODConfigTool.h"
+#include "TrigDecisionTool/TrigDecisionTool.h"
+
+// ROOT include(s):
+#include "TEnv.h"
+#include "TSystem.h"
+
+using std::cout;  using std::endl;
+using std::vector;
```
kratsg: These `using` statements are not needed (and we want to be explicit with namespaces)

Root/HelpTreeBase.cxx
```diff
@@ -58,6 +57,49 @@ HelpTreeBase::HelpTreeBase(xAOD::TEvent* event, TTree* tree, TFile* file, const
...
+    
+    //met
+    delete m_metInfoSwitch;
+    
+    //jet
+    
+    std::map<std::string, HelperClasses::JetInfoSwitch*>::iterator jetInfoSwitchMap_itr = m_thisJetInfoSwitch.begin();
+    std::map<std::string, HelperClasses::JetInfoSwitch*>::iterator jetInfoSwitchMap_end = m_thisJetInfoSwitch.end();
+    for( ; jetInfoSwitchMap_itr != jetInfoSwitchMap_end; ++jetInfoSwitchMap_itr) {
+        delete jetInfoSwitchMap_itr->second;
```
kratsg: Strange way to iterate -- you can just do

```
for(auto jetInfoSwitch: m_thisJetInfoSwitch)
  delete jetInfoSwitch.second;
```

or something like that.

docs/Getters.rst
```diff
@@ -0,0 +1,8 @@
+Getting Objects
+=================
+
+.. toctree::
+:maxdepth: 2
+
+HLTJetGetter
+
```
kratsg: This needs 3 spaces before the information. Should look like

```
Getting Objects
===============

.. toctree::
   :maxdepth: 2

   HLTJetGetter

```

also need to make the `=` exactly as long as the title

docs/HLTJetGetter.rst
```diff
@@ -0,0 +1,9 @@
+HLT jet getter
+=====================
+
+.. doxygenclass:: HLTJetGetter
+:members:
+:undoc-members:
+:protected-members:
+:private-members:
+
```
kratsg: Missing spaces, `=` as long as title:

```
HLT jet getter
===========

.. doxygenclass:: HLTJetGetter
   :members:
   :undoc-members:
   :protected-members:
   :private-members:

```

kratsg:
I'm pretty happy with this commit overall for right now. I'll go ahead and merge it in soon.

kratsg:
retest this please

PR #432, opened 2016-01-11
Changing "private" to "protected" in TreeAlgo

mmilesi:
This change would allow users who are using derived classes from `xAH::HelpTreeBase` to re-implement *only* what is strictly necessary in their  `xAH::TreeAlgo`-derived algorithms for tree management (i.e., make better use of inheritance).

Making this change would allow something like this:

https://github.com/UCATLAS/ASG_AnalysisFrameworkReview

All in all, This change reflects exactly what has been done already in:

https://github.com/UCATLAS/xAODAnaHelpers/pull/361

@kratsg @gfacini @jdandoy does it make sense to you?

### DIFF
--- a/xAODAnaHelpers/TreeAlgo.h
+++ b/xAODAnaHelpers/TreeAlgo.h
@@ -39,7 +39,7 @@ public:
 
   bool m_DC14;
 
-private:
+protected:
   HelpTreeBase* m_helpTree;            //!
 
 public:


urania277:
It's already included in #431 :)

mmilesi:
No way! What a coincidence :blush: 

If there are no objections, can we merge this first (being a very small PR)?

kratsg:
@mmilesi I prefer to merge in @urania277's PR first. Probably in about 5 hours since it's easier to rebase the commits here if needed.

kratsg:
This is merged in now.

PR #435, opened 2016-01-15
Update lepton SFs 

mmilesi:
This is a rather big PR. A short summary of the main changes:

- move on to 2.3.39 and update package script
- add isolation SF for electrons.
- add electron trigger MC efficiency (not SF!) and dump it in tree.
- fix a logic bug in electron trigger SF: in the way the egamma CP tool works, the SF is a **per-electron** weight, hence it must be saved for every electron in the event. I still calculate a global SF according to the formula:

event SF = 1 - Prod( 1 - SF(i) )
(the product is over electrons in the event)

, but it must be taken with a grain of salt. In principle, people should re-calculate their own event SF according to each analysis recommendations: all the ingredients (namely, per-electron SFs  and MC efficiencies) are stored as decorations and also are dumped to the final tree. 
- add "Loose_CutBL" in the list of available WPs for electron LH in `ElectronSelector`
- general cleanup of `ParticlePIDManager.h`
- add `_isTrigMatched` branches for electrons and muons in tree (they are true if at least one match with the input chains is found)

@gfacini @kratsg @jdandoy @johnda102 

Marco

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -261,20 +261,21 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
   Info("initialize()", "BTaggingSelectionTool initialized : %s ", m_BJetSelectTool->name().c_str() );
 
   //
-  // initialize the BJetEfficiencyCorrectionTool
-  //
-  std::string sf_tool_name = std::string("BJetEfficiencyCorrectionTool_") + m_name;
-  if ( asg::ToolStore::contains<BTaggingEfficiencyTool>( sf_tool_name ) ) {
-    m_BJetEffSFTool = asg::ToolStore::get<BTaggingEfficiencyTool>( sf_tool_name );
-  } else {
-    m_BJetEffSFTool = new BTaggingEfficiencyTool( sf_tool_name );
-  }
-  m_BJetEffSFTool->msg().setLevel( MSG::INFO ); // DEBUG, VERBOSE, INFO, ERROR
-
-  //
   //  Configure the BJetEfficiencyCorrectionTool
   //
   if( m_getScaleFactors ) {
+
+    //
+    // initialize the BJetEfficiencyCorrectionTool
+    //
+    std::string sf_tool_name = std::string("BJetEfficiencyCorrectionTool_") + m_name;
+    if ( asg::ToolStore::contains<BTaggingEfficiencyTool>( sf_tool_name ) ) {
+      m_BJetEffSFTool = asg::ToolStore::get<BTaggingEfficiencyTool>( sf_tool_name );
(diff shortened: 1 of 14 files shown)


gfacini:
Hi Marco,

1-SF looks odd..why is that the way the SF is applied?

Cheers,
Gabriel

mmilesi:
Hi Gabriel,

that's how in the Top WG the SF is calculated.. However, I also think -despite is simpler to calculate - that is the right way to compute the event trigger SF.

What the MuonCP group does (internally to their tool) is in fact to calculate the SF as:

event SF = [ 1 - Prod( 1 - eff_data(i) )]  / [ 1 - Prod( 1 - eff_MC(i) ) ]
(the product is over all muons in the event)

and this number is what we directly retrieve from the CP tool in `MuonEfficiencyCorrector`.

I interpret the numerator as (1 - "probability for all muons to fail the trigger in data") divided by (1 - "probability for all muons to fail the trigger in MC").

I have implemented the latter in my analysis code, but I'd prefer not to put it in xAH and let each group do whatever they think is right.

If you agree, I could simply remove the calculation of the global SF for the electron trigger in xAH. I see can be source of misunderstanding and doubt.

gfacini:
Hi Marco,

Okay I see:
SFPass = (1 - ProbFail(MC)) / (1 - ProbFail(data))

I was thinking you were writing

SFFail = 1-SFPass

or similar, which is indeed very wrong.

Cheers,
Gabriel

mmilesi:
Ok cool :)

Shall I simply remove the global SF as I suggested, or do we want to have it still?

MArco

gfacini:
Hi,

On Fri, Jan 15, 2016 at 12:50 PM, Marco Milesi <notifications@github.com>
wrote:

> Ok cool :)
>
> Shall I simply remove the global SF as I suggested, or do we want to have
> it still?
>

If the individual SFs are there then I agree to remove it
.
In the end, the global SF is somewhat analysis
dependent and we try to keep those kinds of things
up to the individual user.

Cheers,
Gabriel

kratsg:
> In the end, the global SF is somewhat analysis
> dependent and we try to keep those kinds of things
> up to the individual user.

I agree. Leave it as configurable as possible. It's pretty easy to write a TFormula after the fact.

kratsg:
Hi @mmilesi I've merged in #441 and rebased this branch. It's being retested now.

mmilesi:
Hi Giordon,

I am adding more stuff to this, so it needs to wait.
I am in the analysis approval process, so my time is limited....sorry!

Marco

mmilesi:
Hi all,

could people please start looking into this?

I know it's a very big PR, but it contains many important updates.

Cheers

Marco

kratsg:
This is :+1: from my end

PR #446, opened 2016-01-21
Update/new submit flags

kratsg:
Lots of subtle interface changes and updates to refresh the set of arguments and clean up the code more...

- `optLSFNFilesPerJob` -> `optFilesPerWorker`
- `optLSFConf` -> `optSubmitFlags` (:exclamation: no defaults)
- `optBatchSharedFileSystem` has no default now. If you need to use condor, you'll need to set this flag to `False` for UChicago's Tier3 systems

This is a relatively major commit only to `xAH_run.py`. This affects:

@kkrizka (I changed condor slightly, let me know if this is acceptable)
@zqhsfz (I changed lsf quite a bit by renaming arguments)

For those who use Tier3 (and probably condor often): @johnda102 @jdandoy @gfacini  -- if you want condor working when this PR is merged in, you will need to run

```
xAH_run.py ... condor --optBatchSharedFileSystem=False
```

which I realize is a little extra effort, but I do want to make `xAH_run.py` as highly configurable as possible for a large swath of users.

------

I have also created a `drivers_common` list of arguments that are (mostly) common to all batch drivers provided. This means that some extra fine-tuning is allowed that wasn't available before as well.


kkrizka:
How many people are using `BatchSharedFileSystem=True`? I think it should have a default of `False`. This default would setting should always work (even on shared file systems) and is required for UCT3/ATLAS Connect (aka us). If someone wants to take advantage of a shared file system, the option is there.

The other changes for the condor driver initialization should be fine. Although it might be better to also move `--optCondorWait` to `drivers_common`. The merging of outputs is a feature of the generic batch driver and not restricted to condor.

kratsg:
> How many people are using BatchSharedFileSystem=True? I think it should have a default of False. This default would setting should always work (even on shared file systems) and is required for UCT3/ATLAS Connect (aka us). If someone wants to take advantage of a shared file system, the option is there.

So according to EL's documentation in the header file, `BatchSharedFileSystem` is used by other batch drivers potentially, but it's not clear which one. Do you know if setting this to `False` by default would affect any other driver other than `condor` now or in the future? (One of the main reasons I set it to default to nothing is because of this).



> The other changes for the condor driver initialization should be fine. Although it might be better to also move --optCondorWait to drivers_common. The merging of outputs is a feature of the generic batch driver and not restricted to condor.

I can do that, change the name to `optBatchWait` which will override the drivers' specific implementation (if it uses `submit_only` to use `submit` instead). Would that work?

kkrizka:
I am not sure how the other batch drivers deal with transferring data to/from nodes. Maybe it's safer to use your approach and require the user to set it explicitly.

Also optBatchWait sounds good.

kratsg:
> I am not sure how the other batch drivers deal with transferring data to/from nodes. Maybe it's safer to use your approach and require the user to set it explicitly.

I'm in the camp where I think it's ok to switch it to False and then deal with any issues someone might have after the fact. Condor is definitely one of the major uses (as well as prun).

PR #454, opened 2016-01-23
Bump up to 2.3.41

mmilesi:
Plus a small change in the initialisation of the cutflow histograms needed b/c of ROOT 6.04.12 (NB the change is still backwards compatible with the ROOT 6 versions in previous analysis releases)


mmilesi:
Need to merge this now to get latest recommendations for Moriond. It should be just fine.

PR #460, opened 2016-01-26
Add new CDI file for BTagging (2016 Winter recommendation)

mmilesi:
@kratsg 
@gfacini 
@johnda102 
@jdandoy 

FYI


kratsg:
Just an aside -- are we ("ATLAS") ok with adding CDI files publicly?

PR #463, opened 2016-01-29
Has type=none before. Needs to be explicitly set to bool

kratsg:
@johnda102 caught this bug. Seems to be that on python 3.x -- the type is default to `bool` when `action='store_true'` but otherwise, `type=None` in python 2.x.


kratsg:
This fixes a bug I added in #446.

johnda102:
get some.

Issue #458, opened 2016-01-23
Events with odd tracks crash HelpTreeBase


urania277:
In a file from a JZ1W Pythia QCD MC sample:

mc15_13TeV.361021.Pythia8EvtGen_A14NNPDF23LO_jetjet_JZ1W.merge.DAOD_EXOT2.e3569_s2576_s2132_r6765_r6282_p2432_tid06693339_00/DAOD_EXOT2.06693339._000016.pool.root.1

I came across an event with a vertex vector of size 1 and location of primary vertex -1. This disagrees with L1888 of HelpTreeBase and crashes the entire job. 

I have no problem nuking the event from the analysis (it does not pass the vertex cut, so the user should know that it's dangerous to ask for track-related variables without that cut) but I wonder whether we should add a protection here. The options are:

1) (minimal) issue a warning and just avoid filling the track variables for jets, and whoever uses other objects and hits this should do the same
2) (cleaner) check for this kind of situation at the start of TreeAlgo and turn off all the branches that deal with tracking

I can do 1) for jets because it seems the jets are the only place accessing the vector without checking.

kratsg:
Why is it exactly crashing? Because the location is `-1` meaning there's no primary vertex to be found?

urania277:
yep, there seems to be one vertex that isn't primary. I wonder who would want those events anyways though.

kratsg:
@johnda102 -- you wrote most of the code for the primary vertex stuff. Do you have a specific opinion on how to handle these cases?

gfacini:
Hi!

I like the idea of crashing but maybe with a meaningful warning. These
events are rarely actually desired by a user. The user should use basic
event selection to apply the PV selection. That would clean up the events
 and is the proper handeling. The HS is so soft it doesn't make a
reconstructable vertex!!

All events have a dummy vertex so the reco (I think) doesn't crash from an
empty vertex container.

Cheers,
Gabriel

PR #468, opened 2016-02-04
Default flags for BasicEventSelection

mmilesi:
Hi,

there's a colleague in Melbourne who's ramping up with xAH running with a super simple configuration.

After his feedback, I thought it'd be better to set some of the flags in BasicEventSelection to `false` by default, such as the primary vertex cut and the GRL cut.

Do you agree?

@johnda102 @kratsg @gfacini

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -86,7 +86,7 @@ BasicEventSelection :: BasicEventSelection (std::string className) :
   m_checkDuplicatesMC	= false;
 
   // GRL
-  m_applyGRLCut = true;
+  m_applyGRLCut = false;
   m_GRLxml = "$ROOTCOREBIN/data/xAODAnaHelpers/data15_13TeV.periodAllYear_HEAD_DQDefects-00-01-02_PHYS_StandardGRL_Atlas_Ready.xml";
   //https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/GoodRunListsForAnalysis
   m_GRLExcludeList = "";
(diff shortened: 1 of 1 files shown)


kratsg:
I think this is a good approach. We should aim to apply as little as possible -- so that when the user starts looking into what they can do, they apply it themselves. We don't want a gotcha where we set a special type of cut or cleaning to `true` by default and making it hard for them to figure out what/how it happened.

johnda102:
yes I like it.

On Thu, Feb 4, 2016 at 6:10 AM, xAH Bot <notifications@github.com> wrote:

> Everything checked out ok, [image: :+1:] to merge. [image: :beers:]
>
> Commit sha1: efbb222
> <https://github.com/UCATLAS/xAODAnaHelpers/commit/efbb222a904248614dff74be2f780e7eb4c9a53f>
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/468#issuecomment-179643570>
> .
>



-- 
University of Chicago

gfacini:
I find it to be odd but if you all like it, go for it!
You already did ;)
Democracy!

kratsg:
Majority rules!

mmilesi:
Haha if this is what the people wants :)

BTW Gabriel, why do you find it odd?

gfacini:
I find it odd because there are more who would use these cuts then not ...
it is almost required to use them! But okay, either way, it is no big deal

urania277:
I suspect that L105 still crashes the jets due to #458

kratsg:
@urania277 : https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/BasicEventSelection.cxx#L105 ?

urania277:
Yep - I didn't get around to fixing it and I'm not sure I can properly do it and test things in the next couple of days. What about putting it to true by default instead (as a way to buy time even)?

kratsg:
@urania277 -- I think it's better to let it crash for two reasons:

1. @gfacini mentioned that crashing is better since it lets users know they have events with problematic vertices
2. I mentioned above that adding more cuts as default would make it harder for users to know exactly EVERYTHING they are doing

urania277:
ok, that works for me - maybe the error could be more meaningful though? Right now it's just a M_range_check, uncaught

kratsg:
#458 should be resolved by implementing a check and erroring out.

PR #470, opened 2016-02-05
provisional CDI file with ak2 track jets

alexandertuna:


kratsg:
This file is ok to leave in a public repo, right?

alexandertuna:
Yeah. The information inside the root file is stored in CalibrationData classes, which aren't public. But yeah also this is pretty low level information, I think atlas people cannot give a shit about this being on a repo.

kratsg:
We'll have to be careful. We don't have a license added to xAH - and if ATLAS has an issue -- then yeah.

PR #488, opened 2016-02-16
JetCalibrator Should Calibrate Jets Only Once

kkrizka:
The JetCalibrator was recalibrating the jets for all of the systematic variations. Now it calibrates only the nominal collection and making systematic containers a shallow copies of it. This means that the nominal shallow copy is always created and put in TStore. However it is only added to the systematics list (outputAlgo) when requested.

My quick test seems to indicate a factor of 2 speed improvement as a result of this change.

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -576,8 +576,6 @@ EL::StatusCode JetCalibrator :: execute ()
   // add vector of systematic names to TStore
   RETURN_CHECK( "JetCalibrator::execute()", m_store->record( vecOutContainerNames, m_outputAlgo), "Failed to record vector of output container names.");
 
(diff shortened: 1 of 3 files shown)


kkrizka:
Whoops, I committed my next change while the check for this request was under consideration. So the script for making the cross-section metadata files from AMI will come with this.

kratsg:
Does the python script for cross-section metadata files get the correct n_events? I have a feeling it doesn't since you need to grab that out of the derivation file itself?

kkrizka:
That is correct. You should get it out of the derivation file itself.

It does not do anything with n_events since there is no column for that in the SUSY MetaData files as far as I can tell. See https://github.com/UCATLAS/xAODAnaHelpers/blob/master/data/metadata/susy_crosssections_13TeV.txt .

Issue #482, opened 2016-02-09
Inheritance Based Particle Histogram Classes


kkrizka:
I am working on adding a `PhotonHists` class for useful photon histograms. Since some of the code is similar to the other particle histogram classes, I think it would be useful to use inheritance to simplify the code. I have a prototype ready for the photons. You can see it at https://github.com/kkrizka/xAODAnaHelpers/commit/69c2c395fa6c68c4013f6573d76d923d525a0300

It adds a new `IParticleHists` class that plots the particle 4-vectors, with separate histograms for `NLeading#` particles. The particles are distinguished by a `prefix` string (output histogram name) and `title` string (for x-axis labels), initialized though the constructor. Then the `PhotonHists` class inherits from it, adding (in the future) photon specific histograms. What do others think of propagating this structure to other histogram classes? It will automatically add support for plotting `NLeading#` in `MuonHists` and ability to plot 4-vector information for particles not supported in xAODAnaHelpers yet (ie: TruthParticles).

Before I perform a pull request, I need to finish the following tasks:
- rebase on latest master
- add inheritance to `ParticleHistsAlgo` (aka share code for the systematics loop)
- add the photon-specific histograms

johnda102:
I think this is clearly the way to go.

On Tue, Feb 9, 2016 at 10:31 AM, Karol Krizka <notifications@github.com>
wrote:

> I am working on adding a PhotonHists class for useful photon histograms.
> Since some of the code is similar to the other particle histogram classes,
> I think it would be useful to use inheritance to simplify the code. I have
> a prototype ready for the photons. You can see it at kkrizka@69c2c39
> <https://github.com/kkrizka/xAODAnaHelpers/commit/69c2c395fa6c68c4013f6573d76d923d525a0300>
>
> It adds a new IParticleHists class that plots the particle 4-vectors,
> with separate histograms for NLeading# particles. The particles are
> distinguished by a prefix string (output histogram name) and title string
> (for x-axis labels), initialized though the constructor. Then the
> PhotonHists class inherits from it, adding (in the future) photon
> specific histograms. What do others think of propagating this structure to
> other histogram classes? It will automatically add support for plotting
> NLeading# in MuonHists and ability to plot 4-vector information for
> particles not supported in xAODAnaHelpers yet (ie: TruthParticles).
>
> Before I perform a pull request, I need to finish the following tasks:
>
>    - rebase on latest master
[...]

PR #493, opened 2016-02-18
Inheritance Particle Histogram Classes

kkrizka:
I am ready to merge the update discussed in #482 . The changes should be mostly backward compatible. The only ones that are not are related to `JetHists`.

- #LeadingJets option was replaced with NLeading#
- Individual jet histograms are now named `jetPt_0` instead of `jetPt_jet0`

Also all pt histograms now have the same range.

I will merge this tomorrow morning CERN time unless there are complaints.


kratsg:
@kkrizka -- use `rebase` next time, not merge.

kratsg:
@kkrizka - how do you expect IParticleHistsAlgo to be used? You have overloads in there but EventLoop is calling the executes()...

PR #495, opened 2016-02-20
Add track z0*sin(theta) as a decoration

mmilesi:
Doing this once and for all in the electron and muon selectors.
In this way any downstream algorithm does not need to recompute it.

### DIFF
--- a/Root/ElectronSelector.cxx
+++ b/Root/ElectronSelector.cxx
@@ -897,9 +897,11 @@ int ElectronSelector :: passCuts( const xAOD::Electron* electron, const xAOD::Ve
 
   double d0_significance = fabs( xAOD::TrackingHelpers::d0significance( tp, eventInfo->beamPosSigmaX(), eventInfo->beamPosSigmaY(), eventInfo->beamPosSigmaXY() ) );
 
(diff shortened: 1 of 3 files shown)


kratsg:
@mmilesi -- we want algorithms to be independent. So recomputing is less of an issue. If you add it in electron and muon selectors, that implies you want to have them in your chain for other downstream algorithms to use the computed values?

mmilesi:
Hi Giordon,

yes, for example, in my analysis I use z0sintheta and d0sig as criterions to define tight and anti-tight leptons for background estimation.
Since these two quantities are not accessible directly from the xAOD::TrackParticle object, it would be  
so much handy to simply set them in the selectors (btw, this happens only for the electrons / muons who actually pass the selection) and pull them back via accessors downstream.
Also, in my analysis tree class, I have to store such info for different "categories" of leptons in different branches, and recompute such values all the time would be just overhead.

Marco

mmilesi:
Can we merge this?

mmilesi:
Just to stress a point here: the way d0 significance and z0*sin(theta) are calculated is NOT analysis-dependent, rather it follows the prescriptions of the Tracking CP group. So it makes a lot of sense to me to have them computed in xAH once, and let any downstream algorithm (including xAH HelpTreeBase!) retrieve them via accessors.

PR #476, opened 2016-02-06
The Exorcism

kratsg:
> The path of the righteous configuration is beset on all sides by the iniquities of the selfish and the tyranny of TE(n)vil files.
> 
> Blessed is he who, in the name of code quality and open-source, shepherds the weak through the valley of ROOTness, for he is truly his brother's keeper and the finder of lost programmers.
> 
> And I will force push upon thee with great vengeance and furious anger those who attempt to commit and destroy our framework. And you will know my name is the Git when I lay my vengeance upon thee.
> 

tl;dr - no more TEnv files. This will resolve #471 indirectly (issues raised about confusing people and stuff).

> The power of Linus compels you!

PR #496, opened 2016-02-20
Make private members in OverlapRemover "protected" 

mmilesi:
After the exorcism #476, some setting in OverlapRemover has been moved from `configure()` to `initialize()`

Since some analyses are using algorithm(s) inherited from OverlapRemover, and overload `initialize()`, I have made this change to allow accessing those data members from the derived class.

If for any reason this is dangerous, I might consider creating an additional method to set those members, and leave them `private`

@kratsg, waiting for your opinion on this.


mmilesi:
Hi @kratsg , any news on this? Ok to merge?

PR #498, opened 2016-02-21
Fixing inheretance

johnda102:


kratsg:
@johnda102 -- this is not a good commit. What's up with this one?

johnda102:
why not a good commit ? Without it the derived classes dont get called.

On Sun, Feb 21, 2016 at 5:37 PM, Giordon Stark <notifications@github.com>
wrote:

> @johnda102 <https://github.com/johnda102> -- this is not a good commit.
> What's up with this one?
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/498#issuecomment-186855365>
> .
>



-- 
University of Chicago

kratsg:
Yes, because the inheritance chain you set up wasn't correct. If you look -- you're pulling super-super execute functions into derived classes. They only need to be pulled in the IParticleHists -- then you can just simply `static_cast<xAOD::IParticleContainer*>` the objects to call the parent IParticleHists -- this is the only line to change: https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetHists.cxx#L309

instead. This also means we can use the compiler to catch issues since the way you've written the code, anyone can put any IParticleContainer-derived object through JetHists and things would crash only during run-time, rather than during compile-time. @kkrizka - why are we doing it this way?

johnda102:
I didnt set up any inhertance chain. Im getting the exisitng code to call
the derived classes.
Before my change the line you pointed to isnt even getting called.  So
making a change only there will not help.

Have a look at what Karol set up in IParticleHistAlg to see whats currently
being done.

I dont feel strongly about one implemetation vs another but you should talk
to karol about it.

On Sun, Feb 21, 2016 at 6:48 PM, Giordon Stark <notifications@github.com>
wrote:

> Yes, because the inheritance chain you set up wasn't correct. If you look
> -- you're pulling super-super execute functions into derived classes. They
> only need to be pulled in the IParticleHists -- then you can just simply
> static_cast<xAOD::IParticleContainer*> the objects to call the parent
> IParticleHists -- this is the only line to change:
> https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetHists.cxx#L309
>
> instead. This also means we can use the compiler to catch issues since the
> way you've written the code, anyone can put any IParticleContainer-derived
> object through JetHists and things would crash only during run-time, rather
> than during compile-time. @kkrizka <https://github.com/kkrizka> - why are
> we doing it this way?
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/498#issuecomment-186871736>
> .
>



-- 
University of Chicago

kkrizka:
The point of the inheritance structure for the histogram classes is to reduce the amount of shared code, more specifically the kinematic histograms and the object loop (filling the inclusive and the individual histograms). The latest commit breaks the filling of the individual histograms because it does not include the `m_numLeading` loop. For this reason I prefer `dynamic_cast` method. We can add generic features to `IParticleHists` without having to reimplement them in the subclass.

kratsg:
> The point of the inheritance structure for the histogram classes is to reduce the amount of shared code, more specifically the kinematic histograms and the object loop (filling the inclusive and the individual histograms). The latest commit breaks the filling of the individual histograms because it does not include the m_numLeading loop. For this reason I prefer dynamic_cast method. We can add generic features to IParticleHists without having to reimplement them in the subclass.

see #505

PR #500, opened 2016-02-21
MinixAOD CutBookkeeper function shouldn't call TEvent::fill.

kkrizka:
This fixes a problem with the `checkxAOD.py` script reporting an error that occurs when reporting an error about a problematic xAOD made by MinixAOD. The problem was that the `m_copyCutBookkeeper` option was calling `TEvent:fill`, which made an empty entry in `CollectionTree`. I removed this call.

Please note that the `m_copyCutBookkeeper` implementation is still not fully correct. I will make a ticket about this. But at least valid xAODs are output.

### DIFF
--- a/Root/MinixAOD.cxx
+++ b/Root/MinixAOD.cxx
@@ -151,7 +151,6 @@ EL::StatusCode MinixAOD :: initialize ()
 
     m_event->recordMeta(output,"CutBookkeepers");
     m_event->recordMeta(output_aux,"CutBookkeepersAux.");
-    m_event->fill();
   }
 
   // parse and split by comma


kratsg:
Does this really fix it? It was recommended by most people that you need to call this after the Cutbookkeeper... but interesting.

kkrizka:
Yup, it does fix it. If you run the extra `TEvent::fill`, `CollectionTree` has more entries than the particle collections. This was causing the error in `checkxAOD.py`. It also makes sense, since the metadata is not stored in `CollecionTree` and `TEvent::fill` is responsible for adding new events to it.

I got this from following your conversation in https://groups.cern.ch/group/hn-atlas-PATHelp/Lists/Archive/DispForm.aspx?ID=16504 . If you look at the last example, the `fill` function is called in the event loop and not at the initialization stage. Also as you mention in the thread, the cutbookkeepers copy should be in `fileExecute`. This is what #501 is referring to.

PR #503, opened 2016-02-21
no need for dynamic_cast checks at runtime. 

kratsg:
You can explicitly call the parent just fine. The overload resolution works normally.


johnda102:
Have you checked that this actually calls the derived classes? 
The way its currently set up i dont think this will.

PR #505, opened 2016-02-22
IParticleHists::execute(container) needs execution

kratsg:


kkrizka:
I think the problem here is that if you call `JetHists::execute(jet)` directly from an algorithm, the kinematic histograms won't get filled.

PR #506, opened 2016-02-22
Revert to using JetHists::execute(IParticle*), but add a public JetHists::execute(Jet*)

kkrizka:
Hopefully the last commit related to the inherited particle histograms. @kratsg and I came up with an implementation that made both of us happy. We readd `JetHists::execute(IParticle*)`, so the `IParticleHists::execute(container)` will take care of the filling. But we make it protected, so users won't call it by mistake with a non-jet object. User can add a single jet by calling the public `JetHists::execute(Jet*)`.

@kratsg , do you want to approve/veto this merge (in case you thought of another problem)?

### DIFF
--- a/Root/JetHists.cxx
+++ b/Root/JetHists.cxx
@@ -305,18 +305,21 @@ StatusCode JetHists::initialize() {
   return StatusCode::SUCCESS;
 }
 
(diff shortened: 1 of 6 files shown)


kratsg:
:+1:

Issue #504, opened 2016-02-22
BasicEventSelection and CutBookkeepers when running directly on xAODs


kkrizka:
We are in the process of finalizing the derivation for the dijet+ISR analysis, so in the meantime I've been running directly on xAODs. But `BasicEventSelection` only uses the `CutBookkeepers` to get the intial `sumOfWeights` if you are running on a derivation. See the check at https://github.com/UCATLAS/xAODAnaHelpers/blob/059735b/Root/BasicEventSelection.cxx#L213 .

My questions:
1. Is there a reason why `BasicEventSelection` is restricted to DxAODs only?
2. Why do we care about the `CutBookkeeper` associated with the derivation? I don't see it being use for anything other than printing out a number (verbosity?).

Also I don't think the "non meta-data" implementation of this is correct. It takes the first event weight and multiplies it by the number of entries. I don't think this works if events have different weights.
`m_MD_initialSumW        = CollectionTree->GetWeight() * CollectionTree->GetEntries()`

mmilesi:
Hi @kkrizka ,

> 1. Is there a reason why BasicEventSelection is restricted to DxAODs only?

AFAIK, only derivations apply event cuts from the primary xAOD, so it's in the derivations that one wants to keep track of the original number of generated events to get the correct normalisation. For a full xAOD, this problem does not exist, hence it skips till https://github.com/UCATLAS/xAODAnaHelpers/blob/059735b/Root/BasicEventSelection.cxx#L288

> 2. Why do we care about the CutBookkeeper associated with the derivation? I don't see it being use for anything other than printing out a number (verbosity?).

You mean https://github.com/UCATLAS/xAODAnaHelpers/blob/059735b/Root/BasicEventSelection.cxx#L257 ?
Well, that is actually used to get https://github.com/UCATLAS/xAODAnaHelpers/blob/059735b/Root/BasicEventSelection.cxx#L277 etc., and finally to fill the metadata histogram https://github.com/UCATLAS/xAODAnaHelpers/blob/059735b/Root/BasicEventSelection.cxx#L167

> Also I don't think the "non meta-data" implementation of this is correct. It takes the first event weight and multiplies it by the number of entries. I don't think this works if events have different weights.
> m_MD_initialSumW = CollectionTree->GetWeight() * CollectionTree->GetEntries()

[...]

PR #507, opened 2016-02-23
Fix for reading CutBookkeepers in full xAODs

mmilesi:
See #504 

@kkrizka , can you check everything is fine now? If there's no problem, please go ahead and merge.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -210,89 +210,74 @@ EL::StatusCode BasicEventSelection :: fileExecute ()
   //check if file is from a DxAOD
   bool m_isDerivation = !MetaData->GetBranch("StreamAOD");
 
-  if ( m_isDerivation && m_useMetaData ) {
+  // check for corruption
+  //
+  // If there are some Incomplete CBK, throw a FAILURE,
+  // unless ALL of them have inputStream == "unknownStream"
+  //
+  const xAOD::CutBookkeeperContainer* incompleteCBC(nullptr);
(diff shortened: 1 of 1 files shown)


kkrizka:
Can we keep support for not using CutBookkepers? For some reason, some TRUTH1 DxAODs don't have this metadata. I can try to implement the correct non-CutBookkeepers sum of weights tonight.

mmilesi:
Hi Karol,

In that case, why not simply reading the first bin of the cutflow and
cutflow weighted histograms?

Marco

kkrizka:
It is what I am doing right now to get around the bad non-metadata sum of weight calculation. The problem is that the `BasicEventSelection` in this PR will crash when running over the TRUTH1 due to the missing CutBookkeepers. Maybe a good compromise would be to use `m_useMetaData` to disable to disable the CutBookkeepers and not include a non-metadata sum of weight calculation. What do you think?

mmilesi:
Ah, I see.
In that case yes, let's have a protection for the whole CutBookkeepers
part, and use the cutflow in the other case.
Actually, I don't know if there's a way to get the right weights sum rather
than getting the weight event by event, which is what the cutflow
stores....maybe it's a PAThelp question.

Is it OK if I add the protection tomorrow?

Cheers
Marco

mmilesi:
Hi @kkrizka ,

I managed to make the change now :wink: 

Feel free to edit if needed

mmilesi:
Hi @kkrizka ,

any news on this? Can we merge, or are you doing some work still?

kratsg:
retest this please

PR #508, opened 2016-02-23
Fixing More of Karols bugs

johnda102:


gfacini:
Karol!!!!!!!!!!!'

kratsg:
Whoa, whoa whoa. @kkrizka - this commit makes no sense to me at all. HistogramManager uses `record(EL::Worker* wk)` and so it should be inherited entirely. What's going on?!

Issue #510, opened 2016-02-23
IParticleHists needs to call FooHists->finalize()


kratsg:
This is so that anything that needs to be cleaned up (see #509 ) will be cleaned up.

PR #511, opened 2016-02-23
Fix/johns leak and karols bugs

### LINKED ISSUE #509: JetHists memory leak
JetHists has a memory leak when it creates a new TrackHists object. This needs to be cleaned up in finalize.

kratsg:
Closes #509, #510 .


kkrizka:
The `finalize()` function does not do anything for `HistogramManager` derived classes. This PR might make sure that future calls to finalize are respected everywhere, but the small memory leak is still there. You need a `delete m_tracksInJet`.

kkrizka:
You want to make sure that the `m_tracksInJet->finalize()` call is also protected by `if(m_tracksInJet)`.

Issue #481, opened 2016-02-08
memory leak in HelpTree Base


johnda102:
Crashing grid jobs when all systematics are run

kratsg:
@johnda102 -- would you be able to identify the latest point which doesn't crash grid jobs?

See history here: https://github.com/UCATLAS/xAODAnaHelpers/commits/master/Root/HelpTreeBase.cxx

jbossios:
I'm having the same problem using 00-03-29. @johnda102: Did you manage to find the problem? 
This seems to not solve my problem: "fix memory leak in jet #512"

kratsg:
@jbossios -- @tongbaojia's commit doesn't fix that though, as pointed out. I'm still trying to find it.

kratsg:
@urania277 may have identified it already.

jbossios:
@johnda102 @kratsg: Any news on this?

kratsg:
I haven't found it :-/

mmilesi:
Sorry to hijack this...in the past I had similar issues because I forgot to call `std::vector::clear()` on a vector branch (not technically a "memory leak", but...). Did you check that already?
However, it would be strange not to have this issue when running the nominal case only.

gbesjes:
Apologies for hijacking this thread (I mostly follow the xAODAnaHelpers package out of interest), but I always notice a leak when I attach a TTree to an output file using wk()->addOutput(tree) in EventLoop-derived code. There's a thread on the ROOT forums here: https://root.cern.ch/phpBB3/viewtopic.php?f=3&t=21282 but no solution has been identified yet. Without the addOutput() call, I don't see any leaks.

jbossios:
I already look for missing std::vector::clear() but I didn't found any. I checked in the nominal case and indeed it seems there is no memory leak.

jbossios:
@gbesjes thanks for the help. I'm not doing this: wk()->addOutput(tree)

kratsg:
If there's a leak that comes up again, let me know. I'm closing this issue otherwise.

PR #512, opened 2016-02-23
fix memory leak in jet

tongbaojia:
The change shouldn't break anything.
I did a test run and valgrind and do see the memory leak associated with disappear.

==17534== 29,960 bytes in 7 blocks are definitely lost in loss record 47,048 of 48,300
==17534==    at 0x4C285FC: operator new(unsigned long) (vg_replace_malloc.c:298)
==17534==    by 0x353BF236: HelpTreeBase::AddJets(std::string, std::string) (HelpTreeBase.cxx:1507)

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -93,6 +93,9 @@ HelpTreeBase::~HelpTreeBase() {
     for(auto jetInfoSwitch: m_thisJetInfoSwitch)
         delete jetInfoSwitch.second;
 
+    for (auto jet: m_jets)
+      delete jet.second;
+
 }
 
 


kratsg:
go ahead

tongbaojia:
Hi I think this change fixes this issue:
https://github.com/UCATLAS/xAODAnaHelpers/issues/481

johnda102:
I dont see how this could fix a leak that scales with nEvents.
This change will only have an effect in finalize, the jobs run out of
memory before getting there.

On Thu, Feb 25, 2016 at 1:57 AM, tongbaojia <notifications@github.com>
wrote:

> Hi I think this change fixes this issue:
> #481 <https://github.com/UCATLAS/xAODAnaHelpers/issues/481>
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/512#issuecomment-188539918>
> .
>



-- 
University of Chicago

urania277:
I think it was a good catch still, but are we sure it's not this? 

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelpTreeBase.cxx#L1596

There's no delete for a InDetTrackSelectionTool that gets created as a new every event. 

I use jets a lot and don't see the leak, but I never use that flag.

kratsg:
@tongbaojia -- can you do a `git fetch upstream && git rebase upstream/master && git push` to update your fork? Then I'll merge it in.

tongbaojia:
Hi Giordon, 

sorry for the delay. I updated my fork now. Let me know if anything else is needed!

Best,
Tony

PR #516, opened 2016-02-23
BasicEventSelector now only creates mcEvtWeight when it does not exist.

kkrizka:
Attempt number 2, with the branch rebased correctly. Creating the `mcEvtWeight` attribute when it already exists (ie: running on MinixAOD) crashes. Should get around the problem reported here https://groups.cern.ch/group/hn-atlas-PATHelp/Lists/Archive/DispForm.aspx?ID=16526 .

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -605,45 +605,49 @@ EL::StatusCode BasicEventSelection :: execute ()
   }
 
   //------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


kratsg:
@kkrizka - one of the ways to get around it is to provide an analysis-level decoration prefix. This way, if I re-run an analysis, I can change the prefix.. and automatically gain the feature of various levels of decorations in case things change between runs.

PR #525, opened 2016-03-04
Fix filelist config search

kkrizka:
The config path was missing the directory, so a search for it always failed. 

Also sneaking in a `chmod +x` on `downloadAndMerge.py` to make running it easier.


kratsg:
@kkrizka -- there's a function we can take advantage of that is sensitive to filepath locations stored inside algorithms (`ROOT.PathResolverFindConfigFile`) I believe. We could switch over to using this, but that might be too silly of a dependency.

PR #527, opened 2016-03-04
Making MinixAODs Work

kkrizka:
Here is my work on the MinixAOD algorithm from the last few weeks. The goal was to correctly record the containers (shallow copy vs container vs ConstDataVector) and implement the systematics loop. This PR does change the configuration interface slightly, so please speak up.

The main changes are
- `m_shallowCopyKeys` no longer takes the name of the parent container. It is up to the user to add it manually. This is to simplify figuring out how to copy the parent container in the code (in `TEvent` or `TStore`? is parent a shallow copy also?) Doing a `shallowIO=false` copy can be done by adding a `*` to the container name.
- `m_deepCopyKeys` is now directly copied into `TEvent`. This allows to maintain the same container name in the output MinixAOD.
- `m_vectorCopyKeys` replaced with `m_particleCopyKeys`. This option automatically figures out how to copy the container. See the documentation in the header on the heuristics used for this. Also one can add a systematic vector list (format `container|algoName`) to save the systematics variations.

There are a few more possible improvements.
- `m_particleCopyKeys` currently determines the copy method for every event. This is only necessary to do for the first one.
- The systematic variation containers are saved as deep copies. Not ideal in terms of disk space. Before ASG allows us to save `ConstDataVector`, it might be useful to implement something similar to CxAODs (see [this presentation](https://indico.cern.ch/event/497708/session/0/contribution/19/attachments/1230593/1805627/CxAODsysts.pdf)).

### DIFF
--- a/Root/MinixAOD.cxx
+++ b/Root/MinixAOD.cxx
@@ -8,7 +8,8 @@
 #include <EventLoop/StatusCode.h>
 #include <EventLoop/Worker.h>
 // output stream
-#include "EventLoop/OutputStream.h"
+#include <EventLoop/OutputStream.h>
+#include <SampleHandler/DiskOutputXRD.h>
 
 // EDM include(s):
 #include "xAODCore/AuxContainerBase.h"
@@ -68,6 +69,7 @@ MinixAOD :: MinixAOD (std::string className) :
 {
   Info("MinixAOD()", "Calling constructor");
 
+  m_outputXRootD = "";
   m_outputFileName = "out_miniXAOD";
   m_createOutputFile = true;
   m_copyFileMetaData = false;
@@ -89,6 +91,7 @@ EL::StatusCode MinixAOD :: setupJob (EL::Job& job)
   // only create the output xaod if requested
   if(m_createOutputFile){
     EL::OutputStream out_xAOD (m_outputFileName, "xAOD");
+    if(!m_outputXRootD.empty()) out_xAOD.output(new SH::DiskOutputXRD(m_outputXRootD));
     job.outputAdd (out_xAOD);
   }
 
--- a/xAODAnaHelpers/MinixAOD.h
+++ b/xAODAnaHelpers/MinixAOD.h
@@ -81,6 +81,9 @@ public:
 class MinixAOD : public xAH::Algorithm
 {
 public:
+  /// XRootD path to copy file to (leave empty to save to disk)
+  std::string m_outputXRootD;
+
   /// name of the output file to use for xAOD dumping
   std::string m_outputFileName;
 


kratsg:
> m_shallowCopyKeys no longer takes the name of the parent container. It is up to the user to add it manually. This is to simplify figuring out how to copy the parent container in the code (in TEvent or TStore? is parent a shallow copy also?) Doing a shallowIO=false copy can be done by adding a * to the container name.

Isn't that why we add the parent with this? It would be very confusing for maintainability to have to manage two sets of configuration -- where if we want to add a shallow copy -- we add the shallow copy name to `configA` and then the parent to `configB`. it's much easier for the end-user to add shallow-copy + parent to `configA` because then you remove it all at once. This new way separates this out more and makes it harder to remove things that are matched up if you don't know that they are matched up.

> m_deepCopyKeys is now directly copied into TEvent. This allows to maintain the same container name in the output MinixAOD.

This is not always something you want. Users [including me] need to have some control over what the output container name should be. We can make it so that if an output name isn't specified, it uses the input name to copy over as well. Will this break if the user does something like

- CDV of AntiKt4EMTopoJets into TStore
- deep copy CDV record into TEvent

assuming the TEvent has an input of the same name? This might be SUPER confusing.

> m_vectorCopyKeys replaced with m_particleCopyKeys. This option automatically figures out how to copy the container. See the documentation in the header on the heuristics used for this. Also one can add a systematic vector list (format container|algoName) to save the systematics variations.
> m_particleCopyKeys currently determines the copy method for every event. This is only necessary to do for the first one.

Is this a one-stop config that can handle all cases?



> The systematic variation containers are saved as deep copies. Not ideal in terms of disk space. Before ASG allows us to save ConstDataVector, it might be useful to implement something similar to CxAODs (see this presentation).

Why? The code I had written saved systematic variations as shallow copies of some parent. This should always be true no? Why is it a deep-copy now? Is something broken?

kkrizka:
On March 4, 2016 08:41:34 AM Giordon Stark wrote:
> Isn't that why we add the parent with this? It would be very confusing for
> maintainability to have to manage two sets of configuration -- where if we
> want to add a shallow copy -- we add the shallow copy name to `configA` and
> then the parent to `configB`. it's much easier for the end-user to add
> shallow-copy + parent to `configA` because then you remove it all at once.
> This new way separates this out more and makes it harder to remove things
> that are matched up if you don't know that they are matched up.

The problem is that then you need to figure out how to save the parent 
container. For example, the jet systematic container is a shallow copy of the 
nominal calibration, which is a shallow copy from the input xAOD. If you add 
`jetSys|jetNom` to `m_shallowCopyKeys`, how do you save the `jetNom`? Some 
people might want to save it as a deep-copy, while others will want to save 
the shallow copy with the original xAOD container. I think supporting all 
possible configurations might get complicated.

> This is not always something you want. Users [including me] need to have
> some control over what the output container name should be. We can make it
> so that if an output name isn't specified, it uses the input name to copy
> over as well. Will this break if the user does something like
> 
> - CDV of AntiKt4EMTopoJets into TStore
> - deep copy CDV record into TEvent
> 
> assuming the TEvent has an input of the same name? This might be SUPER
> confusing.
That is a valid point. I tested this now and you cannot `TEvent::record()` a 
container name that exists in the input xAOD. I'll readd the ability to rename 
containers.

Is this something one would want for all copy methods?

> Is this a one-stop config that can handle all cases?
> 
That's the goal.

> > The systematic variation containers are saved as deep copies. Not ideal in
> > terms of disk space. Before ASG allows us to save ConstDataVector, it
> > might be useful to implement something similar to CxAODs (see this
> > presentation).
> Why? The code I had written saved systematic variations as shallow copies of
> some parent. This should always be true no? Why is it a deep-copy now? Is
> something broken?
> 
I assume you are referring to `m_vectorCopyKeys`. I don't it worked.
- It assumed that the aux container is of the corresponding particle container 
type. This is not true of shallow copies, which is of type 
`ShallowAuxContainer`.
- The outputs of calibrators/selectors are `ConstDataVector`, which do not 
have an aux container.

The second point is the important one. We need to be able to save 
`ConstDataVector`, since they are the ones that are pT sorted/slimmed. Doing a 
deep copy is the simplest solution for now.

--
Karol Krizka

kratsg:
> The problem is that then you need to figure out how to save the parent
> container. For example, the jet systematic container is a shallow copy of the
> nominal calibration, which is a shallow copy from the input xAOD. If you add
> `jetSys|jetNom` to `m_shallowCopyKeys`, how do you save the `jetNom`? Some
> people might want to save it as a deep-copy, while others will want to save
> the shallow copy with the original xAOD container. I think supporting all
> possible configurations might get complicated.

Oof, Good point. We will have to add some educational materials to make sure users know how to handle this themselves. This is definitely not a "beginner-level" tool to use anymore.

> Is this something one would want for all copy methods?

We probably will need to have this lest we run into problems later because we're not entirely configurable? Perhaps we can standardize the format string for all copy methods. Such as...

- `*` to force a deep-copy, and otherwise use the default behavior
- `a>>b` will use inputs `a` to create outputs `b`

and so on... at least this would make all of the parsing a bit easier if we standardize how we parse?

> I assume you are referring to `m_vectorCopyKeys`. I don't it worked.
> - It assumed that the aux container is of the corresponding particle container
> type. This is not true of shallow copies, which is of type
> `ShallowAuxContainer`.

Huh. You're right. This is going to be problematic.

> The second point is the important one. We need to be able to save
> `ConstDataVector`, since they are the ones that are pT sorted/slimmed. Doing a
> deep copy is the simplest solution for now.

There's no other solution. A shallow copy works because we don't remove or reorder the elements of the shallow copy. A CDV, by definition, allows us to remove elements and sort things in a particular order. This effectively means we _have_ to make a deep-copy. This is the case that I already forseen as being a problem when you have a systematic jet object that passes some threshold that the nominal doesn't.

In my mind, you cannot make an easy subset of those objects and still consistently make shallow copies. A way around this, while maintaining shallow copies, is to use decorations to handle which objects have been selected on a given systematic.

kkrizka:
I will implement the following format for specifying which containers to copy: `inputContainer*>>outputContainer|systematicsVector`. The `>>outputContainer` will be optional. Also `|systematicsVector` will only work for `m_particleCopy`, just like now. `*` is also optional and only for `m_shallowCopy`. It will specify `shallowIO=false`.

kratsg:
> I will implement the following format for specifying which containers to copy: inputContainer*>>outputContainer|systematicsVector. The >>outputContainer will be optional. Also |systematicsVector will only work for m_particleCopy, just like now. * is also optional and only for m_shallowCopy. It will specify shallowIO=false.
>  Show all checks

:+1:

kkrizka:
Readded ability to rename containers.

Also I noticed a problem when saving systematics that needs to be addressed. The particle selectors remove systematics that have no particles passing the selection. This means that some events won't have the corresponding containers saved and `TEvent::fill` will complain. I think the best way to fix this is to add an option to the selectors that disabled this behavior.

kratsg:
> Also I noticed a problem when saving systematics that needs to be addressed. The particle selectors remove systematics that have no particles passing the selection. This means that some events won't have the corresponding containers saved and TEvent::fill will complain. I think the best way to fix this is to add an option to the selectors that disabled this behavior.

How can we address this though? This requires not using CDVs unless we just deep copy everything (which is inefficient in the long run).

kkrizka:
The current iteration of his branch produces usable MinixAODs. However turning them into histograms is a bit slow. An empty (deleted everything from `execute()`) `BasicEventSelection` runs at about 20 kHz, dropping to 6 kHz when I just add retrieval of the `EventInfo` container. Processing NTUPles and making histograms runs at 70kHz. This is with both NTUPles and MinixAOD inputs stored at `UC_MWT2_LOCALGROUPDISK`, so the difference isn't due to different storage.

If there are no more complaints, I will complete my work on this by rebasing and merging tonight.

kratsg:
> A side question: how do you process Ntuples with xAH?

https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/EventLoop#Basic_access_using_TTree_and_ent

Notice that you can only have _one_ algorithm when using TTree access (unfortunately).

johnda102:
HI Marco,

  Here's an example:

https://svnweb.cern.ch/trac/atlasphys-exo/browser/Physics/Exotic/JDM/DiJetISR/Run2/Code/ZprimeDM/trunk/Root/ProcessZprimeMiniTree.cxx

https://svnweb.cern.ch/trac/atlasphys-exo/browser/Physics/Exotic/JDM/DiJetISR/Run2/Code/ZprimeDM/trunk/scripts/config_MiniTree_fullsim.py

On Fri, Mar 11, 2016 at 2:01 AM, Marco Milesi <notifications@github.com>
wrote:

> Hi @kkrizka <https://github.com/kkrizka> ,
>
> it's great to see new developments on this, as it could potentially
> improve analysis workflow a lot.
> I think it'd be important to make some documentation on how to use xAH to
> reprocess Mini xAODs, if it doesn't exist yet.
>
> A side question: how do you process Ntuples with xAH?
>
> —
> <https://github.com/UCATLAS/xAODAnaHelpers/pull/527#issuecomment-195123554>
> .
>



-- 
University of Chicago

kratsg:
@kkrizka - this is going to be a needed thing. Do we want to rebase and merge this in? It may need to be tested again.

kratsg:
I think for now, I'll close this.

PR #530, opened 2016-03-06
Allowing xAH_run with svn version.  Works with 2.3.44

jdandoy:
Adderssing an xAH_run.py bookkeeping complaint when used from svn.  Code works with 2.3.44.  Updated GRL list.

@gfacini @kratsg


kratsg:
Surprised nobody complained about this yet. Heh.

PR #532, opened 2016-03-07
Bump to 2.3.45, and start using AnaToolHandle

mmilesi:
See commit messages for details.

The bottom line is:

1. make the tool handle as a member of a xAH algorithm
2. in the xAH algorithm initialisation list, call the tool handle constructor. The argument of the constructor must be a string like  "TOOL_TYPE/TOOL_NAME" . In general,  "TOOL_NAME" here can be a dummy string, as it can be changed afterwards.
3. If the name of the tool needs to be changed, e.g. in `EL::initialize()` (this is the case for the efficiency correctors, where tool names get created depending on the configuration), just call `
my_tool_handle.make("TOOL_TYPE/NEW_TOOL_NAME");`

4. In `EL::initialize()` set the properties and initialise the tool handle. A NB: after `initialize()` has been called on the tool handle, it will effectively become a pointer to the tool itself.

5. In `EL::execute()`, use the tool via calls like  `my_tool_handle->doStuff()` If the tool with that given name needs to be used afterwards, just make a tool handle with the *same* tool_type/tool_name as intended, and call `initialize()` on it. It will automagically get the pointer to the correct tool, and all its properties will be preserved from the previous initialisation.

6. The tool will be automatically destroyed at the end of the job. Hence, no need to call `delete` anywhere.

The content of this PR shows an effective example for the above.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 env:
-  - ABV=2.3.44
+  - ABV=2.3.45
 
 before_install:
   - ls
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -1,18 +1,3 @@
-/********************************************************
- *
- * Basic event selection. Performs general simple cuts
- * (GRL, Event Cleaning, Min nr. Tracks for PV candidate)
- *
- * G. Facini (gabriel.facini@cern.ch)
- * M. Milesi (marco.milesi@cern.ch)
- * J. Dandoy (jeff.dandoy@cern.ch)
- * J. Alison (john.alison@cern.ch)
- *
- *******************************************************/
-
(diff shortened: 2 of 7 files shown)


kratsg:
Hi,

I think we're going to need to add some of these instructions to the xAH documentation somewhere. This is definitely a *gotcha* and is going to bite me in the butt later because I'm definitely not going to remember how to do this... and would have to remember to search the PRs for this content/message explaining how.

mmilesi:
Hi Giordon,

yep, sounds like a good thing to do.

Is https://xaodanahelpers.readthedocs.org/en/latest/FAQs.html the right place for it?

kratsg:
> Is https://xaodanahelpers.readthedocs.org/en/latest/FAQs.html the right place for it?

Yeah, this might be the best spot for now. It's super generic.

Issue #319, opened 2015-11-02
Derivation event counts / weights to output trees


johnda102:
How are people dealing with derivation input and output event counts and weights ?

Naively I just want the MetaData_EventCount histograms written to the tree output streams. 

The regular cutflow histograms appear there, but im actually confused as to where the cut flow histograms are told which output files to be written to.  Any clues as to where I should look ?

kratsg:
Bookkeeper Cutflows does this for you.

johnda102:
what do you mean by bookkeeper cutflows ?

How do I configure it to write to the same output file I write the trees to
?

On Tue, Nov 3, 2015 at 12:38 AM, Giordon Stark <notifications@github.com>
wrote:

> Bookkeeper Cutflows does this for you.
> On Mon, Nov 2, 2015 at 18:16 johnda102 <notifications@github.com> wrote:
>
> > How are people dealing with derivation input and output event counts and
> > weights ?
> >
> > Naively I just want the MetaData_EventCount histograms written to the
> tree
> > output streams.
> >
> > The regular cutflow histograms appear there, but im actually confused as
> > to where the cut flow histograms are told which output files to be
> written
> > to. Any clues as to where I should look ?
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/UCATLAS/xAODAnaHelpers/issues/319>.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/UCATLAS/xAODAnaHelpers/issues/319#issuecomment-153193797>
> .
>



-- 
University of Chicago
http://hep.uchicago.edu/~johnda

kratsg:
See an example here: https://svnweb.cern.ch/trac/atlasphys-susy/browser/Physics/SUSY/Analyses/MultiBJets/MultibjetsAnalysis/trunk/Root/MainAnalysis.cxx#L198

If you make a new histogram, just create it in `finalize()` and use `SetDirectory()` to set it to the output file you want to store it in.

johnda102:
ok thanks @kratsg. 

Any reason to not write this info by default to the standard m_cutflowHist and m_cutflowHistW hists ?

kratsg:
> Any reason to not write this info by default to the standard m_cutflowHist and m_cutflowHistW hists ?

Nope. But I think this is resolved now. Close it if you need more help.

PR #534, opened 2016-03-07
Making BasicEventSelection cutflow and metadata output stream names configurable

lawrenceleejr:
This PR is related to #319. I've simply made the names of the BasicEventSelection output streams ("cutflow" and "metadata") configurable so that those histograms can be sent to other streams that the user might have. It's a small change that should make things a little cleaner for analyses wanting control over the streams.

tl;dr PR let's you cross the streams.

### DIFF
--- a/xAODAnaHelpers/BasicEventSelection.h
+++ b/xAODAnaHelpers/BasicEventSelection.h
@@ -75,6 +75,10 @@ class BasicEventSelection : public xAH::Algorithm
     std::string m_derivationName;
     bool m_useMetaData;
 
+    // Output Stream Names
+    std::string m_metaDataStreamName;
+    std::string m_cutFlowStreamName;
+
     /* Check for duplicated events in Data and MC */
     bool m_checkDuplicatesData;
     bool m_checkDuplicatesMC;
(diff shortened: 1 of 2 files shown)


kratsg:
go ahead :ghost:

xAODAnaHelpers/BasicEventSelection.h
```diff
@@ -75,6 +75,10 @@ class BasicEventSelection : public xAH::Algorithm
     std::string m_derivationName;
     bool m_useMetaData;
 
+    // Output Stream Names
```
kratsg: You need to change this to be

```/* Output Stream Names */```

so that doxygen can parse it correctly.

File: xAODAnaHelpers/BasicEventSelection.h [hunk 2]
```diff
@@ -75,6 +75,10 @@ class BasicEventSelection : public xAH::Algorithm
...
+    std::string m_metaDataStreamName;
+    std::string m_cutFlowStreamName;
```
kratsg: This needs a comment above it as well describing it.

xAODAnaHelpers/BasicEventSelection.h
lawrenceleejr: Huh - was just symmetrizing with what was already there, but can do.

lawrenceleejr:
How's that @kratsg?

kratsg:
@lawrenceleejr that looks better, thanks.

PR #535, opened 2016-03-07
xAH_run: PrunDriver's mergeOutput option is a float

kkrizka:
Reference: https://svnweb.cern.ch/trac/atlasoff/browser/PhysicsAnalysis/D3PDTools/EventLoopGrid/trunk/python/ELG_prun.py#L55


kratsg:
Thanks for the catch. These things are SUPER random and not obvious.

PR #546, opened 2016-03-16
Updating trigger access

johnda102:


kratsg:
Can you document the changes in the header as well? It looks like these changes are not obvious about how you use it.

johnda102:
Sure. 
The changes should be transparent to the end user

johnda102:
these checks are taking a awfully long time WTF

kratsg:
I've emailed Lincoln and Judith about it.

PR #561, opened 2016-04-02
Fix for missing primaryVertex

mmilesi:
Need to make sure the vertex is retrieved ONLY if the index is not -1

Unfortunately I have found in my latest DAODs that a few events do not have a primary vertex (which is a bit worrying, but that's it...NB: in our analysis we do not apply the event cut on the priVtx quality), so this check is vital.

I have also added some warning messages in case the primary vertex is not found.


kratsg:
Is this related to #563?

mmilesi:
Hi @kratsg ,

oh yes, it looks like it's tightly related to that.

mmilesi:
Can we merge this?

PR #566, opened 2016-04-05
TTrees produced with systematics!

### LINKED ISSUE #558: Make TreeAlgo handle systematics like a bawss
Shouldn't be too hard theoretically. It's chaining things down the line as usual.

Example config: https://github.com/US-ATLAS-HFSF/HFSF2015/blob/master/example/gamma_b.py

Code looks like it changes from

```
  // algo
  TTree * outTree = new TTree(m_name.c_str(),m_name.c_str());
  TFile* treeFile = wk()->getOutputFile ("tree");
   m_helpTree = new HelpTreeBase( m_event, outTree, treeFile, m_units, m_debug, m_DC14 );
   outTree->SetDirectory( treeFile );
```

to

```
  for(const auto& syst: *systematic_list){
    // algo
    auto treeName = (m_name + syst.name());
    TTree * outTree = new TTree(treeName.c_str(),treeName.c_str());
    TFile* treeFile = wk()->getOutputFile ("tree");
    m_helpTree = new HelpTreeBase( m_event, outTree, treeFile, m_units, m_debug, m_DC14, syst );
    outTree->SetDirectory( treeFile );
  }
```

kratsg:
This should resolve #558 . I tested this locally using the [`gamma_b.py`](https://github.com/US-ATLAS-HFSF/HFSF2015/blob/master/example/gamma_b.py) file. The diff is below:

```
49c49,53
<                            "m_jetAlgo": "AntiKt4EMTopo"
---
>                            "m_JESUncertConfig": "$ROOTCOREBIN/data/JetUncertainties/JES_2015/Prerec/PrerecJES2015_AllNuisanceParameters_25ns.config",
>                            "m_jetAlgo": "AntiKt4EMTopo",
>                            "m_systName": "All",
>                            "m_systVal": 1,
>                            "m_outputAlgo": "JetSysts"
104c108,109
<                       "m_jetContainerName": "AntiKt4EMTopoJetsCalibOR",
---
>                       "m_jetContainerName": "AntiKt4EMTopoJetsCalib",
>                       "m_jetSystsVec": "JetSysts",
```

I didn't want to deal with threading systematics through OR just yet, so I just decided to dump the jets from `JetCalibrator` which looks like it works fine. The output tree file looks like so:

```
root [2] GammaB->ls()
TDirectoryFile*		GammaB	GammaB
 KEY: TTree	nominal;1	nominal
 KEY: TTree	JET_BJES_Response__1up;1	JET_BJES_Response__1up
 KEY: TTree	JET_BJES_Response__1down;1	JET_BJES_Response__1down
 KEY: TTree	JET_EtaIntercalibration_Modelling__1up;1	JET_EtaIntercalibration_Modelling__1up
 KEY: TTree	JET_EtaIntercalibration_Modelling__1down;1	JET_EtaIntercalibration_Modelling__1down
 KEY: TTree	JET_EtaIntercalibration_TotalStat__1up;1	JET_EtaIntercalibration_TotalStat__1up
 KEY: TTree	JET_EtaIntercalibration_TotalStat__1down;1	JET_EtaIntercalibration_TotalStat__1down
 KEY: TTree	JET_Flavor_Composition__1up;1	JET_Flavor_Composition__1up
 KEY: TTree	JET_Flavor_Composition__1down;1	JET_Flavor_Composition__1down
 KEY: TTree	JET_Flavor_Response__1up;1	JET_Flavor_Response__1up
 KEY: TTree	JET_Flavor_Response__1down;1	JET_Flavor_Response__1down
 KEY: TTree	JET_Gjet_Generator__1up;1	JET_Gjet_Generator__1up
 KEY: TTree	JET_Gjet_Generator__1down;1	JET_Gjet_Generator__1down
 KEY: TTree	JET_Gjet_OOC__1up;1	JET_Gjet_OOC__1up
```

and branches are set up correctly

```
root [4] ((TTree*)_file0->Get("GammaB/nominal"))->Scan("njets")
************************
*    Row   *     njets *
************************
*        0 *        10 *
*        1 *         7 *
*        2 *         8 *
*        3 *        11 *
*        4 *         8 *
*        5 *         7 *
*        6 *         9 *
*        7 *         8 *
*        8 *         4 *
*        9 *         5 *
************************
(Long64_t) 10
root [5] ((TTree*)_file0->Get("GammaB/JET_Flavor_Composition__1up"))->Scan("njets")
************************
*    Row   *     njets *
************************
*        0 *        10 *
*        1 *         7 *
*        2 *         8 *
*        3 *        11 *
*        4 *         8 *
*        5 *         7 *
*        6 *         9 *
*        7 *         8 *
*        8 *         4 *
*        9 *         5 *
************************
```

There should be some more sanity checks done, but there are no memory leaks, correct containers appear to be retrieved, and all is pretty good.

This also assumes that all container names being retrieved take the form of `m_inputContainerName+systName` for the most part (for jets, muons, electrons, and photons). This is not the most generic piece of code, but it makes some reasonable assumptions about rational users.

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -401,23 +401,13 @@ EL::StatusCode JetCalibrator :: initialize ()
     RETURN_CHECK("JetCalibrator::initialize()", m_JVTTool->initialize(), "");
   }
 
-
-  std::vector< std::string >* vecOutContainerNames = new std::vector< std::string >;
-
-
-
   for ( const auto& syst_it : m_systList ) {
     if ( m_systName.empty() ) {
       Info("initialize()","\t Running w/ nominal configuration only!");
       break;
     }
     Info("initialize()","\t %s", (syst_it.name()).c_str());
-
-    vecOutContainerNames->push_back( syst_it.name() );
   }
-
-  // add vector of systematic names to TStore
-  RETURN_CHECK( "JetCalibrator::initialize()", m_store->record( vecOutContainerNames, m_outputAlgo), "Failed to record vector of output container names.");
 
   return EL::StatusCode::SUCCESS;
 }
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -401,13 +401,23 @@ EL::StatusCode JetCalibrator :: initialize ()
     RETURN_CHECK("JetCalibrator::initialize()", m_JVTTool->initialize(), "");
   }
 
+
+  std::vector< std::string >* vecOutContainerNames = new std::vector< std::string >;
+
+
+
   for ( const auto& syst_it : m_systList ) {
     if ( m_systName.empty() ) {
       Info("initialize()","\t Running w/ nominal configuration only!");
       break;
     }
     Info("initialize()","\t %s", (syst_it.name()).c_str());
+
+    vecOutContainerNames->push_back( syst_it.name() );
   }
+
+  // add vector of systematic names to TStore
+  RETURN_CHECK( "JetCalibrator::initialize()", m_store->record( vecOutContainerNames, m_outputAlgo), "Failed to record vector of output container names.");
 
   return EL::StatusCode::SUCCESS;
 }
--- a/Root/TreeAlgo.cxx
+++ b/Root/TreeAlgo.cxx
@@ -139,12 +139,12 @@ EL::StatusCode TreeAlgo :: treeInitialize ()
 
   // get the file we created already
   TFile* treeFile = wk()->getOutputFile ("tree");
+  treeFile->mkdir(m_name.c_str());
 
   // let's make the tdirectory and ttrees
   for(const auto& systName: all_systNames){
     std::string treeName = systName;
     if(systName.empty()) treeName = "nominal";
-    treeName = m_name+"/"+treeName;
 
     Info("treeInitialize()", "Making tree %s", treeName.c_str());
     TTree * outTree = new TTree(treeName.c_str(),treeName.c_str());
@@ -157,9 +157,10 @@ EL::StatusCode TreeAlgo :: treeInitialize ()
     const auto& helpTree = m_trees[systName];
 
     // tell the tree to go into the file
-    outTree->SetDirectory( treeFile );
+    outTree->SetDirectory( treeFile->GetDirectory(m_name.c_str()) );
     // choose if want to add tree to same directory as ouput histograms
     if ( m_outHistDir ) {
+      if(m_trees.size() > 1) Warning("treeInitialize()", "You're running systematics! You may find issues in writing all of the output TTrees to the output histogram file... Set `m_outHistDir = false` if you run into issues!");
       wk()->addOutput( outTree );
     }
 
--- a/Root/TreeAlgo.cxx
+++ b/Root/TreeAlgo.cxx
@@ -88,6 +88,13 @@ EL::StatusCode TreeAlgo :: initialize ()
   return EL::StatusCode::SUCCESS;
 }
 
+EL::StatusCode TreeAlgo :: histInitialize ()
+{
+  Info("histInitialize()", "%s", m_name.c_str() );
+  RETURN_CHECK("xAH::Algorithm::algInitialize()", xAH::Algorithm::algInitialize(), "");
+  return EL::StatusCode::SUCCESS;
+}
+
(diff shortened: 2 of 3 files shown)


kratsg:
I have to sit and think about whether we need to allow for custom systematic container names... meaning the user passes in a string like

```
"{container}_{syst}"
```

and we substitute in the values, because they may not use a standard `"{container}{syst}"` but i dunno. Other common cases might be that they write `{container}{syst}ShallowCopy` or something more custom.

mswiatlo:
Any chance we can merge this @kratsg ? I'd like to use this in an upcoming data processing.

kratsg:
@mswiatlo - have you tested it yourself? You can run `git checkout add/ntupleSystematics` and make sure it looks good for you.

mswiatlo:
Not yet. I can do so if that’s what’s holding us back here. Will get back to you shortly!

Best,
Max

urania277:
I'm keen too and I've tested the jet systematics. I will need to think about how to do multiiple event-level variables containing jets in a TreeAlgo-inherited algo (my previous solution does not work anymore as it involved an initializeTree outside the execute) but I can work around that.

kratsg:
> I'm keen too and I've tested the jet systematics. I will need to think about how to do multiiple event-level variables containing jets in a TreeAlgo-inherited algo (my previous solution does not work anymore as it involved an initializeTree outside the execute) but I can work around that.

`treeInitialize` and `treeFinalize` were renamed to `histInitialize` and `histFinalize` to clean up and keep things uniform for all algorithms. But I don't know why you're inheriting TreeAlgo. You're not really meant to be inheriting that when you add your own code for ntuples -- you inherit from HelpTreeBase. @mmilesi could probably clarify this last point.

urania277:
It made sense at the time (I had event-level variables from 3 jet collections and other event-level variables that needed to be calculated somewhere, as in the case of the cross-sections) but I realize it's not ideal so I'm going to be rewriting that. What's @mmilesi's solution?

mmilesi:
Hi all,

@kratsg, @urania277 , I do indeed inherit from TreeAlgo. The main reason is that I define a set of custom methods:

```
AddLeptons()
FillLeptons()
ClearLeptons()
```
  
that are analysis-specific, and would then not fit into the base xAH classes. And I could need to define more in the future in case brand new "high-level objects" get defined in my analysis.

Basically the only 2 methods of `TreeAlgo` that I need to overload are `treeInitialize()` and `execute()`
Clearly, I have a custom class that inherits from `HelpTreeBase` as well.

I think we already discussed quite extensively about this when we went through the FW review at the ASG meetings, and we did *not* conclude people should not inherit from `TreeAlgo`.

Is this PR going to change the class interface so that inheritance wouldn't work anymore?
If that's the case, I'd refrain from merging now.

People who need urgently this could just check out the branch and work on that (they can even make a tag).

Cheers

Marco

jbossios:
Hi

Is this a fix for #481?

Cheers,
Jona

urania277:
I was talking about this with @boveia and he was suggesting the following:
- make a 'getTreeHelper' function that returns a handle to the HelpTree or to the inherited class) in TreeAlgo and its descendent. 
- call that in the execute()
- either do some casting if you have functions that are uniquely on TreeAlgo's descendent, or add extra virtual 'Add[Thing]' functions in HelpTreeBase. If you have all the Add[Thing] functions as hooks you probably don't need the getTreeHelper either.
I am a bit pressed for time right now but I am trying to find some time to prototype this or something like this, rather than hack extra things in HelpTreeBase as I'm doing in my branch of xAH to replace the lack of treeInitialize. 

Not sure what was discussed at the FW review - is there a summary somewhere?

kratsg:
> Is this PR going to change the class interface so that inheritance wouldn't work anymore?
> If that's the case, I'd refrain from merging now (if the change is just about method names, that's not a problem of course).

The change is just simply renaming the methods. The backwards-compatibility holds up just fine.

> Is this a solution for #481?

I ran this and saw no memory leaks on my end. I might have fixed it up in the process of adding systematics.



> - make a 'getTreeHelper' function that returns a handle to the HelpTree or to the inherited class) in TreeAlgo and its descendent.
> - call that in the execute()
> - either do some casting if you have functions that are uniquely on TreeAlgo's descendent, or add extra virtual 'Add[Thing]' functions in HelpTreeBase. If you have all the Add[Thing] functions as hooks you probably don't need the getTreeHelper either.

We have the `AddUser` and `FillUser` functions in `HelpTreeBase` and the `HelpTree` instance is a protected member so inherited classes can grab it, so you don't need a `getHelpTree` function.

PR #574, opened 2016-04-12
Fix Multiple WP sigmas for Jet systematics

jdandoy:
Remove the setSyst functions from Algorithm which were only used with the deprecated TEnv method (#573).  Parse the string of systematic sigma values in Algorithm, rather than passing a vector.

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -97,7 +97,7 @@ JetCalibrator :: JetCalibrator (std::string className) :
 
   // Initialize systematics variables
   m_systName                = "";
-  m_systVal                 = 0.;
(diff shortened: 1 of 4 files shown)


Root/JetCalibrator.cxx
```diff
@@ -301,6 +301,7 @@ EL::StatusCode JetCalibrator :: initialize ()
     Info("initialize()"," Initializing Jet Systematics :");
 
     //If just one systVal, then push it to the vector
+    this->parseSystValVector();
```
kratsg: Can we add a RETURN_CHECK here?

jdandoy: Good point, added with  #575

PR #587, opened 2016-05-13
Use AnaToolHandle in lepton selectors

mmilesi:
See commits for details


mmilesi:
@kratsg , could you please check why the bot is hanging?

kratsg:
I wonder if it's because connect is down. Let me bug someone.

PR #589, opened 2016-05-16
2.4.X upgrade

### LINKED ISSUE #543: deal with 2.4.X and 2.3X development
HI All, 
 
 Im seeing that there are a number of changes in base code we need to make for 2.4.X which are incompatible with 2.3.X (these are all related to underlying atlas tools changin in 2.4.X) 

How to we want to deal with this? Seems like we need to be able to support development in both releases.

kratsg:
@UCATLAS/xah - this is the pull request for upgrading. It was done by rebasing all latest changes in `master` up to `00-03-36` into `2.4.X` branch. The current master of `2.3.X` tagged at `00-03-36` will be in the `2.3.X` branch for those who still need 20.1 releases. We need to move over to a brand new version because this is a headache.

This will resolve #543 because we'll deprecate constant support of 2.3.X in favor of 2.4.X.


kratsg:
retest this please

PR #608, opened 2016-06-14
adding d0 decorator

alexandertuna:
For convenience downstream.


kratsg:
@alexandertuna -- just so I'm clear... Why isn't

```
electron->trackParticle()->d0()
```

convenient enough?

alexandertuna:
Hm, that looks pretty convenient. Gimme a sec.

alexandertuna:
Good call, I'm an idiot. Can we kill this PR?

kratsg:
💀 killed

PR #619, opened 2016-07-06
Add ntuple reader

johnda102:
Adding: 
  - NTuple readers classes (can alse for ntuple writing...next commit) 
  - Jet Hists filling based on NTuple classes

### DIFF
--- a/Root/IParticleHists.cxx
+++ b/Root/IParticleHists.cxx
@@ -122,3 +122,31 @@ StatusCode IParticleHists::execute( const xAOD::IParticle* particle, float event
   return StatusCode::SUCCESS;
 }
 
+
+StatusCode IParticleHists::execute( const xAH::Particle* particle, float eventWeight ) {
+
+  if(m_debug) std::cout << "IParticleHists: in execute " <<std::endl;
+
+  const TLorentzVector& partP4 = particle->p4;
+
+  //basic
+  m_Pt_l ->      Fill( partP4.Pt(),    eventWeight );
+  m_Pt ->        Fill( partP4.Pt(),    eventWeight );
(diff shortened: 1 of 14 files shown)


kratsg:
Wow, this is quite a major commit. What's goal here? There's also a lack of comments and the documentation needs updating for this.

johnda102:
This is doing two things. 
 -) Introduce object containers for reading ntuples produced with HelpTree Base 
 -) Extend the histogram classes to fill histogram classes based on these objects. 

I have another update where the object containers are also used to fill the data in HelpTree base (jetData -> jetContainer) so we ensure that the keep consistency from ntuple creation and reading.


I will update docs once I have the second update.

kratsg:
This is a really smart commit. I've already started doing something similar for L1Calo simulation samples.

https://gitlab.cern.ch/l1calo-gfex/KinematicPlots/blob/master/KinematicPlots/Objects.h

Note I might suggest / convince you to change from a TLV to using the `FourMom_t` base-class. This means you can initially set objects using pt, eta, phi, m from the initialization.

Giordon

johnda102:
I have no strong objections to using FourMom_t instread of TLorentzVector if all the relevant methods are availible

kratsg:
> I have no strong objections to using FourMom_t instread of TLorentzVector if all the relevant methods are availible

It's just a subset of the TLV. It just starts the TLV as a pt-eta-phi-m TLV rather than an t-x-y-z TLV by default. It's just a more sensible default for what people use and expect, I think.

johnda102:
Merging now,  the second commit will require more extensive testing, which I can not get to until the weekend.  (You can see work in progress in CoordinateNtupleWritingReading if people are interested)

kratsg:
I will file documentation issues.

PR #623, opened 2016-07-16
Coordinate ntup io

johnda102:


johnda102:
Giordon, you want to have a look at this. 

This is part 2/4.

3 will be an example for Muons

4 will be the docs

johnda102:
Now also added in part 3/4.

This moves the muons to using coordinated ntuple writing/reading.  

From these two examples it should be clear what one needs to do to coordinate other objects

Issue #617, opened 2016-06-22
Create new AsgAlgorithm base class in separate package


will-cern:
Hi,

I would propose we start the dual use effort by creating a separate package with just the AsgAlgorithm base class in it. Then make your base class inherit from this class.

This will become where we collect all dual-use algorithm features over time. If we keep this base algorithm in its own package, it will be easier for us to test compilation in the two environments too.

Best thing to do would be to create the package with it's algorithm using the cmt starter commands:

asetup AthAnalysisBase,2.4.12
cmt new_pkg AsgAlgorithms
cd AsgAlgorithms
cmt new_analysisalg AsgAlgorithm

That should compile for you in cmt already (cmt find_packages ; cmt compile).

Then import that wherever you want in github, and modify the algorithm. These things are necessary:

1. Move the algorithm into a 'Root' directory instead and the header into a header directory
2. ifdef the inheritance from AthAnalysisAlgorithm so that it inherits from EL::Algorithm if compiling with rootcore
3. think about how to solve this problem: https://its.cern.ch/jira/browse/ATLASG-494?filter=19067 ... the problem is that initialize in athena happens at a different stage to initialize in eventloop (the former is pre-event loop and the latter happens on the first event in the eventloop, I believe??? Does it run before fileExecute? What happens if the first file has no events and the user relies on things happening in initialize before happening in fileExecute?).
We can continue discussion here about the order in which methods of the algorithm should be called. In athena it will currently be:

initialize
<event loop starts>
beginInputFile
firstEvent   (this was introduced into the nightlies only this week)
execute
...
execute
<new file encountered>
beginInputFile
execute
...
<event loop ends>
finalize

will-cern:
The following 'features' of xAH::Algorithm are not dual use and need to be hidden inside methods in the base class if we are to make xAH::Algorithm run as an algorithm in athena.

1. m_event and m_store. These should be hidden inside an SgTEvent (https://svnweb.cern.ch/trac/atlasoff/browser/Control/AthToolSupport/AsgTools/trunk/AsgTools/SgTEvent.h), and then add to base class:

`
//this method already provided by AthAnalysisAlgorithm
#ifdef XAOD_STANDALONE
SgTEvent* evtStore() const { return &m_event; }
#endif
`

[...]

PR #633, opened 2016-08-05
Major changes to Muon and Electron algos to work in 2.4 release

mmilesi:
1. Use common Trigger Matching Tool for electrons and muons, plus some bug fixes.
2. Revert to using CP tool pointers instead of AnaToolHandles in MuonEfficiencyCorrector until Nils fixes the issues w/ re-using the same tool with different ATHs.

This comes with quite some delay, apologies!

See commits for more details.


kratsg:
The only thing that seems worrisome to me is commit 727a247 . Maybe @will-cern knows what's up here.

mmilesi:
Hi Giordon,

can you please argument on what's wrong with that commit?

kratsg:
Well, we're moving away from AnaToolHandle which is not what I was hoping (if you look at #617 ) -- we should move closer to trying to get differences between RootCore and Athena squashed.

mmilesi:
Unfortunately this AnaToolHandle business is still not completely under control of ASG.
Specifically, if you look at the "Issues retrieving AnaToolHandle configured in separate class
" thread on PATHelp, I am trying to figure out with Nils why some unexpected behaviour appears  when trying to configure ATHs with the same tool name.
He's currently debugging, hope to have more updates soon.

For now, since people (e.g. Federico from Melbourne) need to progress, I had to revert back to the "old" setup, at least where needed.

Marco

mmilesi:
Thanks for merging @kratsg 

FYI, Nils has made a major rewriting of the AnaToolHandle class that I could test in the 2.5.X nightlies.
It seems to finally work as it should!

He told me the ASG plan is now to backport this change also in the latest 2.4 releases, so we could hopefully make the switch to ATH everywhere soon. This should be discussed at the next ASG meetings.

Marco

jdandoy:
Hey @mmilesi ,

Was this change implements, and can we revert back to using AnaToolHandles here?

kratsg:
The change was implemented. We should revert back entirely as of 2.4.20.

PR #644, opened 2016-08-11
Support for multiple jet collections

jroloff:
Changed from single fatjet collection to map of fatjet collections. Branches are printed to the tree with the name of each collection at the end. This addresses issue #544 ... I've tested that it does work, but feel free to do it a different way as well. 

Currently, you have to use the same detail string for each jet collection, which is fine for studies comparing the performance of different jet collections -- you usually want to compare the same observables for each collection.


jroloff:
Also, here's an example of a config file:

```
c.setalg("TreeAlgo", {"m_debug": False,
                      "m_name": "OutputTree",
                      "m_fatJetContainerName": "SoftDrop_Beta1_Z01_Jets SoftDrop_Beta2_Z02_Jets SoftDrop_Beta2_Z01_Jets CamKt12LCTopoJets",
                      "m_truthJetContainerName": "CamKt12TruthJets",
                      "m_fatJetDetailStr": "kinematic energy scales",
                      "m_truthJetDetailStr": "kinematic energy scales",
                      "m_evtDetailStr": "pileup shapeEM shapeLC truth"
                    })
```

kratsg:
Before I start testing this -- we'll need to get the code cleaned up a bit and the logic sorted out. But thanks! Here are my comments for fixes:

- decorators should be prefixed with `dec_`. Accessors with `acc_`. For example, you have

  ```
  static SG::AuxElement::ConstAccessor<float> Split12("Split12");
  ```

   which should be changed to

  ```
  static SG::AuxElement::ConstAccessor<float> acc_Split12("Split12");
  ```

- `HTB::AddFatJets` should take in a single container and a single detail string to provide maximum flexibility. Other people write their own custom algorithm that initializes HelpTreeBase. Look at how `HTB::AddTruthParts` is written - making it look like that allows you to add in multiple groups of objects there

- Move the splitting of the string into the algorithm itself [eg: configuring with `"obj1 obj2 ..."`] to split into different container names. That logic should be consistently handled in the algorithm similar to how you do your `FillFatJets` so far.

- Please don't use the `TString::Form` function as this means we're adding a strong dependency on ROOT. See my next comment on how to handle this in a subjectively better way.

- Aim for backwards compatibility. The original function is

  ```
  void HelpTreeBase::AddFatJets(std::string detailStr)
  ```

  and we don't want this change to break other code people have, so do the following:

  ```
  void HelpTreeBase::AddFatJets(std::string detailStr, std::string fatjetContainerName="", std::string suffix="")
  ```

  and then simply do

  ```
  if(suffix.empty()){ suffix = fatjetContainerName };

  // lots of code...

  m_tree->Branch(("fatjet_E"+suffix).c_str(),   &m_fatjet_E);
  ```

  everywhere appropriate most likely. This gives output that people get before the change normally (I think). This also won't break anyone's code when they update xAH and try to compile.

kratsg:
Great! Ok to test

jroloff:
Wait a couple minutes... I forgot to change one thing, but I'm about to commit it

jroloff:
Ok, should be good to go now!

kratsg:
> Ok, should be good to go now!

Good catch. Don't worry about overwhelming the build system, it cancels a previously running test and restarts if you make a new commit.

PR #645, opened 2016-08-17
Update JetSelector.cxx

AvivCukierman:
According to the current jet cleaning recommendations (https://twiki.cern.ch/twiki/bin/view/AtlasProtected/HowToCleanJets2016), events should be removed if there are jets that pass JVT that are bad. The old functionality would remove an event if it contained a pileup jet that was bad.

@mswiatlo @kratsg @gfacini


kratsg:
test this please

kratsg:
test this please

PR #651, opened 2016-08-29
Default value for fatJet branches

mmilesi:
Set a default value to `fatjetName` in the method definition:

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/xAODAnaHelpers/HelpTreeBase.h#L109

so that the user is not obliged to pass it as argument if not needed. Something similar to what `FillJets` does.


kratsg:
test this please

Issue #660, opened 2016-09-02
HelpTreeBase fatjet container names / suffix don't work properly


cshimmin:
The `AddFatJets` function is doing something weird. The function parameters accept both a container (i.e. prefix) name, as well as a suffix name. However the actual implementation either treats the container name as the suffix or just ignores it all together. In either case, there is no way to specify the container/prefix name. Seems like this was maybe a misunderstanding/misimplementation of @kratsg's comment about backwards compatibility in #644. Relevant code here:
https://github.com/UCATLAS/xAODAnaHelpers/blob/5cb612bb94521f86a19d412af9a38b48831ece9b/Root/HelpTreeBase.cxx#L1331

I'd be happy to submit a PR to fix it if it would be accepted, but it might break some code depending on this functionality. One compromise would be to change the signature to `AddFatJets(std::string detailStr, std::string suffix="", std::string fatjetContainerName = "")` (i.e. commuting the second two arguments); this could be implemented "as advertised" and still maintain backwards  compatibility with anyone using the code since #644.

PR #661, opened 2016-09-02
fix AddFatJets[User]() to support multiple container names as well.

cshimmin:
Support for multiple fatjet collections was added in #644 but the relevant arguments were not propagated to `HelpTreeBase::AddFatJetsUser()`, preventing user-added extensions from knowing which jet collection they're supposed to be writing to. This PR fixes things in a functionally backwards-compatible way; however existing users implementing `AddFatJetsUser()` will have to update their function signatures to expect the additional arguments.

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -1343,7 +1343,8 @@ void HelpTreeBase::ClearTruth(const std::string truthName) {
  *
  ********************/
 
-void HelpTreeBase::AddFatJets(std::string detailStr, std::string fatjetContainerName, std::string suffix) {
+void HelpTreeBase::AddFatJets(const std::string& detailStr, std::string fatjetContainerName,
+        std::string suffix) {
 
   if(m_debug) Info("AddFatJets()", "Adding fat jet variables: %s", detailStr.c_str());
 	if(suffix.empty()){ suffix = fatjetContainerName; }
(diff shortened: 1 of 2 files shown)


cshimmin:
Actually this is a bigger mess than I initially thought... also the `FillFatJetsUser()` and `ClearFatJetsUser()` methods need to be updated. Please hold on this PR while I update my branch...

cshimmin:
Okay, the relevant changes should be propagated through the Add/Fill/Clear functions. I've tested it successfully with my own functions overridden in user code.

kratsg:
go ahead and test this @xAH-bot

kratsg:
Will this close #660 ?

kratsg:
@cshimmin can you rebase your changes on top of the latest in master? You're working on a fork, so just do

```
git remote add upstream git@github.com:UCATLAS/xAODAnaHelpers
git fetch upstream
git rebase upstream/master
git push -f origin helptree-fix-fatjets
```

and this will be retested and then merged!

cshimmin:
Will rebase soon but it does not close #660, this just at least makes the user's fatJets work at all in a way that is consistent with the current naming scheme. More work needs to be done to fix #660 which I'm happy to take care of if you think it could be merged (even if it may break some backwards compatibility).

kratsg:
> Will rebase soon but it does not close #660

ok



> More work needs to be done to fix #660 which I'm happy to take care of if you think it could be merged (even if it may break some backwards compatibility).

if you submit another pull request later on (separate from this one) - I can think about the backwards compatibility more.

PR #667, opened 2016-09-03
Refactor HelpTreeBase fatjets

### LINKED ISSUE #660: HelpTreeBase fatjet container names / suffix don't work properly

cshimmin:
Closes #660. However some backwards-compatibility will be broken:

 * Any existing implementers of `[Add/Fill/Clear]FatJetsUsers()` will need to update these functions to accept both collection name and suffix in the call signature. However until PR #661 (accepted yesterday) these functions were impossible to implement successfully anyways, so I doubt there are any affected users "in the wild".

 * Existing users of the built-in `[Add/Fill]FatJets()` functions will experience a change in the names of output branches unless they update their code. I expect this will affect quite a few users, who may need to update code that does further processing of the output ntuples and/or makes plots. But the whole point of the PR is to fix the broken branch names.

For example, users who currently call `AddFatJets(detailStr)` with no additional arguments, will see the branch `fatjet_E_` change to `fatjet_E`.

Users who currently call `AddFatJets(detailStr, "foo")` will see the branch `fatjet_E_foo` change to `foo_E`.

Users who currently call `AddFatJets(detailStr, "foo", "bar")` will see the branch `fatjet_E_bar` change to `foo_E_bar`.

Also, this PR fixes inconsistency in the naming of branches by `AddFatJets()` wherein the suffix is sometimes separated by an underscore and sometimes not.

One possible mitigation of the backwards incompatibility would be to commute the last two arguments (suffix, prefix) so that at least users calling the `AddFatJets(detailStr, "foo")` version of the function will have their branch names _mostly_ left unchanged (i.e. `fatjet_E_foo` -> `fatjet_E_foo` but `fatjet_numConstituentsfoo` -> `fatjet_numConstituents_foo`).

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -1343,54 +1343,72 @@ void HelpTreeBase::ClearTruth(const std::string truthName) {
  *
  ********************/
 
-void HelpTreeBase::AddFatJets(const std::string& detailStr, std::string fatjetContainerName,
-        std::string suffix) {
+// make a unique container:suffix key to lookup the branches in the maps
+std::string HelpTreeBase::FatJetCollectionName(const std::string& fatjetName,
+    const std::string& suffix) {
+  return suffix.empty() ? fatjetName : (fatjetName + ":" + suffix);
+}
+
+void HelpTreeBase::AddFatJets(const std::string& detailStr, const std::string& fatjetName,
+        const std::string& suffix) {
 
   if(m_debug) Info("AddFatJets()", "Adding fat jet variables: %s", detailStr.c_str());
-  if(suffix.empty()){ suffix = fatjetContainerName; }
 
+  // Simple function to formulate the branch names given the variable name.
+  // Eliminates a lot of verbosity in the rest of the function...
+  auto bname = [&](const std::string& varName) {
+    std::string name = fatjetName + "_" + varName;
+    if (not suffix.empty()) {
+      name += "_" + suffix;
+    }
+    return name;
+  };
+
+  const std::string collectionName = FatJetCollectionName(fatjetName, suffix);
 
   m_fatJetInfoSwitch = new HelperClasses::JetInfoSwitch( detailStr );
 
   // always
-  m_tree->Branch(("nfatjets"+suffix).c_str(),    &m_nfatjet[suffix],"nfatjets/I");
+  std::string counterName = "n"+fatjetName;
+  if (not suffix.empty()) { counterName += "_" + suffix; }
+  m_tree->Branch(counterName.c_str(), &m_nfatjet[collectionName], (counterName+"/I").c_str());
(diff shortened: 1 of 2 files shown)


kratsg:
I think this is ok. Go ahead and test @xAH-bot . The backwards compatibility will need to be broken since the older behavior is clearly not good enough.

kratsg:
Great! One last thing if you don't mind an extra commit. Can you briefly document some of the following functions you've updated? All of the comments would go into the `xAODAnaHelpers/HelpTreeBase.h` header file:

- `void AddFatJets     (const std::string& detailStr = "", const std::string& fatjetName = "fatjet", const std::string& suffix = "")`
- `static std::string FatJetCollectionName(const std::string& fatjetName = "fatjet", const std::string& suffix = "")`
- `void FillFatJets( const xAOD::JetContainer* fatJets , const std::string& fatjetName = "fatjet", const std::string& suffix = "")`

Look at how `Algorithm.h` is commented (https://github.com/UCATLAS/xAODAnaHelpers/blob/master/xAODAnaHelpers/Algorithm.h#L78-L82) for some inspiration (the `setName` function's level of documentation will be sufficient here).

You can also expand with the extra details mentioned in your PR above if you wish.

cshimmin:
No problem but I'm out the door, probably for the evening. So I'll ping the
thread when it's done.

-Chase

kratsg:
> No problem but I'm out the door, probably for the evening. So I'll ping the
> thread when it's done.

No rush. Just push the commit/changes and the bot will ping the thread for all of us when it's finished testing 😄

kratsg:
All good, thanks! Docs are being rebuilt now.

kratsg:
Should've pointed this out earlier: https://xaodanahelpers.readthedocs.io/en/master/HelpTreeBase.html#_CPPv2N12HelpTreeBase10AddFatJetsERKNSt6stringERKNSt6stringERKNSt6stringE

PR #678, opened 2016-09-19
Fix small bug when checking Metadata on full (x)AOD

mmilesi:
See commit message.


mmilesi:
@kratsg , the bot was hanging on the code quality check stage. I went ahead and merged, since it's a very simple one.

kratsg:
It's ok if the code quality hangs. That's not a necessary check for merging :) It'll run even after merging.

PR #683, opened 2016-09-27
Fix default pt selection for JVT to 60 GeV

jdandoy:
As detailed at https://twiki.cern.ch/twiki/bin/view/AtlasProtected/JVTCalibration


kratsg:
@jdandoy -- is this documented in our docs? https://xaodanahelpers.readthedocs.io/en/master/ToolsUsed.html

jdandoy:
Both the updated and the outdated JVT recommendations are linked in ToolsUsed.  I've gotten rid of the outdated one in #684

PR #684, opened 2016-09-27
Protect Warnings

jdandoy:
Protect warnings from #676 

When the PRW tool returns a random run number of 0 the else statement produces a bunch of Warnings, though the behavior is expected.  
```
MuonEfficiencyCorrecto... WARNING Random runNumber generated outside hardcoded run number ranges
MuonEfficiencyCorrecto... WARNING Setting the year as randomly chosen in the years list
```


kratsg:
Thanks a lot!

PR #685, opened 2016-09-28
Etmiss syst implementation (for TST met and calo jets)

martaperego:
Etmiss systematics are now implemented in METConstructor:
- TST soft term syst are implemented using METSystematicsTool
- systematics for electrons, photons, muons, jets (calo jets) are propagated to the met
- systematics for taus and for track jets are not yet propagated

### DIFF
--- a/Root/ElectronCalibrator.cxx
+++ b/Root/ElectronCalibrator.cxx
@@ -217,13 +217,17 @@ EL::StatusCode ElectronCalibrator :: initialize ()
   m_systList = HelperFunctions::getListofSystematics( recSyst, m_systName, m_systVal, m_debug );
 
   Info("initialize()","Will be using EgammaCalibrationAndSmearingTool systematic:");
+  std::vector< std::string >* SystElectronsNames = new std::vector< std::string >;
   for ( const auto& syst_it : m_systList ) {
     if ( m_systName.empty() ) {
       Info("initialize()","\t Running w/ nominal configuration only!");
       break;
     }
+    SystElectronsNames->push_back(syst_it.name());
     Info("initialize()","\t %s", (syst_it.name()).c_str());
   }
+
+  RETURN_CHECK("ElectronCalibrator::initialize()",m_store->record(SystElectronsNames, "ele_Syst" ), "Failed to record vector of ele systs names.");
 
   // ***********************************************************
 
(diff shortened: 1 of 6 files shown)


kratsg:
go ahead @xAH-bot

jdandoy:
Hi @martaperego , @kratsg 

I get a lot of warnings with this update, one warning each time the version with sysListItrString is not found in m_store and the nominal is instead taken.  This only occurs for Photon and TauJets containers, which I general don't use anywhere but for MET and so are in m_event.  Can this warning be removed?  

Also I don't see why so many retrieve calls are needed for each container, (i.e. around [line 383](https://github.com/UCATLAS/xAODAnaHelpers/pull/685/files#diff-526772a862575336b3a7c3f07f54ab5bR383) )

The Warning:
```
xAOD::TStore::retrieve  WARNING Couldn't (const) retrieve "DataVector<xAOD::Photon_v1>/Photons"
```

kratsg:
Hi @jdandoy,

Can you file an issue and link to #685? Looking at the fact that there's an existing issue (#687), we might have to recommend everyone to use 16cf354a47340d1f379b472e7a4db6ead00fd940 (the commit before this PR) until I can fix it later today -- and it would help if you detail an example fix in your filed issue.

martaperego:
Hi @kratsg , @jdandoy,

sorry, I saw now all the messages. 

The reason of all those warnings is that the systematics of all the objects are propagated to the Etmiss if those systematics have been previously stored in m_store, otherwise you just use the nominal case. 
If you have those warnings just for photons and taus it means that for those objects you do not have the corresponding systematic-varied objects. Is it correct?

To check if the systematics are in m_store I was using the following line of code:
_if ( m_store->retrieve(eleCont ,  m_inputElectrons.Data() +sysListItrString ).isSuccess() ) { // check if the syst are in m_store_
     _RETURN_CHECK("METConstructor::execute()", HelperFunctions::retrieve(eleCont, m_inputElectrons.Data()+sysListItrString, m_event, m_store, m_debug), "Failed retrieving electron cont."); // retrieve the syst
(...)
}_

is there another way to do this in order to avoid the list of warnings?

kratsg:
@martaperego it's because you're recording it to the TStore without a unique name for each systematic really.

Issue #679, opened 2016-09-23
Alternative electron working points


MihaMuskinja:
Dear developers,

what is the easiest way to add additional working points for electrons to the output ntuples? A concrete example would be isolation working points. Looking at the HelpTreeBase it seems that only two working points are defined for isolation:

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelpTreeBase.cxx#L633
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelpTreeBase.cxx#L634

what should we do if we want to include additional isolation scale-factors to the output ntuples? There are 24 different isolation scale-factors:

http://atlas.web.cern.ch/Atlas/GROUPS/DATABASE/GroupData/ElectronEfficiencyCorrection/2015_2016/rel20.7/ICHEP_June2016_v3/isolation/

Best,
Miha

kratsg:
I think the best thing is to follow the template for the current working points and add in the others through a pull request 😄

MihaMuskinja:
Dear Giordon,

it is actually not so trivial to add isolation working points. The reason is that isolation scale-factors are a function of ID level and isolation, but the framework seems to be sensitive only to the isolation level.

Lets take this scale-factors as an example:

efficiencySF.Isolation.**LooseAndBLayerLLH**_d0z0_v11_**isolLoose**.root 

In the ElectronEfficiencyCorrector (https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L249) we can see that it would only read the isolation level from the string:

m_Iso_WP would be set to "IsoLoose" in the above case completely regardless of the ID level. m_Iso_WP is then used to decorate electrons (L302, L788, and L822).

This means that it is not trivial to write isolation scale-factors for two working points with the same isolation level (e.g. LooseAndBLayerLLH+isoLoose / MediumLLH+isoLoose) to the same output. However, this is what we would actually need in the output for the analysis.

Best,
Miha

kratsg:
Hi @MihaMuskinja -- indeed. I'm not very familiar with this portion for electrons. Indeed, it seems like the following combinations exist:

```
Lord Stark:~$ pbpaste | cut -d ' ' -f 5 | cut -d '.' -f 3 | cut -d '_' -f1,4
LooseAndBLayerLLH_isolFixedCutLoose
LooseAndBLayerLLH_isolFixedCutTight
[...]

PR #686, opened 2016-09-29
Update config parsing and SF decorations in ElectronEfficiencyCorrector

mmilesi:
This addresses #679


mmilesi:
This solves #679

PR #691, opened 2016-10-03
bump to 2.4.18

kratsg:
/cc @jdandoy


kratsg:
This will be merged before #690 .

PR #713, opened 2016-10-14
Add taus

johnda102:


kratsg:
@johnda102 be careful because this contained commits from a previous PR.

PR #714, opened 2016-10-17
add m_isMC for met syst tool

martaperego:
I added a flag "m_isMC" to use the met syst tool only on MC


johnda102:
@martaperego 

 I think you will need to rebase this branch and repush. 
If I understand from @kratsg  the update branch button no longer works. (This is a shame)

kratsg:
go ahead and test

PR #721, opened 2016-10-17
Make electron container

johnda102:


johnda102:
@MihaMuskinja 

Here is the commit with the ElectronContainer

MihaMuskinja:
thanks for the info @johnda102 , I will look into it.

PR #722, opened 2016-10-18
improved the loops over PID/ISO working points

MihaMuskinja:
Hi @johnda102 , @kratsg this was possible to do with the addition of the ElectronContainer.

This improves the way the loops over PID/iso working points work. Instead of running the loop over all the possible WPs and doing:

```
if (!m_infoSwitch.m_PIDWPs[PID]) continue;
if (!m_infoSwitch.m_isolWPs[isol]) continue;
```

each time, the user requested WPs are formed in the initialization of the ElectronContainer and the loops are therefore much shorter.

This should increase the speed of the algorithm, especially when we add another loop for the trigger names.

P.S.:
I'm sorry for copy-pasting a large chunk of the code, I had some problems with spaces/tabs in my text editor.

### DIFF
--- a/Root/ElectronContainer.cxx
+++ b/Root/ElectronContainer.cxx
@@ -103,6 +103,26 @@ ElectronContainer::ElectronContainer(const std::string& name, const std::string&
     m_trkPixdEdX                 = new std::vector<float> ();
   }
 
+  if ( m_infoSwitch.m_effSF && m_mc ) {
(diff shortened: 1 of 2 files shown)


johnda102:
ok to test

johnda102:
no sure what I need to do here ...

kratsg:
ok to test

PR #726, opened 2016-10-20
Bad jet event cleaning

jdandoy:
This update primarily deals with jet cleaning. The [cleaning recommendation](https://twiki.cern.ch/twiki/bin/view/AtlasProtected/HowToCleanJets2016) is to reject events with any clean jets after your event selection.  Jet cleaning therefore needs to be the last event selection checked.  I.e. if the jet fails JVT, then cleaning should not be checked.  

A new flag, m_cleanEvent, is introduced to remove events with one bad signal jet.  “m_cleanEvtLeadJets” (only removes event if first N jets are bad) is kept.
This also changes the order of the cutflow, but otherwise does not change the results for anyone who chooses not to use m_cleanEvent.

This update also removed the unprotected retrieve used in the METConstructor.
It also introduces an override of the AFII decision in ElectornCalibrator, in cases where the simulation metadata is not available in the derivation.

### DIFF
--- a/Root/ElectronCalibrator.cxx
+++ b/Root/ElectronCalibrator.cxx
@@ -77,6 +77,8 @@ ElectronCalibrator :: ElectronCalibrator (std::string className) :
   m_esModel                 = "";
   m_decorrelationModel      = "";
 
+  m_setAFII                 = false;
+
   m_useDataDrivenLeakageCorr = false;
 
 }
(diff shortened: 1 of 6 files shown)


kratsg:
@jdandoy Can I ask for one more thing? Can you add the jet cleaning twiki to this docs page?

- [docs/ToolsUsed.rst](https://github.com/UCATLAS/xAODAnaHelpers/blob/master/docs/ToolsUsed.rst)

Lastly, does this address any current issues on GH?

jdandoy:
I've added the requested information.  There was no issue for this change.

PR #738, opened 2016-10-26
Electron Selector Update

jdandoy:
This is a pretty big update to ElectronSelector to address #737 .  It'd be great if @mmilesi @fscutti  and @kratsg could take a look before it's added.  There may be a few more things to do, but this should be a start.

### DIFF
--- a/Root/ElectronSelector.cxx
+++ b/Root/ElectronSelector.cxx
@@ -874,8 +874,6 @@ int ElectronSelector :: passCuts( const xAOD::Electron* electron, const xAOD::Ve
   //
   float eta   = ( electron->caloCluster() ) ? electron->caloCluster()->etaBE(2) : -999.0;
 
-  int oq      = static_cast<int>( electron->auxdata<uint32_t>("OQ") & 1446 );
-
   // fill cutflow bin 'all' before any cut
   if(m_useCutFlow) m_el_cutflowHist_1->Fill( m_el_cutflow_all, 1 );
   if ( m_isUsedBefore && m_useCutFlow ) { m_el_cutflowHist_2->Fill( m_el_cutflow_all, 1 ); }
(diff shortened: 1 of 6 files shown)


johnda102:
Looks fine to me

Root/ElectronSelector.cxx
```diff
@@ -278,15 +276,15 @@ EL::StatusCode ElectronSelector :: initialize ()
 
   if ( // m_LHOperatingPoint != "VeryLoose"       &&
        m_LHOperatingPoint != "Loose"           &&
-       m_LHOperatingPoint != "LooseAndBLayer"  &&
+       m_LHOperatingPoint != "LooseBL"  &&
```
mmilesi: Any specific reason for changing the name of the WP?

Root/ElectronSelector.cxx
```diff
@@ -96,29 +96,27 @@ ElectronSelector :: ElectronSelector (std::string className) :
...
-  m_confDirPID              = "mc15_20150224";
 
   // likelihood-based PID
+  m_doLHPID                 = true;
   m_doLHPIDcut              = false;
   m_LHOperatingPoint        = "Loose";
-  m_LHConfigYear            = "2015";
 
   // cut-based PID
+  m_doCutBasedPID           = true;
```
mmilesi: Shall we set this to "false" here?

mmilesi:
Looks good to me as well (I have left few comments for you to address)

Root/ElectronSelector.cxx
jdandoy: LooseBL is the name of the [working point ](https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/EGammaIdentificationRun2#Usage_of_the_Electron_ID_tool)in AsgElectronLikelihoodTool.  I guess they changed it, and it's unfortunate that it's different from the electron efficiency naming convention, but we should stick with the tool naming convention.

fscutti:
all good to me as well

Root/ElectronSelector.cxx
mmilesi: I see. But if possible I would stick to the old "LooseAndBLayer" naming, as it is consistent w/ the naming of the electron SFs. You can always let the code internally make the conversion to the naming that works for the tool.
Do you agree?

jdandoy: I generally prefer for xAODAnaHelpers to not assume too much and to stick as close to the CP tools as possible.  But it's a bad situation and also breaks backwards compatibility, so I'll change it to accept both LooseBL and LooseAndBLayer.

mmilesi:
All good from my side, thanks!

PR #748, opened 2016-11-05
updated the PileupReweightingTool.DataScaleFactor

MihaMuskinja:
Updated as suggested by PATHelp. Will close https://github.com/UCATLAS/xAODAnaHelpers/issues/743.

Issue #743, opened 2016-11-02
PileupReweightingTool.DataScaleFactor needs to be updated


MihaMuskinja:
Dear All,

I'm just pasting the recent email from PATHelp:

> Dear all,
> 
> Following discussions at last week's ASG meeting and since, it has been agreed that the scale factor to correct for mismodelling of central activity, also known as the PRW DataScaleFactor, should be updated for analyses using luminosity calculated with Ofllumi-13TeV-005 lumitag. This will almost certainly be the case for any analyses going for winter conferences. Please ensure you have the following values set in your PileupReweightingTool:
> 
> PileupReweightingTool.DataScaleFactor = 1.0/1.09     #was previously 1.0/1.16
> PileupReweightingTool.DataScaleFactorUP = 1.0        #unchanged
> PileupReweightingTool.DataScaleFactorDOWN = 1.0/1.18 #was previous 1.0/1.23
> 
> The old factor was calculated with an older lumi tag, and these new numbers were calculated with the latest lumi tag. Hence the update. The twiki has been updated accordingly: https://twiki.cern.ch/twiki/bin/view/AtlasProtected/ExtendedPileupReweighting
> 
> Please note that versions of the PileupReweightingTool in the near future (timescale of a month or so) will drop these properties entirely, and will automatically set them for you on the basis of the lumitag of the lumicalc file you are using. But for now you must have them manually set.

It seems that the tool needs to be updated here:

https://github.com/UCATLAS/xAODAnaHelpers/blob/7d7117dcfd944a0cc8211363d6d2368d76d55ec7/Root/BasicEventSelection.cxx#L571

PR #749, opened 2016-11-07
updated PRW configuration

fscutti:
New configuration for pile-up reweighting. This would be identical to #748 which had problems merging.


fscutti:
Closed #748 and #743

PR #757, opened 2016-11-19
set toolhandles to have specific names

tongbaojia:
in hh4b we have two jet selectors, one for antikt4 and the other for lctopo10, hence the jettoolhandles need to have different names.


kratsg:
Hi, in order for this to be compatible with athena in the future -- we need to set these names in the constructor -- which is safest.

PR #768, opened 2016-12-20
remove improper static

johnda102:


johnda102:
Looks like the jenkins build is dead.  How can I get this restarted ?

PR #779, opened 2017-02-02
Save multiple jet collections with TreeAlgo

jbossios:
These changes allows to store more than one jet (reco or truth) collection with TreeAlgo. The backwards compatibility is maintained. 

Usage example:
      "m_jetContainerName": "AntiKt4EMTopoJets_Signal AntiKt2LCTopoJets_Signal",
      "m_jetBranchName": "RefJet 2LCJet",
      "m_truthJetContainerName": "AntiKt4TruthJets_Signal AntiKt2TruthJets_Signal",
      "m_truthJetBranchName": "Truth4Jet Truth2Jet",

When saving only one jet colletion there is no need to specify "m_jetBranchName" and "m_truthJetBranchName" and the standard names will be used ("jet" and "truthJet"). These variables are used to set the name of the branches for each one of the jet collections.

When running systematics, it's assumed that only the first jet collection is the one which needs to be variated.

These changes are crucial for the JetEtmiss R-scan team.

### DIFF
--- a/Root/TreeAlgo.cxx
+++ b/Root/TreeAlgo.cxx
@@ -45,6 +45,9 @@ TreeAlgo :: TreeAlgo (std::string className) :
   m_muContainerName             = "";
   m_elContainerName             = "";
   m_jetContainerName            = "";
+  m_jetBranchName               = "jet";
+  m_truthJetContainerName       = "";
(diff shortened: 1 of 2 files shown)


kratsg:
go ahead

PR #780, opened 2017-02-03
Allow not to force "_Insitu" in data (JetCalibrator.cxx)

jbossios:
These changes allow to not applied the insitu correction in data, which is needed for the R-scan and p-flow jets. The default still is to force the insitu calibration in data maintaining the exception on trimmed jets.


kratsg:
go ahead

PR #783, opened 2017-02-07
Fix in HelpTreeBase to save L1 jets in >2.4.22

jbossios:
This only changes the implementation of the storing of L1 jets in HelpTreeBase to make it work in AB >2.4.22.


kratsg:
go ahead

PR #804, opened 2017-02-17
Syst updates

jdandoy:
Bump to 2.4.28, and updates for b-tagging efficiency systematics strategy.

@fscutti @kratsg Also included are a local change I had to ElectronCalibration and MuonCalibrator to let me access the recommended uncertainties.  How were you currently accessing "All" uncertainties?  If there's already a better way to do it, I'll remove those two updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 env:
-  - ABV=2.4.27
+  - ABV=2.4.28
 
 before_install:
   - ls
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -57,6 +57,7 @@ BJetEfficiencyCorrector :: BJetEfficiencyCorrector (std::string className) :
   m_taggerName             = "MV2c10";
   m_useDevelopmentFile     = true;
   m_coneFlavourLabel       = true;
+  m_systematicsStrategy    = "SFEigen";
 
   // allowed operating points:
   // https://twiki.cern.ch/twiki/bin/view/AtlasProtected/BTaggingCalibrationDataInterface#xAOD_interface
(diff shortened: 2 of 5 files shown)


fscutti:
Hi @jdandoy, atm I am not accessing sys uncertainties. I would say I'm fine with what you propose. In the MuonCalibrator you might have some conflicts now as parts of your updates where already included in #803

MihaMuskinja:
Hi @jdandoy , 

sorry for a naive question, but how should the ElectronCalibrator be configured to get the systematics? Is there an example somewhere?

I tried with:
```
                           "m_decorrelationModel"         : "FULL_ETACORRELATED_v1",
                           "m_systName"                   : "All",
                           "m_systVal"                    : 0.0,
```
but the code crashes:

> ElectronEfficiencyCorr... ERROR   /afs/cern.ch/work/g/gorisek/miha/analysisFramework/xAODAnaHelpers/Root/ElectronEfficiencyCorrector.cxx:540 Failed to execute: HelperFunctions::retrieve(inputElectrons, m_inContainerName+systName, m_event, m_store, m_verbose)

jdandoy:
Hi @MihaMuskinja,

Thanks for testing this.  Here, the efficiency corrector is not able to find the container with the systematically shifted electrons, of container name m_inContainerName+systName. 

To debug, you can print the name of m_inputAlgoSystNames, and make sure it matches the list output by the electron calibrator (m_outputAlogSystNames).

Also, were you using systematics previously?  It's odd to have the systematic shift (m_systVal) set to 0 sigma, and yet request all systematic uncertainties.  This should generally be set to 1.0 if you're using systematic uncertainties, or else set m_systName to "None" if you're not.

MihaMuskinja:
Hey @jdandoy,

I never run it with systematics before, thanks for helping.. I set the val to 1.0 and now I'm debugging the names. I will let you know the result :)..

MihaMuskinja:
Hi @jdandoy,

I think the problem in our case is the order of the algorithms:

c.setalg("ElectronCalibrator",  ElectronCalibratorDict)
c.setalg("ElectronSelector",    ElectronSelectorDict)
c.setalg("OverlapRemover",      OverlapRemoverDict)
c.setalg("ElectronEfficiencyCorrector", generateElectronEfficiencyCorrector(path_el_eff,ID,isol,trig) )

ElectronEfficiencyCorrector is set to:  "m_inputAlgoSystNames"    : "ElectronSelector_Syst", however, there is the OR step in between. Could this be the problem?

jdandoy:
Hi @MihaMuskinja 

Yes, that is likely the culprit.   Each set of systematic uncertainties needs to be passed through each algorithm.  The list of output systematic names from the calibratior first needs to be passed into the selector.  The output of the selector then needs to be passed into the overlap remover.  And again the same for the overlap remover to the efficiency corrector.

I'll note that from the electron calibrator to the selector, only the electrons have a container for each electron systematic.  Once you reach the overlap remover, the electron systematic shifts can affect the jet, muon, and photon collectors as well.  So after the overlap removal stage, you'll have a set of jet, muon, and photon collections for each electron systematic shift (if you're using those objects).  Similarly, you'll have an electron container for each set of jet, muon, and photon systematic shifts.

MihaMuskinja:
hi @jdandoy,

OK, thanks. It seems that we have to set the 
```
m_inputAlgoElectrons = ""; // name of vector<string> of syst retrieved from TStore
m_inputAlgoJets = ""; // name of vector<string> of syst retrieved from TStore
m_inputAlgoMuons = ""; // name of vector<string> of syst retrieved from TStore
etc.
```
variables. Is there an example somewhere?

MihaMuskinja:
hi @jdandoy ,

never mind my previous comment, it just has to be set to
```
"m_outputAlgoSystNames"        : "ElectronCalibrator_Syst",
```
it works now! Thanks :+1:

Issue #802, opened 2017-02-15
Missing sstream headers


kratsg:
For the 2.5 releases, it appears there are compilation errors because the default headers linked have changed:

```
#include <sstream>
```

where we have 

```
std::istringstream
```

at least.

PR #811, opened 2017-02-23
Fix to compile in 2.5.X

johnda102:


kratsg:
This closes #802 . Also note that 2.5.X will be dropped. But we'll need this for 2.6.X which is CMake based too.

PR #816, opened 2017-02-28
Trees of Tracks

mattleblanc:
Adding functionality to get `InDetTrackParticles` from the `TreeAlgo`. This required adding the necessary container, etc.

Available `detailStr` options are currently `kinematic trackpars numbers vertex` ... these might not be the most optimal choices, but they cover the content that is available in the derivation I'm looking at, and should be easily expandable if others want to do that in the future.

🍻

### DIFF
--- a/xAODAnaHelpers/HelperClasses.h
+++ b/xAODAnaHelpers/HelperClasses.h
@@ -500,10 +500,10 @@ namespace HelperClasses {
         ================ ============== =======
         Parameter        Pattern        Match
         ================ ============== =======
-        m_kinematic      kinematic      exact
-        m_fitpars        fitpars        exact
+	m_kinematic      kinematic      exact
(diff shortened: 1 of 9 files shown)


xAODAnaHelpers/HelperClasses.h
```diff
@@ -495,6 +495,32 @@ namespace HelperClasses {
...
   /**
     @rst
+        The :cpp:class:`HelperClasses::InfoSwitch` struct for Track Information.
+
+        ================ ============== =======
+        Parameter        Pattern        Match
+        ================ ============== =======
+        m_kinematic      kinematic      exact
+        m_fitpars        fitpars        exact
+	m_numbers        numbers        exact
```
kratsg: Need to fix the spacing here. Documentation won't build.

PR #825, opened 2017-03-07
three more leaks

### LINKED ISSUE #824: More memory leaks
Hi @kratsg ,

unfortunately, our jobs are still failing due to excess memory usage. I looked for some more leaks, this time only in "execute()" step of the framework:

- https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelpTreeBase.cxx#L290

m_muonInfoSwitch is never deleted.
I suggest to add "delete m_muonInfoSwitch" in HelpTreeBase::~HelpTreeBase().

- https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetCalibrator.cxx#L460

This is never deleted I believe. I'm not even sure if this object is used at all.
I suggest to possibly delete lines 460 and 461.

- https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L647

These are still leaking as m_store->record( sysVariationNamesPID, m_outputSystNamesPID ) is still encapsulated in an "if" sentence:
```
  // Do it only if a tool with *this* name hasn't already been used, and has been previously initialised
  //
  if ( !m_corrFileNamePID.empty() && !isToolAlreadyUsed(m_pidEffSF_tool_name) ) {
```
I suggest to move each of them in the corresponding IF sentence as they are not used anywhere else.

MihaMuskinja:
Closes #824 .

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -61,6 +61,9 @@ HelpTreeBase::~HelpTreeBase() {
     delete m_trigInfoSwitch;
 
     //muon
+    delete m_muonInfoSwitch;
+
+    //muon
     for (auto muon: m_muons)
       delete muon.second;
 
--- a/Root/ElectronEfficiencyCorrector.cxx
+++ b/Root/ElectronEfficiencyCorrector.cxx
@@ -654,9 +654,7 @@ EL::StatusCode ElectronEfficiencyCorrector :: executeSF ( const xAOD::ElectronCo
   //
   if ( !m_corrFileNamePID.empty() && !isToolAlreadyUsed(m_pidEffSF_tool_name) ) {
 
-    if(countSyst == 0) {
-      sysVariationNamesPID       = new std::vector< std::string >;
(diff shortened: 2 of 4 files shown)


johnda102:
go ahead

MihaMuskinja:
Sorry, the same update is needed for the MuonEffCorrector. I will add a new commit in a sec.

Root/HelpTreeBase.cxx
```diff
@@ -61,6 +61,9 @@ HelpTreeBase::~HelpTreeBase() {
     delete m_trigInfoSwitch;
 
     //muon
+    delete m_muonInfoSwitch;
```
kratsg: @johnda102 @MihaMuskinja -- @mattleblanc is seeing a segfault at this line, probably because of a deletion. Might need protection.

MihaMuskinja:
Hi @kratsg , the PR makes sense to me :). Thanks for notifying.

Issue #827, opened 2017-03-08
Name of the JVT tool needs to be set (m_JVT_tool_name)


MihaMuskinja:
Hi @kkrizka , @kratsg ,

I think that the m_JVT_tool_name string must be set as it is "" by default. This is done e.g. for electronEffTool in initialize:
```
m_TrigEffSF_tool_name = "ElectronEfficiencyCorrectionTool_effSF_Trig_" + m_WorkingPointTrigTrig + "_" + m_WorkingPointIDTrig;
RETURN_CHECK("ElectronEfficiencyCorrector::initialize()", checkToolStore<AsgElectronEfficiencyCorrectionTool>(m_TrigEffSF_tool_name), "" );

```
The problem is that the SFs are not calculated if the name is not set due to this "if" sentence:
```
if ( !isToolAlreadyUsed(m_JVT_tool_name) ) {
```

MihaMuskinja:
Adding the three lines below helps, but then the SF computation crashes for me..
```
  m_JVT_tool_name = "JetJvtEfficiency_effSF_" + m_name;
  RETURN_CHECK("ElectronEfficiencyCorrector::initialize()", checkToolStore<CP::IJetJvtEfficiency>(m_JVT_tool_name), "" );
  m_JVT_tool_handle.setTypeAndName("CP::JetJvtEfficiency/" + m_JVT_tool_name);
```

Error:
```
JetSelector::executeSe... INFO    --------------------------------------
JetSelector::executeSe... INFO    Applying JVT SF
JetSelector::executeSe... INFO    ===>>>
JetSelector::executeSe... INFO     
JetSelector::executeSe... INFO    Jet 2, pt = 194.42 GeV, |eta| = 1.80
JetSelector::executeSe... INFO     
JetSelector::executeSe... INFO    JVT SF decoration: JetJvtEfficiency_JVTSyst_JVT_Medium
JetSelector::executeSe... INFO     
JetSelector::executeSe... INFO    Systematic: 
JetSelector::executeSe... INFO     
JetSelector::executeSe... INFO    JVT SF:
JetSelector::executeSe... INFO    	 1.000000 (from getEfficiencyScaleFactor())
JetSelector::executeSe... INFO    --------------------------------------
JetSelector::executeSe... INFO    Applying JVT SF
--------------------
An exception was caught!
Traceback (most recent call last):
[...]

PR #828, opened 2017-03-08
added m_JVT_tool_name and Truth Jet info needed for JVT SFs

MihaMuskinja:
Closes https://github.com/UCATLAS/xAODAnaHelpers/issues/827,  but is the implementation really OK?

I added a loop over true jets in the execute method, is this fine? Name of the truth jet container is hard-coded "AntiKt4TruthJets".


kratsg:
go ahead

kratsg:
Can we not hard-code the name? Add it as configuration and set default to Akt4TruthJets

PR #833, opened 2017-03-09
fixed a crash in data due to truthJet container

MihaMuskinja:
fixes a bug that came with https://github.com/UCATLAS/xAODAnaHelpers/issues/827


kratsg:
go ahead

PR #839, opened 2017-03-14
Update/eff cont deep copy

### LINKED ISSUE #838: making subset containers
Hi @mmilesi, @kratsg, @jdandoy, 
when performing the OR, the function HelperFunctions::makeSubsetCont is used to create a container pointing at the objects passing the OR. The syst output containers created in this way may be just be a subset in memory of the nominal one. I find this to be a problem when decorating objects downstream e.g. with reco efficiencies. For instance, one of these containers could be muons_OR+ele_calibration_sys. Now, while these objects should be saved in a different tree than nominal, when decorating them with efficiencies, also the nominal ones will receive the decoration, which should obviously be written in the nominal tree. 

I believe the solution to this would be to change the memory location of the objects in the OR output container, hence, changing a bit the HelperFunctions::makeSubsetCont, but I am unsure if this would be the best thing to do. What do you guys think?

fscutti:
a deepcopy is performed for syst containers of muons and electrons to be decorated with efficiencies and their uncertainties, if a list of syst is passed in the configuration of the efficiency algo. This closes #838.

### DIFF
--- a/Root/ElectronEfficiencyCorrector.cxx
+++ b/Root/ElectronEfficiencyCorrector.cxx
@@ -9,6 +9,7 @@
 // EDM include(s):
 #include "xAODEventInfo/EventInfo.h"
 #include "xAODEgamma/ElectronContainer.h"
+#include "xAODEgamma/ElectronAuxContainer.h"
 #include "xAODEgamma/Electron.h"
 #include "xAODBase/IParticleHelpers.h"
 #include "xAODBase/IParticleContainer.h"
@@ -51,10 +52,12 @@ ElectronEfficiencyCorrector :: ElectronEfficiencyCorrector (std::string classNam
   // input container to be read from TEvent or TStore
   //
   m_inContainerName         = "";
+  m_outContainerName        = "";
 
   // Systematics stuff
   //
   m_inputAlgoSystNames      = "";
+  m_outputAlgoSystNames     = "";
   m_systValPID              = 0.0;
   m_systValIso              = 0.0;
   m_systValReco             = 0.0;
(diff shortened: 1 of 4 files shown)


fscutti:
Hi @kratsg,
the deepcopy is performed only if the container does not exist as, if the algo is executed multiple times, the deepcopy would fail as the container cannot be overwritten and it should not be. The same goes for the list of systematic names. If you're happy with the changes I'll merge.

kratsg:
But if the algorithm is performed multiple times won't it be running on different containers? Or maybe the algo-name should be included in the deep-copy name?

fscutti:
this decoration algorithm can run on the same container multiple times, for instance on selected muons. Each time would correspond to a different working point of the efficiency. The algo name should not enter in the deep-copy name as downstream we might just want to retrieve the name of the decorations.

kratsg:
> this decoration algorithm can run on the same container multiple times, for instance on selected muons. Each time would correspond to a different working point of the efficiency. The algo name should not enter in the deep-copy name as downstream we might just want to retrieve the name of the decorations but just one name for the decorated muons

Great point! This might want to be described quite briefly in the docs somewhere. Other than that, I'm fine with this being closed.

fscutti:
Alright. I'll have a look at the docs.

btcarlso:
This crashes for me (worked before the change). 
`ElectronEfficiencyCorr... ERROR   /afs/cern.ch/work/b/bcarlson/public/hinv/jan13/xAODAnaHelpers/Root/ElectronEfficiencyCorrector.cxx:563 Failed to execute: (HelperFunctions::makeDeepCopy<xAOD::ElectronContainer, xAOD::ElectronAuxContainer, xAOD::Electron>(m_store, m_outContainerName+systName, inputElectrons))`

Does this change also require a change to the configuration?

fscutti:
Hi @btcarlso, 
 [here](https://github.com/SameSignDiLepton/SSDiLepAnalysis/blob/master/scripts/config_v3.py) is the configuration I am using. On correctors you now have to create an output container

PR #841, opened 2017-03-14
Update Electron, Muon and Jet Hists for bJetTrig Studies

laurie97:
- electronHists: update to have more isolation plots and quality plots
- muonHists: Update to have quality helper function
- jetHists: Update to plot jetPt in eta bins, online bs from tool, adjusted plot logic for HLT vertex switches (speeds up some cases).
    - Also corrected a load of strange indentation from L1860 onwards 
          - (In pre-change version L1788 is same level of logic as L1688 but different indentation level)


johnda102:
go ahead

PR #844, opened 2017-03-16
Feature/forward jvt

btcarlso:
Add forward JVT

### DIFF
--- a/Root/JetSelector.cxx
+++ b/Root/JetSelector.cxx
@@ -559,7 +559,7 @@ bool JetSelector :: executeSelection ( const xAOD::JetContainer* inJets,
   //
   SG::AuxElement::Decorator< char > passSelDecor( m_decor );
 
-  for ( auto jet_itr : *inJets ) { // duplicated of basic loop
+  for ( auto jet_itr : *jets_copy ) { // duplicated of basic loop
 
     // if only looking at a subset of jets make sure all are decorated
     if ( m_nToProcess > 0 && nObj >= m_nToProcess ) {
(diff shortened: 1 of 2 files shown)


kratsg:
go ahead

kratsg:
add to whitelist too @xAH-bot please.

johnda102:
HI @btcarlso  @kratsg 

This commit breaks code where you want to run Jet Selector on an input collection multiple times. 

The first instance writes the inputContainer+"Copy" to TStore, then the next instance tries to do the same and crashes. 

Can you put in a fix to check if the inputContainer+"Copy" already exists ?

johnda102:
Also,  are we really making a deep copy of all the input jets, all the time ?

This seems like overkill.  From what I can tell this is only need if one is using forward JVT. 
As this seem to be more of a niche market can we only do a deep copy if fJVT selection is requested ? This will give more performant code for the majority of clients who dont use forward JVT.

kratsg:
> This commit breaks code where you want to run Jet Selector on an input collection multiple times.
> 
> The first instance writes the inputContainer+"Copy" to TStore, then the next instance tries to do the same and crashes.
> 
> Can you put in a fix to check if the inputContainer+"Copy" already exists ?

@johnda102 I've hopefully addressed these issues in #852.

> Also, are we really making a deep copy of all the input jets, all the time ?
> 
> This seems like overkill. From what I can tell this is only need if one is using forward JVT.
> As this seem to be more of a niche market can we only do a deep copy if fJVT selection is requested ? This will give more performant code for the majority of clients who dont use forward JVT.

I've also switched to using a shallow copy as the modify code appears to only add/set decorations (which technically can be done without needing a shallow copy or any kind of copy at all). If this still breaks, fJVT developers need to know as this is a problem with their code.

PR #846, opened 2017-03-17
adds a configurable member for AFII on/off for the electron eff corr

MihaMuskinja:
This adds a configurable m_setAFII member. The same is already implemented in the electron calibrator: 

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronCalibrator.cxx#L202


kratsg:
add to whitelist

kratsg:
go ahead

PR #867, opened 2017-04-14
Set up CppCheck and uncrustify testing

kratsg:
This PR starts adding CppCheck (install only) and uncrustify testing using travis.


kratsg:
Ok, so uncrustify is set-up correctly now.

PR #883, opened 2017-05-08
[skip ci] update xAH_run verbosity

kratsg:
This just makes the xAH_run.py script a little less verbose. There is a new option called `log-level` which can be set using `xAH_run.py -l DEBUG` or `xAH_run.py --log-level DEBUG` for example. The default is `INFO` level.


johnda102:
looks good to me

johnda102:
go ahead

PR #895, opened 2017-05-12
Move configuration to header

### LINKED ISSUE #886: Migrate streamable variable defaults back to the header

kratsg:
This closes #886.


xAODAnaHelpers/JetCalibrator.h
```diff
@@ -81,43 +81,41 @@ class JetCalibrator : public xAH::Algorithm
...
 
   @endrst */
-  bool m_setAFII;
-  bool m_forceInsitu;
-
-  /// @brief whether the jet collection is trigger or not (soon: different calibrations)
-  bool m_isTrigger;
+  bool m_setAFII = false;
+  /// @brief when running data "_Insitu" is appended to this string
+  bool m_forceInsitu = false;
```
jbossios: @kratsg  I just realized about that the default was changed here, from true to false. That means that "_Insitu" is not longer forced. Just wondering if it was intended.

kratsg: That might be a typo. Feel free to add a MR to fix this.

jbossios: Ok, I will add that change in a MR I'm about to create.

PR #902, opened 2017-05-13
ExpandPathName is no more

### LINKED ISSUE #776: Switch to new pathresolver
```
#include "PathResolver/PathResolver.h"
...
const std::string fileName = PathResolverFindCalibFile( "MyLoop/file.root" );
```

kratsg:
Closes #776.

### DIFF
--- a/Root/OnlineBeamSpotTool.cxx
+++ b/Root/OnlineBeamSpotTool.cxx
@@ -8,7 +8,7 @@
 #include "TTree.h"
 
 using namespace xAH;
-using std::cout;  using std::endl; 
+using std::cout;  using std::endl;
 using std::make_pair;
 
 OnlineBeamSpotTool::OnlineBeamSpotTool() :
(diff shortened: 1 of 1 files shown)


jdandoy:
@kratsg PathResolver does not seem to work in RootCore (2.4.31), as DATAPATH is not being set.  Maybe we need a workaround to use gSystem->ExpandPathName() when in 2.4.X?

kratsg:
Use CalibFile, not DataFile

Giordon

kratsg:
Also, the expandpathname isn't a workaround as the paths you pass in differ between PathResolver and ExpandPathName. Can you file a JIRA bug report or ask on PAThelp?

jdandoy:
Okay CalibFile seems to work in both 2.4 and 2.6 (it's actually for the GRL xml path in BasicEventSelection), is there any reason to use DataFile instead of CalibFile there?  In fact, what's the difference between the two, does DataFile only correspond to atlas_install_data? 

Given CalibFile works fine with different paths in 2.4, I don't think there's a problem.

Issue #946, opened 2017-06-22
Broken JetCalibrator for 2.4


ntadej:
It seems that https://github.com/UCATLAS/xAODAnaHelpers/pull/928/commits/33572d9a2f02da29cba829436b78199a937b62e5 broke the JetCalibrator for 2.4. Commenting out the changes fixes running it. Should this be 2.6/21.0 specific?

(Also this doesn't seem to be photon specific, but was merged in the same PR. Maybe not a good idea 😊)

kratsg:
@jdandoy can you please look at this? Thanks!

jdandoy:
Hi @ntadej,

Can you say specifically how things are broken for you?  I'm guessing it's because the recommended AB version is out of date, as this necessary feature for release 21 was only introduced for 2.6.3 and 2.4.32 (we're listed as 2.6.2 and 2.4.31).  

See #928, the name of the JVT decoration is different in Release 21 and Release 20.7, and the official tool (and therefore our tool) was updated to allow this name to be set in the config file (as detailed in the comment above the variable [here](https://github.com/UCATLAS/xAODAnaHelpers/blob/33572d9a2f02da29cba829436b78199a937b62e5/xAODAnaHelpers/JetCalibrator.h#L106).  As we don't have a way for the algorithms to know generally if you're running on 20.7 or 21, you will need to set this in your config file if you're using 20.7.  The default value is set for release 21.

ntadej:
The issue is that `JVFCorrName` property doesn't even exist in 2.4.31 or 2.4.32. I set the 20.7 version of the decoration and it didn't help.

jdandoy:
Okay, it seems that [JetMomentTools-00-03-62](https://svnweb.cern.ch/trac/atlasoff/browser/Reconstruction/Jet/JetMomentTools/tags/JetMomentTools-00-03-62/Root/JetVertexTaggerTool.cxx) is not actually available yet in any 2.4.X, and there doesn't seem to be a request for it.  I'll make a request, and update a patch for the checkout tags code.  In the meantime, please download JetMomentTools-00-03-62 manually and give it a try.

Jeff

kratsg:
@jdandoy do you need a flag for rel21/rel20.7? Because as far as I'm aware, there is a way to check using the metadata in the file? Would this help?

jdandoy:
Hi @kratsg,

[...]

PR #948, opened 2017-06-28
Protect JVFCorrName for r20.7

jdandoy:
Solves #946. The JVT Update tool accesses aux data with a different name between r20.7 and r21.  Unfortunately many mc16a samples have the old name, so a new property "JVFCorrName" was added to the tool to specify the name.  This property is only available in the r21 JetMomentTools tag, and after discussions with the experts it would be inconvenient to update the r20.7 tag to include it, especially given that no r20.7 samples have the r21 name.   I've therefore edited the logic so JVFCorrName will only be set if m_JvtAuxName is not empty, keeping backwards compatibility.


kratsg:
@jdandoy Can you explain the name of the variable `m_JvtAuxName`? Is "Aux" for "Auxiliary" or something else? Otherwise, I like this fix.

jdandoy:
Hi @kratsg,
You're correct, the string reflects the name of the auxdata variable.  I'm open to suggestions for a different name.

Jenkin's failed, but the console output suggests an issue with Jenkins?  Can you check this?

kratsg:
Connect is down for upgrades

kratsg:
go ahead

PR #955, opened 2017-07-12
Fix overwritten b-tagging SF for multiple trees

ntadej:
Make a deep copy of jet containers in `BJetEfficiencyCorrector` when we have multiple systematics trees so SFs do not get overwritten. Inspired by what is currently done for electrons and muons.

Also, copy over jets from nominal tree if not present in systematics (e.g. MET systematics). This is currently also done for muons.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -89,6 +89,17 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
   m_isMC = ( eventInfo->eventType( xAOD::EventInfo::IS_SIMULATION ) );
 
   ANA_MSG_INFO( "Number of events in file: " << m_event->getEntries() );
+  
+  // several lists of systematics could be configured
+  // this is the case when MET sys should be added 
+  // to the OR ones
+
+  // parse and split by comma
+  std::string token;
+  std::istringstream ss(m_inputAlgo);
+  while (std::getline(ss, token, ',')) {
+    m_inputAlgoList.push_back(token);
+  }
 
   m_decorSF = m_decor + "_SF";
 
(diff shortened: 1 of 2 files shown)


Root/BJetEfficiencyCorrector.cxx
```diff
@@ -86,6 +89,23 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
...
+    size_t pos = tmp_sysNames.find_first_of(',');
+    if ( pos == std::string::npos ) {
+      pos = tmp_sysNames.size();
+      m_inputAlgoList.push_back(tmp_sysNames.substr(0, pos));
+      tmp_sysNames.erase(0, pos);
+    } else {
+      m_inputAlgoList.push_back(tmp_sysNames.substr(0, pos));
+      tmp_sysNames.erase(0, pos+1);
+    }
+  }
```
kratsg: Use `istringstream` instead for performance reasons

```c++
// parse and split by comma
std::string token;
std::istringstream ss("");
 
// A B C D ... Z -> {A, B, C, D, ..., Z}
ss.clear(); ss.str(m_simpleCopyKeys);
while(std::getline(ss, token, ' '))
  m_simpleCopyKeys_vec.push_back(token);
```

ntadej: Thanks @kratsg, this also simplifies the code a lot. I updated the code.

To be honest I was lazy and copied this from muons. Will also update this part of code there.

kkrizka:
Why was this required? What exactly is being overwritten? My understanding is that all systematic variations of jets used as inputs are already individual shallow copies. All that `BJetEfficiencyCorrector` does is set a decoration (tagging decision + scale factor) for each shallow copy jet.

Also why do we need deep copies instead of shallow copies of all jets? Isn't the whole point of a shallow copy to be able to modify the attributes of an object without having to make a copy of all the other "stable" attribuets?

PR #960, opened 2017-07-19
Check for existence of input container in BJetEfficiencyCorrector

ntadej:
It seems that input containers are sometimes missing after using overlap removal. I'm not sure why this happens, but this kind of check for existence of input containers already exists for electrons and muons.

A reference job that crashed without this fix:
http://bigpanda.cern.ch/job/3517326552/

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -284,12 +284,13 @@ EL::StatusCode BJetEfficiencyCorrector :: execute ()
   // then get the one collection and be done with it
   //
   if ( m_inputAlgo.empty() ) {
+    // Check the existence of input container
+    if ( m_store->contains<xAOD::JetContainer>( m_inContainerName ) ) {
+      // this will be the collection processed - no matter what!!
+      ANA_CHECK( HelperFunctions::retrieve(inJets, m_inContainerName, m_event, m_store, msg()) );
 
-    // this will be the collection processed - no matter what!!
-    ANA_CHECK( HelperFunctions::retrieve(inJets, m_inContainerName, m_event, m_store, msg()) );
-
-    executeEfficiencyCorrection( inJets, eventInfo, true);
-
+      executeEfficiencyCorrection( inJets, eventInfo, true);
+    }
   }
   //
   // get the list of systematics to run over
@@ -349,17 +350,20 @@ EL::StatusCode BJetEfficiencyCorrector :: execute ()
 
       bool doNominal = (systName == "");
 
-      ANA_CHECK( HelperFunctions::retrieve(inJets, m_inContainerName+systName, m_event, m_store, msg()) );
+      // Check the existence of the container
+      if ( m_store->contains<xAOD::JetContainer>( m_inContainerName+systName )  ) {
+        ANA_CHECK( HelperFunctions::retrieve(inJets, m_inContainerName+systName, m_event, m_store, msg()) );
 
-      if ( !m_store->contains<xAOD::JetContainer>( m_outContainerName+systName ) ) {
-        ANA_CHECK( (HelperFunctions::makeDeepCopy<xAOD::JetContainer, xAOD::JetAuxContainer, xAOD::Jet>(m_store, m_outContainerName+systName, inJets)));
+        if ( !m_store->contains<xAOD::JetContainer>( m_outContainerName+systName ) ) {
+          ANA_CHECK( (HelperFunctions::makeDeepCopy<xAOD::JetContainer, xAOD::JetAuxContainer, xAOD::Jet>(m_store, m_outContainerName+systName, inJets)));
+        }
+
+        ANA_CHECK( HelperFunctions::retrieve(outJets, m_outContainerName+systName, m_event, m_store, msg()) );
+
+        executeEfficiencyCorrection( outJets, eventInfo, doNominal );
+
+        vecOutContainerNames->push_back( systName );
       }
-
-      ANA_CHECK( HelperFunctions::retrieve(outJets, m_outContainerName+systName, m_event, m_store, msg()) );
-
-      executeEfficiencyCorrection( outJets, eventInfo, doNominal );
-
-      vecOutContainerNames->push_back( systName );
     }
 
     if ( !m_outputAlgo.empty() && !m_store->contains< std::vector<std::string> >( m_outputAlgo ) ) { // might have already been stored by another execution of this algo


kratsg:
> It seems that input containers are sometimes missing after using overlap removal. I'm not sure why this happens, but this kind of check for existence of input containers already exists for electrons and muons.

I don't understand what you mean by missing. This is in BJetEffCorrector *only*? Are you sure it's just not because the systematic container wasn't created because the systematic didn't pass some selection?

The point here is that if something is missing, you should find out why it's missing -- not protect against it just to fix it. it might introduce some physics bugs that are easily undetectable.

ntadej:
> I don't understand what you mean by missing. This is in BJetEffCorrector only? Are you sure it's just not because the systematic container wasn't created because the systematic didn't pass some selection?

I agree with you, that this is not a good way to fix it, but the same thing is also done for muons and electrons. Actually in the linked job, a nominal tree is missing. Can you give me any directions how to check this?

kratsg:
You can use the ```TStore::print()``` functionality and debug this locally if it's reproducible for sure. And then start figuring out if something is clearing the TStore after OR... but I really doubt this can happen.

MihaMuskinja:
Hi @kratsg , @ntadej ,

I am debugging this event right now. The "AntiKt4EMTopoJets_OR" container exists at the end of the execute() in the OR tool, but not any more at the beginning of the execute in BJetEff, which is the next algo.. How can this happen?

Best,
Miha

ntadej:
One more thing to add here is that checks in `MuonEfficiencyCorrector` have been added when systematics were being implemented. @fscutti, do you remember if you had similar errors with missing containers after OR?

MihaMuskinja:
Hi @ntadej , @kratsg , @fscutti ,

just to add another observation: it does not crash with nominal only. The container is there for nominal only but not for systematics. This is an extremely suspicious behaviour.

kratsg:
I'll look into this too after my practice talk for DPF.

kratsg:
> I am debugging this event right now. The "AntiKt4EMTopoJets_OR" container exists at the end of the execute() in the OR tool, but not any more at the beginning of the execute in BJetEff, which is the next algo.. How can this happen?

How do you know it exists at the end of execute() in the OR tool? is there an example configuration I can run on my mc15(rel20)/mc16(rel21) file to check this out?

> just to add another observation: it does not crash with nominal only. The container is there for nominal only but not for systematics. This is an extremely suspicious behaviour.

This is probably indicative of the problem.

kratsg:
Looking at this OR code -- it can be significantly refactored to update the `executeOR()` function. I can do this in another branch?

ntadej:
@kratsg, I agree there is a lot of duplicated code in OR. If you think OR might one of the possible causes for issues described here, feel free to try clean it up in a branch and I can cherry-pick it to test it locally.

Our code is available here: https://gitlab.cern.ch/ATLAS-TypeIIISeeSaw/AnalysisCode
We are using 2.4.33 at the moment with rel20 samples. You can run `localrun sys <filename>` after building it and it should pick up the correct configuration.

MihaMuskinja:
@kratsg @ntadej if you are interested in the exact event where it happens:

rucio download mc15_13TeV:DAOD_EXOT12.10711285._000017.pool.root.1

localrun sys ../mc15_13TeV/DAOD_EXOT12.10711285._000017.pool.root.1  --skip 17165 -n 1 2>&1 | tee log.xAH

kratsg:
@MihaMuskinja @ntadej I refactored the code (hopefully) in https://github.com/UCATLAS/xAODAnaHelpers/tree/refactor/OverlapRemover

```git fetch && git checkout refactor/OverlapRemover```

ntadej:
@kratsg, I've found the issue!

`*Selector` algorithms have event selection, but some systematics can actually pass the event selection so event in general should not be rejected. `*_Selected` containers (our nomenclature) are thus not created and then missing later. My fix is then (short term) also the right thing to do, although TreeAlgo needs some changes that we already have in our code. 

I guess a (shallow) copy for `xAOD::EventInfo` is needed for each systematic to be decorated if event is rejected for it. Then we ignore this event for each systematic in the following algorithms. This would need a substantial refactor though. Will add this to my proposal ticket. 😞

kratsg:
> *Selector algorithms have event selection, but some systematics can actually pass the event selection so event in general should not be rejected. *_Selected containers (our nomenclature) are thus not created and then missing later. My fix is then (short term) also the right thing to do, although TreeAlgo needs some changes that we already have in our code.

I don't understand how this is causing the problem. For example, let's say do use a JES UP systematic -- the nominal might not pass an event-level jet selection (requiring 4 jets with pT > 20 GeV) because the 4th jet is like 18 GeV... while the JES UP systematic would bump up that 4th jet to > 20 GeV so the JESUP systematic passes. Are you saying that you're missing the nominal container downstream or?

Because technically, you're supposed to only use systematics that pass the event selections. Which is one of the other reasons we don't just store the systematics globally -- since the list of what we look at can change on event-by-event basis?

> I guess a (shallow) copy for xAOD::EventInfo is needed for each systematic to be decorated if event is rejected for it. Then we ignore this event for each systematic in the following algorithms. This would need a substantial refactor though. Will add this to my proposal ticket. 😞

Why would this be fixed? Are you talking about #964?

ntadej:
> I don't understand how this is causing the problem. For example, let's say do use a JES UP systematic -- the nominal might not pass an event-level jet selection (requiring 4 jets with pT > 20 GeV) because the 4th jet is like 18 GeV... while the JES UP systematic would bump up that 4th jet to > 20 GeV so the JESUP systematic passes. Are you saying that you're missing the nominal container downstream or?

Yes nominal container is then missing if nominal does not pass event selection. Or other way around, the JES UP systematic container will be missing downstream.

kratsg:
I believe we decided to do this method moving forward. What is supposed to be the suggested way of handling this case when nominal passes but systematic doesn't (inefficiency?) and when nominal doesn't pass and systematic does. Is your suggestion simply to decorate the eventInfo object with whether the systematic passed?

ntadej:
Yes, my suggestion is simply to decorate the EventInfo object. This way we would know that the container is missing because of that, not some upstream issue.

kratsg:
> Yes, my suggestion is simply to decorate the EventInfo object. This way we would know that the container is missing because of that, not some upstream issue.

To be honest, that's the only reason the container would ever be missing in our code. You should assume upstream code is reliable enough that if it's missing a container because it can't record it, that it's already did something like `return EL::StatusCode::FAILURE` or called `wk()->skipEvent()` entirely. If you go under that assumption, then this is easier because then you do not need to decorate eventInfo?

kratsg:
The other thing could just be modifying the list of systematics (including nominal) that get passed through to downstream and only putting in the ones that pass. Since the TreeAlgo/etc.. only sets things for the systematics it finds and resets all other values, this could be somewhat reliable?

ntadej:
As we now know why the containers are missing, I think this can be merged in as it is a valid fix (I can rebase it to remove conflicts).

Regarding systematics lists: calibration systematics could be updated as in your proposal. I guess they don't need to be written out at the end (as they are obvious from the tree names). Scale factor systematics should be the same for all events, is that correct?

kratsg:
> As we now know why the containers are missing, I think this can be merged in as it is a valid fix (I can rebase it to remove conflicts).

Can you double-check? You might need a `git push -f`.

> Regarding systematics lists: calibration systematics could be updated as in your proposal. I guess they don't need to be written out at the end (as they are obvious from the tree names). Scale factor systematics should be the same for all events, is that correct?

I suspect they need to be the same. So perhaps just protecting against missing containers is all we need to do... and we need to place this sort of thinking somewhere in the docs that the only reason a container is missing is that it did not pass a selection upstream?

PR #961, opened 2017-07-19
JVT improvements

ntadej:
Several improvements for JVT SF calculation:
 - JVT cut can be disabled in `JetSelector` but SFs are still calculated, a decorator is added for passing
 - added inefficiency SFs for above case
 - SFs are only stored for nominal tree
 - only JVT SFs for a specific working point can now be stored

### DIFF
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -277,6 +277,20 @@ namespace HelperClasses{
     m_vsLumiBlock         = has_exact("vsLumiBlock");
     m_lumiB_runN          = has_exact("lumiB_runN");
 
+    if( has_match( "sfJVT" ) ) {
+      std::string input(m_configStr);
+      // erase everything before the interesting string
+      input.erase( 0, input.find("sfJVT") );
+      // erase everything after the interesting string
+      // only if there is something after the string
+      if( input.find(" ") != std::string::npos ) {
+        input.erase( input.find_first_of(" "), input.size() );
+      }
+      // remove sfJVT to just leave the working point
+      input.erase(0,5);
+      m_sfJVTName = input;
+    } // sfJVTName
+
     m_sfFTagFix.clear();
     if( has_match( "sfFTagFix" ) ) {
       std::string input(m_configStr);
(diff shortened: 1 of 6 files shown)


Root/HelperClasses.cxx
```diff
@@ -277,6 +277,20 @@ namespace HelperClasses{
...
+      input.erase( 0, input.find("sfJVT") );
+      // erase everything after the interesting string
+      // only if there is something after the string
+      if( input.find(" ") != std::string::npos ) {
+        input.erase( input.find_first_of(" "), input.size() );
+      }
+      // remove sfJVT to just leave the working point
+      input.erase(0,5);
+      m_sfJVTName = input;
+    } // sfJVTName
```
kratsg: This logic should be incorporated into the InfoSwitch itself... Look at how we do it for `m_numLeading` and be inspired!

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelperClasses.cxx#L111-L125

The trick is that the entire string you look at when trying to match something is like `NLeadingXXXX` only so you know the rest of the string can just be parsed as a number. A similar procedure can be done here to make this easier for anyone to use.

Root/JetContainer.cxx
```diff
@@ -2154,6 +2179,55 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
     } // trackPV
 
   }
+  
+  if ( m_mc ) {
```
kratsg: move all the static accessors to the top inside the `if(m_mc)`

Root/HelperClasses.cxx
ntadej: I added two new functions `get_working_point` and `get_working_points`, currently only for `std::string` based WPs. I can make `int` based functions also and update all classes for unification in a separate MR.

Root/JetContainer.cxx
ntadej: Done.

ntadej:
It seems that last jenkins build failed with an error not related to this PR.

kratsg:
test this please

PR #970, opened 2017-07-24
Do not output isolation kinematics by default

ntadej:
It would be nice to remove isolation kinematics from the ntuples to keep sizes low. This adds another flag `isolationKinematics` in addition to existing `isolation` for electrons and muons. Unfortunately this is a breaking change but keeping backwards compatibility would require two new flags (we can go either way).


kratsg:
Nah. I'm fine with this.

PR #971, opened 2017-07-24
Create an xAODAnaHelpers python module for importing

kratsg:
This will make life somewhat easier as we'll be able to clean up our python code a lot moving forward.


CMakeLists.txt
```diff
@@ -104,5 +104,6 @@ atlas_add_library( xAODAnaHelpersLib xAODAnaHelpers/*.h Root/*.h Root/*.cxx ${xA
 )
 
 # Install files from the package:
+atlas_install_python_modules( python/*.py )
```
kratsg: this is the change that allows us to do `import xAODAnaHelpers`

Issue #964, opened 2017-07-20
Proposal: Simplifying systematics calculations


ntadej:
As we are currently observing some strange issues with disappearing containers, I dived a bit into the how xAH handles containers for systematics and how are scale factors applied at the end of the calculations.

If my knowledge is correct only `*Calibrator` algorithms actually change quantities on physics objects. All others add decorators based on these quantities, which are the same for systematics not affecting the objects (e.g. electron SFs will be the same for all muon systematics). Calibrators already make shallow copies of the initial objects so these containers should be safe for different SFs.

There is a lot of deep copying going around in the `*EfficiencyCorrector` algorithms which is probably not necessary. Also we should expect all input containers to be available and not just ignore if they are missing.

My proposal:
 - writing systematics names into metadata (or another appropriate location), not per event
 - decoupling of nominal and systematic scale factors (to be decided if vector also contains nominal SF)
 - remove all deep copying of containers, using `ConstDataVector`s is enough to sort and filter the objects
 - remove quiet checks if input container is missing - they should all be there or something is wrong
 - Scale factors changes:
   - Nominal container will have all SFs calculated (nominal and systematics)
   - Systematics containers that actually affect these objects e.g. electron container with electron calibration systematics will have nominal SFs calculated.
   - All other containers should use the SFs from nominal tree as we are using shallow copies initially in the calibrator.
  - write out systematics SFs only for nominal tree and keep branches in other trees empty (or maybe even not include them, to avoid confusion)

I would really like the input from core developers, but this would simplify things a lot. I can also volunteer to make the initial implementation.

kratsg:
> writing systematics names into metadata (or another appropriate location), not per event

This is potentially possible, but you need to write it into a location that non-xAH algorithms can access. Nominally, this is the point of the `TStore`. We don't have another location that is shared amongst all algorithms. We can do it for xAH::Algorithms internally only, but this breaks the flexibility of xAH with non-xAH algorithms.

> decoupling of nominal and systematic scale factors (to be decided if vector also contains nominal SF)

seems like a good idea (do we not do this already?)



> remove all deep copying of containers, using ConstDataVectors is enough to sort and filter the objects

Yes. We'll probably need this. Deep-copying is needed in very specific cases and I suspect CDV wasn't available at the time when we were doing deep-copying back in 2015.



> remove quiet checks if input container is missing - they should all be there or something is wrong

Yep. This should probably be done first.

ntadej:
[...]

PR #972, opened 2017-07-25
Electron Efficiency Corrector simplification

ntadej:
This acts as a prototype of what I proposed in #964. I'm open for comments and criticism. This PR can later be either improved or closed.

Outline of the changes:
 - el. efficiency corrector modifies existing containers - no deep copying anymore, output flag removed
 - systematics list is computed in advanced - removal of duplicated code for no syst vs. syst case
 - only accepts upstream electron systematics - input flag renamed to be explicit
 - systematics SFs are only written for nominal tree (new `safeSFVectorFill` function)
 - <s>no sys names output specifications - to be able to successfully write out in the tree</s>
 - <s>nominal and systematics SFs in separate branches (the latter currently still contains nominal SFs - to be discussed)</s>

### DIFF
--- a/Root/ElectronEfficiencyCorrector.cxx
+++ b/Root/ElectronEfficiencyCorrector.cxx
@@ -119,17 +119,6 @@ EL::StatusCode ElectronEfficiencyCorrector :: initialize ()
 
   // *******************************************************
 
-  // several lists of systematics could be configured
-  // this is the case when MET sys should be added
-  // to the OR ones
-
-  // parse and split by comma
-  std::string token;
-  std::istringstream ss(m_inputAlgoSystNames);
-  while (std::getline(ss, token, ',')) {
-    m_sysNames.push_back(token);
-  }
-
   int sim_flav(1); // default for FullSim
   if ( m_isMC ) {
     const std::string stringMeta = wk()->metaData()->castString("SimulationFlavour");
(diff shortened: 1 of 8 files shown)


Root/FatJetContainer.cxx
```diff
@@ -7,7 +7,7 @@ using namespace xAH;
 
 FatJetContainer::FatJetContainer(const std::string& name, const std::string& detailStr, const std::string& suffix,
 				 float units, bool mc)
-  : ParticleContainer(name,detailStr,units,mc, false, suffix),
+  : ParticleContainer(name,detailStr,units,mc, true, false, suffix),
```
kratsg: Why is this change in this PR?

Root/JetContainer.cxx
```diff
@@ -6,7 +6,7 @@
 using namespace xAH;
 
 JetContainer::JetContainer(const std::string& name, const std::string& detailStr, float units, bool mc)
-  : ParticleContainer(name,detailStr,units,mc),
+  : ParticleContainer(name,detailStr,units,mc, true),
```
kratsg: Why is this change in this PR?

Root/MuonContainer.cxx
```diff
@@ -7,7 +7,7 @@ using std::vector;
 using std::string;
 
 MuonContainer::MuonContainer(const std::string& name, const std::string& detailStr, float units, bool mc)
-  : ParticleContainer(name, detailStr, units, mc, true)
+  : ParticleContainer(name, detailStr, units, mc, true, true)
```
kratsg: Why is this change in this PR?

Root/PhotonContainer.cxx
```diff
@@ -7,7 +7,7 @@ using std::vector;
 using std::string;
 
 PhotonContainer::PhotonContainer(const std::string& name, const std::string& detailStr, float units, bool mc)
-  : ParticleContainer(name, detailStr, units, mc, true)
+  : ParticleContainer(name, detailStr, units, mc, true, true)
```
kratsg: Why is this change in this PR?

Root/TauContainer.cxx
```diff
@@ -7,7 +7,7 @@ using std::vector;
 using std::string;
 
 TauContainer::TauContainer(const std::string& name, const std::string& detailStr, float units, bool mc)
-  : ParticleContainer(name, detailStr, units, mc, true)
+  : ParticleContainer(name, detailStr, units, mc, true, true)
```
kratsg: Why is this change in this PR?

Root/ElectronEfficiencyCorrector.cxx
```diff
@@ -738,21 +667,19 @@ EL::StatusCode ElectronEfficiencyCorrector :: executeSF ( const xAOD::ElectronCo
   //
   if ( !m_corrFileNameIso.empty() && !isToolAlreadyUsed(m_IsoEffSF_tool_name) ) {
 
-    if(countSyst == 0) sysVariationNamesIso = new std::vector< std::string >;
+    if ( writeSystNames ) sysVariationNamesIso = new std::vector< std::string >;
+    
+    // Create the names of the SF weights to be recorded
+    std::string sfName = "ElPIDEff_SF_" + m_IsoPID_WP + "_isol" + m_Iso_WP;
+    std::string sfNameSyst = "ElPIDEff_SF_syst_" + m_IsoPID_WP + "_isol" + m_Iso_WP;
+    std::string sfSystNames = sfName + "_systNames";
```
kratsg: This name is used to write the list of systematics to an output for downstream algorithms to use. How are users supposed to know this name if they don't set it?

xAODAnaHelpers/ParticleContainer.h
```diff
@@ -23,11 +23,13 @@ namespace xAH {
 		      const std::string& detailStr="",
 		      float units = 1e3,
 		      bool mc = false,
+          bool nominal = true,
```
kratsg: I'm not entirely sure ParticleContainers need to know about systematics/nominal or not. There should be no difference between a nominal particle container and a systematic particle container - at least the way I've understood the ParticleContainer so far (@johnda102 can correct me, comment here).

Is it the assumption that ScaleFactors only exist on the nominal objects? If so, why not call it "storeSFs" instead of nominal/not-nominal. I don't know if there's a situation where we need to store SFs on systematic objects.

Root/FatJetContainer.cxx
ntadej: This is due to change in the `ParticleContainer` to ensure the code compiles (same for other containers).

Root/ElectronEfficiencyCorrector.cxx
ntadej: Actually I don't think scale factor systematics need to be reported downstream. I can restore configurability of this but then it also needs to be provided to the `TreeAlgo`/`HelpTreeBase` to successfully write those in the tree/metadata (to be done).

xAODAnaHelpers/ParticleContainer.h
ntadej: Yes, we can rename this if needed.

The only reference on ParticleContainers I've found is in the `HelpTreeBase`, so I guess a tree could be marked as nominal (or to store SFs). We can also decide if we want to add flags to calculate all SFs for systematics tree.

ntadej:
I general this tool is only used to calculate SFs and decorate electrons. If we provide the algorithm only with systematics of electrons, then all shallow copies should already be in place for electron containers and no additional copying is needed.

MihaMuskinja:
Hi @kratsg , @ntadej , just to be sure - will this create actually different branches in the output TTree for nominal scale factors and sys scale factors? This would break all the analyses using them as so far we have one std::vector<float> for each electron, where the 0th component is the nominal SF and the others are sys variations.

ntadej:
I consider this PR now done.

I made some minor changes:
 - for now I kept SFs in the same branch, keeping backwards compatibility and simplicity
 - `ParticleContainer` now has `storeSystSFs` flag which is true by default to avoid breaking something
 - I kept output syst names configurable

alexandertuna:
Hi @kratsg, @ntadej,

@tongbaojia and I are trying to write an ntuple with electron SFs now. Can we merge this PR soon, or kill it? I don't want to understand all the code in Electron*.cxx and then have it completely overhauled.

For the record: I think "nominal and systematics SFs in separate branches" is obviously the correct approach. But we're entrenched in the "one vector of SFs per lepton" approach, so be it.

alexandertuna:
@ntadej @kratsg -- do you have an example config of this? My previous config seg faults out of the box.

PR #990, opened 2017-08-16
Copy data containers when applying electron and muon SF

ntadej:
Copy data containers when applying electron and muon SF. This way, we can use the same configuration for data and MC. If we copy containers for MC we should also do it for data.


kratsg:
Why do you need a deep copy here? Wouldn't a shallow copy work?

ntadej:
Oh, I haven't thought of that ... it's actually an old commit I wanted to submit upstream. I will investigate and update this PR.

ntadej:
I now just make a new `ConstDataVector` and copy pointers to objects there. I think this is the cleanest option.

ntadej:
I think this is not needed anymore.

PR #991, opened 2017-08-16
Write out systematics names to metadata

ntadej:
I was looking for the best way to store systematics names and probably this is the most optimal. It stores syst. names in bin names of a histogram and it's:
 - easy to store (it's as complicated to store a list of strings to root file and to read it then as using a histogram for it)
 - hadd friendly (no effect when merging)
 - stored once per job so also disk space friendly

This stores all systematics. For scale factors, they should all exist always anyway (at least we rely on it as we use vectors to store them). For calibrations, we can read tree names anyway but it's useful to have them for filtering (i.e. doing only jet systematics).

This feature is disabled by default but this can be changed.

### DIFF
--- a/Root/HelperFunctions.cxx
+++ b/Root/HelperFunctions.cxx
@@ -468,6 +468,29 @@ std::vector< CP::SystematicSet > HelperFunctions::getListofSystematics(const CP:
 
 }
 
+void HelperFunctions::writeSystematicsListHist( const std::vector< CP::SystematicSet > &systs, std::string histName, TFile *file )
+{
+  std::string folderName = "systematics";
+  std::string name = folderName + "/" + histName;
+  if (!systs.size() || histName.empty() || file->Get(name.c_str())) {
+    return;
+  }
+
+  if (!file->Get(folderName.c_str())) {
+    file->mkdir(folderName.c_str());  
+  }
+  file->cd(folderName.c_str());
+
+  TH1D hist(histName.c_str(), histName.c_str(), systs.size(), 0.5, systs.size() + 0.5);
+  for (size_t i = 0; i < systs.size(); i++) {
+    if (systs[i].name().empty()) {
+      hist.GetXaxis()->SetBinLabel(i + 1, "nominal");
+    } else {
+      hist.GetXaxis()->SetBinLabel(i + 1, systs[i].name().c_str());
+    }
+  }
+  hist.Write();
+}
 
 float HelperFunctions::dPhi(float phi1, float phi2)
 {
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -257,6 +257,12 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
   if( m_runAllSyst ){
     ANA_MSG_INFO(" Running w/ All systematics");
   }
+  
+  // Write output sys names
+  if ( m_writeSystToMetadata ) {
+    TFile *fileMD = wk()->getOutputFile ("metadata");
+    HelperFunctions::writeSystematicsListHist(m_systList, m_outputSystName, fileMD);
+  }
 
   ANA_MSG_INFO( "BJetEfficiencyCorrector Interface succesfully initialized!" );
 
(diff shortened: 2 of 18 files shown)


Root/HelperFunctions.cxx
```diff
@@ -445,6 +445,20 @@ std::vector< CP::SystematicSet > HelperFunctions::getListofSystematics(const CP:
...
 }
 
+TH1D *HelperFunctions::writeListofSystematicsToHist( const std::vector< CP::SystematicSet > &systs, std::string histName, TFile *file )
+{
+  std::string name = "SystNames_" + histName;
+  if (!systs.size() || histName.empty() || file->Get(name.c_str())) {
+      return 0;
+  }
+
+  TH1D *hist = new TH1D(name.c_str(), histName.c_str(), systs.size(), 0.5, systs.size() + 0.5);
```
kratsg: it's not good to have a `new` inside a function call like this. You're going to start introducing memory leaks this way. Do you really need a `new`?

Root/HelperFunctions.cxx
```diff
@@ -445,6 +445,20 @@ std::vector< CP::SystematicSet > HelperFunctions::getListofSystematics(const CP:
 
 }
 
+TH1D *HelperFunctions::writeListofSystematicsToHist( const std::vector< CP::SystematicSet > &systs, std::string histName, TFile *file )
+{
+  std::string name = "SystNames_" + histName;
```
kratsg: the name of the histogram should encapsulate inside a systematics folder `"systematics/"+name`

Root/HelperFunctions.cxx
```diff
@@ -445,6 +445,20 @@ std::vector< CP::SystematicSet > HelperFunctions::getListofSystematics(const CP:
...
+  if (!systs.size() || histName.empty() || file->Get(name.c_str())) {
+      return 0;
+  }
+
+  TH1D *hist = new TH1D(name.c_str(), histName.c_str(), systs.size(), 0.5, systs.size() + 0.5);
+  for (size_t i = 0; i < systs.size(); i++) {
+    hist->GetXaxis()->SetBinLabel(i + 1, systs[i].name().c_str());
```
kratsg: Shouldn't we care about sorting the systematics by name, such that the nominal is always first? Also, nominal in systematics has an empty `name()` so is this accounted for or will you see an empty bin label?

Root/HelperFunctions.cxx
ntadej: Not sure, it has to be created as a pointer I guess not to be deleted at the end of function execution. As far as I understand it will be properly written out and cleaned up by the EventLoop?

Root/HelperFunctions.cxx
ntadej: For now, nominal is just an empty first bin name. It can be set to `nominal`. We shouldn't sort the systematics though so we can match them to scale factors in the vectors.

Another option for the future would be, to store scale factors in a `std::map<std::string, double>` and then they could be sorted.

Root/HelperFunctions.cxx
kratsg: This is going to cause a memory leak. Nobody's going to know that this needs to be immediately registered. You should just have this be a `void` function and pass in the objects you need.

Root/HelperFunctions.cxx
kratsg: Makes sense. I'm ok if you write it out in the same order as this is a good standard.

ntadej:
There is still `new` in the helper function but it accepts a null pointer and I renamed it to `createSystematicsListHist`. If you want to make it clearer, I can split it to "create" and "fill", but I wanted to add as much common code in the helper function as possible.

Root/HelperFunctions.cxx
```diff
@@ -468,6 +468,28 @@ std::vector< CP::SystematicSet > HelperFunctions::getListofSystematics(const CP:
...
+  if (hist || !systs.size() || histName.empty() || file->Get(name.c_str())) {
+    return;
+  }
+
+  if (!file->Get(folderName.c_str())) {
+    file->mkdir(folderName.c_str());  
+  }
+  file->cd(folderName.c_str());
+
+  hist = new TH1D(histName.c_str(), histName.c_str(), systs.size(), 0.5, systs.size() + 0.5);
```
kratsg: change this to a std::uniq_ptr and we should be ok.

ntadej: This would delete the histogram before it could be written out. This pointer should survive until the end. Other cutflow/metadata histograms are also never deleted.

Another way would be to make a histogram, write it and delete it. What do you think? I don't see a case where this histogram would be updated at later point.

kratsg: We really need to avoid the `new` somehow. Can you not pass in the file directly and write to the file inside the function instead and not expose the hist externally?

ntadej: Will do that.

PR #999, opened 2017-08-23
Fix typo in working point parsing in ElectronEfficiencyCorrector (closes #998)

### LINKED ISSUE #998: Electron trigger scale factor can't be retrieved unless electron trigger efficiency is also retrieved?
c.f.: 
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L325

`m_WorkingPointTrigTrig` is an empty string unless `m_corrFileNameTrigMCEff` is supplied. If `m_WorkingPointTrigTrig` is an empty string, the downstream decorator will be pissed, and I can't access the scale factor even further downstream. But if I only care about a scale factor, and not the efficiency, then I have no reason to supply `m_corrFileNameTrigMCEff`.

Am I reading the code correctly? Is there a reason `m_corrFileNameTrigMCEff` is required here, instead of `m_corrFileNameTrig`? The file names are identical except one is "efficiencySF..." and the other is "efficiency...".

ntadej:
We should parse the trigger working point from `m_corrFileNameTrig` not `m_corrFileNameTrigMCEff`. I haven't observed this issue as we also calculate the efficiency.

### DIFF
--- a/Root/ElectronEfficiencyCorrector.cxx
+++ b/Root/ElectronEfficiencyCorrector.cxx
@@ -311,7 +311,7 @@ EL::StatusCode ElectronEfficiencyCorrector :: initialize ()
 
     m_WorkingPointIsoTrig = HelperFunctions::parse_wp( "ISO", m_corrFileNameTrig, msg() );
     m_WorkingPointIDTrig  = HelperFunctions::parse_wp( "ID", m_corrFileNameTrig, msg() );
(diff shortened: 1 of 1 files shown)


kratsg:
This is related to #998 so perhaps this fixes that issue, but @alexandertuna needs to confirm.

PR #1000, opened 2017-08-23
Muon Efficiency Corrector simplification

ntadej:
Muon efficiency corrector simplification similar to what was done for electrons in #972. This also fixes a debug log typo for electrons.

Where could I note that this should be now run on muons before OR (as some jet systematics can cause nominal muon container to contain less muons that jet systematics one)?

This is almost the same change as for electrons. Jets might need some more work so I will do this for them later.

### DIFF
--- a/Root/MuonEfficiencyCorrector.cxx
+++ b/Root/MuonEfficiencyCorrector.cxx
@@ -389,12 +389,11 @@ EL::StatusCode MuonEfficiencyCorrector :: execute ()
 
   // if m_inputSystNamesMuons = "" --> input comes from xAOD, or just running one collection,
   // then get the one collection and be done with it
-  std::vector<std::string>* systNames(nullptr);
-  if ( m_inputSystNamesMuons.empty() ) {
-    *systNames = std::vector<std::string>{""};
-  } else {
-    ANA_CHECK( HelperFunctions::retrieve(systNames, m_inputSystNamesMuons, 0, m_store, msg()) );
-  }
+  std::vector<std::string>* systNames_ptr(nullptr);
+  if ( !m_inputSystNamesMuons.empty() ) ANA_CHECK( HelperFunctions::retrieve(systNames_ptr, m_inputSystNamesMuons, 0, m_store, msg()) );
+
+  std::vector<std::string> systNames{""};
+  if (systNames_ptr) systNames = *systNames_ptr;
 
   // Declare a write status set to true
   // For the systematically varied input containers, we won't store again the vector with efficiency systs in TStore ( it will be always the same!)
@@ -402,7 +401,7 @@ EL::StatusCode MuonEfficiencyCorrector :: execute ()
   bool writeSystNames(true);
 
   // loop over systematic sets available
-  for ( auto systName : *systNames ) {
+  for ( auto systName : systNames ) {
     const xAOD::MuonContainer* inputMuons(nullptr);
 
     // some systematics might have rejected the event
--- a/Root/MuonEfficiencyCorrector.cxx
+++ b/Root/MuonEfficiencyCorrector.cxx
@@ -506,13 +506,13 @@ EL::StatusCode MuonEfficiencyCorrector :: executeSF ( const xAOD::EventInfo* eve
   // Every systematic will correspond to a different SF!
   //
 
-  std::vector< std::string >* sysVariationNamesReco  = nullptr;
+  std::unique_ptr< std::vector< std::string > > sysVariationNamesReco = nullptr;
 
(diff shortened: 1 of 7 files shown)


xAODAnaHelpers/MuonEfficiencyCorrector.h
```diff
@@ -102,8 +93,8 @@ class MuonEfficiencyCorrector : public xAH::Algorithm
   std::vector<CP::SystematicSet> m_systListIso;  //!
   std::vector<CP::SystematicSet> m_systListTrig; //!
   std::vector<CP::SystematicSet> m_systListTTVA; //!
-
-  std::vector<std::string> m_sysNames; //!
+  
+  std::string m_outputSystNamesTrigBase; //!
```
kratsg: might want to double-check if this needs to be a class variable, I think it's only local to the `execute()` function here.

xAODAnaHelpers/MuonEfficiencyCorrector.h
```diff
@@ -64,12 +63,8 @@ class MuonEfficiencyCorrector : public xAH::Algorithm
   std::string   m_WorkingPointTTVA = "TTVA";
 
   // systematics
-  /// @brief this is the name of the vector of names of the systematically varied containers produced by the upstream algo (e.g., the SC containers with calibration systematics)
-  std::string   m_inputAlgoSystNames = "";
-  // this is the name of the vector of names of the systematically varied containers to be fed to the downstream algos. We need that as we deepcopy the input containers
-  std::string   m_outputAlgoSystNames = "";
-  // this is the name of the vector of names for the systematics to be used for the creation of a parallel container. This will be just a copy of the nominal one with the sys name appended. Use cases: MET-specific systematics.
-  std::string   m_sysNamesForParCont = "";
+  /// @brief this is the name of the vector of names of the systematically varied muons-related containers produced by the upstream algo (e.g., the SC containers with calibration systematics)
```
kratsg: `SC` -> `shallow-copied`

Root/MuonEfficiencyCorrector.cxx
```diff
@@ -1079,21 +930,16 @@ EL::StatusCode MuonEfficiencyCorrector :: executeSF ( const xAOD::EventInfo* eve
   //
   if ( !isToolAlreadyUsed(m_TTVAEffSF_tool_name) ) {
 
-    if(countSyst == 0) sysVariationNamesTTVA  = new std::vector< std::string >;
+    if ( writeSystNames ) sysVariationNamesTTVA  = new std::vector< std::string >;
```
kratsg: `sysVariationNamesTTVA` will leak memory if `writeSystNames` and `TStore` contains it already. We should match the conditions of when we record this to handle memory, or at least delete it.

Root/MuonEfficiencyCorrector.cxx
```diff
@@ -916,46 +792,31 @@ EL::StatusCode MuonEfficiencyCorrector :: executeSF ( const xAOD::EventInfo* eve
     for ( const auto& trig_it : m_SingleMuTriggers ) {
 
       std::vector< std::string >* sysVariationNamesTrig  = nullptr;
-      if(countSyst == 0) sysVariationNamesTrig  = new std::vector< std::string >;
+      if ( writeSystNames ) sysVariationNamesTrig  = new std::vector< std::string >;
```
kratsg: potential memory leak here

Root/MuonEfficiencyCorrector.cxx
```diff
@@ -720,21 +612,16 @@ EL::StatusCode MuonEfficiencyCorrector :: executeSF ( const xAOD::EventInfo* eve
   //
   if ( !isToolAlreadyUsed(m_isoEffSF_tool_name) ) {
 
-    if(countSyst == 0) sysVariationNamesIso   = new std::vector< std::string >;
+    if ( writeSystNames ) sysVariationNamesIso  = new std::vector< std::string >;
```
kratsg: potential memory leak here

xAODAnaHelpers/MuonEfficiencyCorrector.h
ntadej: The string is set at `initialize` and then used in `execute`. The reason is to support configurability of `m_outputSystNamesTrig` and properly adapt to specific trigger working point (as other working points are also appended). The whole functionality in `initialize` could also be moved to `execute`.

ntadej:
I've added some pointer protection for syst. names vectors and updated docs. I also ported #1002 to this PR for muons (thanks for spotting this).

PR #1014, opened 2017-08-28
Move fJVT calculation to JetCalibrator

ntadej:
In continuous effort to try to reduce the amount of needed deep copies, I moved fJVT calculation to JetCalibrator. JVT is also recalculated there so this might be a better location for it.

Two configuration properties are also supported now:
 - `CentralMaxPt` - to limit max pt of central jets, required by MET recommendations
 - `UseTightOP` - if one wants to use Tight working point

Note that this is a breaking change. I'm open to suggestions if some of the introduced `JetCalibrator` properties should be renamed.


kratsg:
@jdandoy what do you think?

kratsg:
@jdandoy gave a thumbs-up during a quick chat with him, and it's a good idea to move this over.

PR #1017, opened 2017-08-29
Add MET significance calculation

ntadej:
This adds optional MET significance calculation to METConstructor:
 - the outputs are grouped in 2 groups - one containing all different significances, the other resolution-related variables
 - significance can be only stored for "Clus" or "Trk" MET type
 - added ability to store only one type of MET (if `metTrk` is set, "Clus" type is disabled)

As always, I'm open to property name changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ atlas_depends_on_subdirs( PUBLIC
                           Reconstruction/Jet/JetResolution
                           Reconstruction/Jet/JetSubStructureUtils
                           Reconstruction/Jet/JetUncertainties
+                          Reconstruction/MET/METInterface
                           Reconstruction/MET/METUtilities
                           Tools/PathResolver
                           Trigger/TrigAnalysis/TrigDecisionTool
--- a/.jenkins.yml
+++ b/.jenkins.yml
@@ -1,5 +1,5 @@
 env:
-  - ABV_RC=2.4.37 ABV_CM=21.2.0
+  - ABV_RC=2.4.37 ABV_CM=21.2.3
 
 before_script:
   - ls
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -374,6 +374,8 @@ namespace HelperClasses{
   }
 
   void METInfoSwitch::initialize(){
+    m_metClus   = has_exact("metClus");
+    m_metTrk    = has_exact("metTrk");
     m_sigClus   = has_exact("sigClus")  || has_exact("all");
     m_sigTrk    = has_exact("sigTrk")   || has_exact("all");
     m_sigResolutionClus = has_exact("sigResolutionClus") || has_exact("all");
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -374,6 +374,10 @@ namespace HelperClasses{
   }
 
   void METInfoSwitch::initialize(){
+    m_sigClus   = has_exact("sigClus")  || has_exact("all");
+    m_sigTrk    = has_exact("sigTrk")   || has_exact("all");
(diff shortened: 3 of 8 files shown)


ntadej:
I'm not sure about the Jenkins error. I'm using 2.4.37-based METInterface (METInterface-00-02-11) and it works fine locally ...

kratsg:
@ntadej the error is because this breaks our CMake releases (rel 21)

```
[ 63%] Building CXX object xAODAnaHelpers/CMakeFiles/xAODAnaHelpersLib.dir/Root/METConstructor.cxx.o
In file included from /home/jenkins/workspace/RootCore/xAH/xAODAnaHelpers/Root/METConstructor.cxx:7:0:
/home/jenkins/workspace/RootCore/xAH/xAODAnaHelpers/xAODAnaHelpers/METConstructor.h:12:43: fatal error: METInterface/IMETSignificance.h: No such file or directory
 #include "METInterface/IMETSignificance.h"
                                           ^
compilation terminated.
```

looking at the [log attached](http://build.ci-connect.net:8080/job/RootCore/job/xAH/1027/console). However I do see it [here](https://gitlab.cern.ch/atlas/athena/blob/21.2/Reconstruction/MET/METInterface/METInterface/IMETSignificance.h) so it must be in rel 21... but we don't link against it yet - so you will need to update the CMakeLists.txt to do so (similar to METUtilities): https://github.com/UCATLAS/xAODAnaHelpers/blob/master/CMakeLists.txt#L66

ntadej:
This header is available since 21.2.1 and is not in 21.2.0 so no CMake modification is needed. Should we bump xAH to 21.2.3? Might be a good idea.

kratsg:
> This header is available since 21.2.1 and is not in 21.2.0 so no CMake modification is needed. Should we bump xAH to 21.2.3? Might be a good idea.

Sure,, but you still have a dependency on the METInterfaces library which we do *need* to add to the CMakeLists.txt.

ntadej:
Then we should already depend on it as there is already `IMETMaker` and `IMETSystematicsTool` there. Also it seems to be automatically added (or it just works as AnalysisBase is build with it and nothing needs to be linked, not sure). I will add it to CMakeLists.txt to be on the safe side.

kratsg:
> Then we should already depend on it as there is already IMETMaker and IMETSystematicsTool there. Also it seems to be automatically added (or it just works as AnalysisBase is build with it and nothing needs to be linked, not sure). I will add it to CMakeLists.txt to be on the safe side.

Yeah, we should be adding these. I would suggest you update to 21.2.3 (update the `.jenkins.yml` file accordingly).

CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ atlas_depends_on_subdirs( PUBLIC
                           Reconstruction/Jet/JetResolution
                           Reconstruction/Jet/JetSubStructureUtils
                           Reconstruction/Jet/JetUncertainties
+                          Reconstruction/MET/METInterfaces
```
kratsg: Should be called "METInterface".

kratsg: Also add a link to the library generated in the `LINK_LIBRARIES` section of the `atlas_add_library` command. The name of the library generated is `METInterface`

https://gitlab.cern.ch/atlas/athena/blob/master/Reconstruction/MET/METInterface/CMakeLists.txt

ntadej: Sorry for the mistake.

Shouldn't cmake complain though? Do you think this should be reported upstream to Athena?

kratsg: It's a linking thing. It'll show up if you try to run on the grid, or during run-time. Dependency resolution is during compile-time, but it doesn't quite check that the library is there.

ntadej: Done. It seems we already linked to METInterface but it wasn't added as a dependency.

PR #1018, opened 2017-08-29
Add prompt lepton variables to electrons and muons if desired

alexandertuna:
It's not your grandpappy's isolation.

https://twiki.cern.ch/twiki/bin/view/AtlasProtected/PromptLeptonTagging

For now, just grab the variables from the input DAOD. I don't think anyone is recalculating it offline, to my knowledge.


kratsg:
I'm fine with this change. Will this be eventually a tool, or just there for maybe offline studies in the future? One might want to add this to the documentation somewhere so people know where to find information about it, maybe in https://xaodanahelpers.readthedocs.io/en/master/ToolsUsed.html [docs/ToolsUsed.rst](docs/ToolsUsed.rst)

alexandertuna:
I'll add doc stuff to the commit in the morning.

For now, there are no specific tools to the tagger for xAH. They have a tool for the AOD->DAOD step, but I haven't seen anything publicized about how to recalculate the decisions from DAODs. And they have scale factors, but they're already wrapped into the existing lepton SF tools.

PR #1027, opened 2017-09-13
Store systematic variations of mcEventWeight

ntadej:
It can be useful to have systematic variations of mcEventWeight stored - they contain theoretical/generator uncertainties.

A summary of changes:
 - new `weightsSys` flag for `EventInfo` output
 - only stored for nominal tree (we always have to care about the output size ...)

Unfortunately it's impossible to read the actual names of the systematic variations using RootCore/AnalysisBase.

### DIFF
--- a/Root/EventInfo.cxx
+++ b/Root/EventInfo.cxx
@@ -8,8 +8,8 @@
 
 using namespace xAH;
 
-EventInfo::EventInfo(const std::string& detailStr, float units, bool mc)
-  : m_infoSwitch(detailStr), m_mc(mc), m_debug(false), m_units(units)
+EventInfo::EventInfo(const std::string& detailStr, float units, bool mc, bool storeSyst)
+  : m_infoSwitch(detailStr), m_mc(mc), m_debug(false), m_storeSyst(storeSyst), m_units(units)
 {
 }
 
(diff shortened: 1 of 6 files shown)


kratsg:
> Unfortunately it's impossible to read the actual names of the systematic variations using RootCore/AnalysisBase.

What do you mean by this?

ntadej:
Reading metadata from folders like '/Generation/Parameters' is not supported with RootCore accorting to Zach and Attila (we had a discussion on PATHelp today). This information is stored in `xAOD::TruthMetaData` but it seems only TRUTH derivations have it (at least EXOT do not).

For our analyses I created a simple AthAnalysis dump script which creates a histogram like I made for systematics names. I then manually add it to metadata in our xAH wrapper.

kratsg:
> Reading metadata from folders like '/Generation/Parameters' is not supported with RootCore accorting to Zach and Attila (we had a discussion on PATHelp today). This information is stored in xAOD::TruthMetaData but it seems only TRUTH derivations have it (at least EXOT do not).

Yeah, the `TagInfo` in MetaData isn't in AnalysisBase, so that just won't work. It's just another slight reason why being dual-use is pretty hard from an analysis perspective.

PR #1037, opened 2017-09-25
Updates and corrections for bcid, rho, and PRW tool

jdandoy:
Various corrections and updates.
1) Apply PRW to data, necessary for corrected average mu and data weight (see [twiki](https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/ExtendedPileupReweighting))
2) Add corrected average mu to data tree, remove prescale_DataWeight from pileup flag, and add bcid to MC tree
3) Add DistEmptyBCID calculation and to data tree.  This is the distance from this bcid to the previous empty bcid
4) Update Rho for R21, fix LCTopo version

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -737,14 +737,16 @@ EL::StatusCode BasicEventSelection :: execute ()
   //------------------------------------------------------------------------------------------
   // Update Pile-Up Reweighting
   //------------------------------------------------------------------------------------------
-  if ( m_isMC && m_doPUreweighting ) {
+  if ( m_doPUreweighting ) {
     m_pileup_tool_handle->applySystematicVariation(CP::SystematicSet()).ignore();
     m_pileup_tool_handle->apply( *eventInfo ); // NB: this call automatically decorates eventInfo with:
                                                  //  1.) the PU weight ("PileupWeight")
                                                  //  2.) the corrected mu ("corrected_averageInteractionsPerCrossing")
                                                  //  3.) the random run number ("RandomRunNumber")
                                                  //  4.) the random lumiblock number ("RandomLumiBlockNumber")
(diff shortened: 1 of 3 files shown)


kratsg:
Overarching comments: do these changes affect both r20 as well as r21? Because as I understand, we don't use PURW for r20 data, but we need it for r21 data? Similar for the rho calculation which looks like it only works in r21?

jdandoy:
Hi @kratsg 

I believe we need the PRW tool for both r20.7 and r21 (why wouldn't we?).  We should have the corrected average mu for both 2016 and 2017 data.  

The rho calculation (performed when setting eventDetailStr to shapeEM or shapeLC) was previously broken in r21, and would now be broken in r20.7.  I can try to get it to work in both.

kratsg:
Yeah, I guess I'm not sure if it was to be applied on 2015 data. In fact, it isn't clear from the twiki on which data this needs to be applied or how this changes between r20 and r21 [if it's a change we make or the tool handles this logic internally].

> The rho calculation (performed when setting eventDetailStr to shapeEM or shapeLC) was previously broken in r21, and would now be broken in r20.7. I can try to get it to work in both.

You can use (for now) the `#ifdef USE_CMAKE` to switch between r20 (RootCore) and r21 (CMake). This will make it easier for us to fully migrate to r21 by groking for those `ifdef` statements and removing them.

jdandoy:
The twiki states that getCorrectedMu "only changes values in data, not in MC. And it only changes values in data when the luminosity tag used to make the lumicalc file differs from the one used at the time the input (D)xAOD was actually made."

So it shouldn't have an effect on 2015 and 2016 data, but it doesn't hurt to include.  R21 reprocessing likely corrected the mu in the AOD.  The tool seems to handle everything internally.

PR #1039, opened 2017-09-28
Add hybrid b-tagging working point option to jets.

kkrizka:
Added option to select hybrid b-tagging working points to HelperClasses.

For more info, see:
https://twiki.cern.ch/twiki/bin/view/AtlasProtected/BTagCalib2017#Pre_Recommendation_August_2017


kratsg:
I'm ok with this change. Off-hand, do you think we should make this particular type of pattern-parsing a built-in function so it doesn't need to be duplicated?

kkrizka:
Yeah, it would be useful to have a generic function for this.

ie: split attributenameXXYYZZ to a vector containing XX,YY,ZZ...

However I will merge this as it is for now.

--
Karol Krizka

kratsg:
> Yeah, it would be useful to have a generic function for this.
> 
> ie: split attributenameXXYYZZ to a vector containing XX,YY,ZZ...

I'll file an issue for this.

PR #1041, opened 2017-09-28
Revert deep copy of b-tagged jet containers.

kkrizka:
This reverts the deep-copying of all jet containers introduced in #955 just so a single decoration can be added. This reverts back to the old behaviour of the algorithm with no need for `m_outContainerName`.

@ntadej, can you comment why you think that was necessary? I copied the comment that I put on your closed merge request for discussion.

What exactly is being overwritten? My understanding is that all systematic variations of jets used as inputs are already individual shallow copies. All that `BJetEfficiencyCorrector` does is set a decoration (tagging decision + scale factor) for each shallow copy jet. To me making container copies for all systematics seems unneeded. Am I missing a specific systematics flow here?

Furthermore why do we need deep copies instead of shallow copies of all jets? Isn't the whole point of a shallow copy to be able to modify the attributes of an object without having to make a copy of all the other "stable" attributes?

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -287,13 +287,10 @@ EL::StatusCode BJetEfficiencyCorrector :: execute ()
   // then get the one collection and be done with it
   //
   if ( m_inputAlgo.empty() ) {
-    // Check the existence of input container
-    if ( m_store->contains<xAOD::JetContainer>( m_inContainerName ) ) {
-      // this will be the collection processed - no matter what!!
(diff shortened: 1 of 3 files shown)


ntadej:
I will do this a proper way as I’ve done for electrons and muons when I come back from holidays. It’s hard for me to remember why this was needed in the first place, but I encountered some strange crashes.

For now this was just copied from the old muon implementation. There is no need for deep copies but systematics need to be properly handled.

kratsg:
Thanks @kkrizka for being vigilant about this. This was my fault for not being more careful.

PR #1044, opened 2017-10-03
Minor cleanup of BJetEfficiencyCorrector

ntadej:
@kkrizka already fixed most of my mistakes. Again sorry for that, that was one of my first PRs. This additionally cleans-up BJetEfficiencyCorrector:
 - Use common code with or without SYS.
 - Check the existence of input container - just in case.
 - Is `m_outputAlgo` needed? I removed it for now as this algorithm doesn't change any systematics but I can add it back.

I also added the disclaimer for `m_inputAlgo` that only jet systematics with separate trees should be used. I kept the same name for now (although I renamed it for electrons and muons to be more clear e.g. to `m_inputSystNamesElectrons`).

It would be nice if @kkrizka can review this.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -277,88 +277,28 @@ EL::StatusCode BJetEfficiencyCorrector :: execute ()
   ANA_CHECK( HelperFunctions::retrieve(eventInfo, m_eventInfoContainerName, m_event, m_store, msg()) );
   ANA_MSG_DEBUG("\n\n eventNumber: " << eventInfo->eventNumber() << std::endl );
 
-  //
-  //  input jets
-  //
-  const xAOD::JetContainer* inJets(nullptr);
-
-  //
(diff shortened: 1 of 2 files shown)


kkrizka:
Hi @ntadej,

Looks good to me.

I don't think `m_outputAlgo` is needed and I agree with you about the removal. Does anyone use it?

Also for my education, how does the `BJetEfficiencyCorrector` work with MET systematic? I see you removed the `m_sysNamesForParCont` part that you introduced earlier to handle them (at least according to the code comments).

--
Karol Krizka

ntadej:
`BJetEfficiencyCorrector` should be applied only to jet systematics. All other systematics should contain the same jets.

Also (for now) all SF calculations should be done before OR and on specific object systematics (e.g. `BJetEfficiencyCorrector` on jet systematics with own trees). As we have shallow copies we only need to process each "tree" once. Somehow sizes of the vectors should be checked and decided if SF calculation is needed or not so we are safe on which jets we run the algorithm.

PR #1047, opened 2017-10-09
Make `m_significanceSoftTermReso` a `double` instead of `float`

vpascuzz:
In `METUtilities::METSignificance` the varible `m_softTermReso` that
is assigned through xAH's `METConstructor` is a `double`. Need
consistent typing for property-setting, otherwise crash.


kratsg:
go ahead

kratsg:
The only bug this will introduce (offhand) is that a python float to a C++ double may not be the best conversion [accidental loss of precision].

vpascuzz:
Isn't a `double` higher precision than a `float`? How is precision lost here?

kratsg:
@vpascuzz good point, I need more sleep.

PR #1054, opened 2017-10-17
Add support for the b-tagging algorithms provided in R21.

kkrizka:
This implements the working points of the multiple b-taggers available in R21. More information:
https://twiki.cern.ch/twiki/bin/view/AtlasProtected/BTaggingBenchmarksRelease21
I've listed all of the changes below. There are a few backward non-compatible changes, but support for reading existing ntuples is maintained.

* `BJetEfficiencyCorrector` now appends the tagger name (in addition to the working point) to the decoration names.

* `JetInfoSwitch` for different b-taggers is now set using the `jetBTag_tagger_type_AABB..MM..YY.ZZ` option, with `tagger` set to the algorithm name (MV2c10, MV2c10rnn, DL1...), `type` set to the efficiency point definition (ie: `HybBEff`) and `AABB` the requested efficiency as before. This replaces the old `sfFTagFix/Flt/Hyb` settings. The old settings are kept and are used when loading older ntuples. The reading of old ntuples should be removed in the future.

* The branch name with the decision now includes the tagger name. The old branches `jet_isHyb77` and `jet_SFHyb77` are now `jet_is_MV2c10rnn_HybBEff_77` and `jet_SF_MV2c10rnn_HybBEff_77` (replace MV2c10rnn with your tagger of choice).

* The `xAH::Jet` attribute names also follow the same scheme. I've kept them hardcoded, because I see significant performance gains. This is opposed to having a map with the algorithm and working point as the key. This limits the taggers that can be used to the supported one. Since this is used during the histogramming step that is run often, I think optimizing for performance is good.

* When reading old ntuples (aka using `sfFTagFix/Flt/Hyb`), the `xAH::Jet` attributes corresponding to MV2c10 are filled.

* The `xAH::is_btag` and `SF_btag` are helper functions that return the b-tagging decision and scale factor for the tagger given as the argument. The argument is of type `xAH::Jet::BTaggerOP`, which is an enum with the supported taggers. These functions are helped for setting the tagger as an option in an xAH_run config file.


ntadej:
I have one wish here, that might be nice to be done now that you are making substantial changes. Would it be possible to make global SF optional (or be able to disable it)? We are not using it and it just takes CPU and space in the output.

kratsg:
This looks good, I am curious about why we decided (unlike the other algorithms) to try and handle multiple btagging working points in the same TreeAlgo, instead of having multiple ones write to the same tree. They should all technically not overlap, or there's something else I'm missing?

kkrizka:
@ntadej I've killed it from both `BJetEfficiencyCorrector` and `JetContainer`. Based on the previous implementation in the `JetContainer`, I'm not sure if it was even used. I'm going to wait until tomorrow morning before merging this request. Also this feature is easy to readd in the future if someone complains.

Similarly, would someone complain if I removed the `njets_tagger` branch?

@kratsg What do you mean? A `TreeAlgo` creates its own `TTree` that it writes to. I don't see currently how you can have multiple `TreeAlgo`'s write to the same `TTree`.

kratsg:
> @kratsg What do you mean? A TreeAlgo creates its own TTree that it writes to. I don't see currently how you can have multiple TreeAlgo's write to the same TTree.

You can retrieve a pointer to a previously created TTree just by knowing it's name, since it's written to the output. So a TreeAlgo could be configured to create a tree, or write to an existing one.. I think.

PR #1071, opened 2017-10-26
TreeAlgo only looks in TStore for containers, use TEvent too

### LINKED ISSUE #1053: Compile-time warning for HelperFunctions::retrieve instead
```
HelperFunctions          WARNING retrieve<T>(..., bool) is deprecated. See https://github.com/UCATLAS/xAODAnaHelpers/pull/882
```

is a runtime warning. We should be less mean to our users.

kratsg:
@mattleblanc found an interesting bug where `TStore::contains<>` only looks inside `TStore` while `TStore::retrieve<>` can grab from both `TStore` and `TEvent`. For now, use the existing `HelperFunctions::isAvailable<>` to handle this.

This also will close #1053, so @mswiatlo will be happier.


kratsg:
test this please

Issue #1072, opened 2017-10-27
JERTool crashes due to MC property assignment error


kkrizka:
I am seeing the ntupling code crash with the following error. Seems to be caused by #1069. Replacing `isMC()` with `true` (`isMC()==1` in my case) when setting the `isMC` property of the tool gets rid of the crash.

> PropertyMgr              ERROR   /build1/atnight/localbuilds/nightlies/21.2/athena/Control/AthToolSupport/AsgTools/AsgTools/PropertyMgr.icc:43 (StatusCode PropertyMgr::setProperty(const string&, const T&) [with T = int; std::__cxx11::string = std::__cxx11::basic_string<char>]): Value assignment failed for isMC
> AsgTools_AnaToolHandle   ERROR   /build1/atnight/localbuilds/nightlies/21.2/athena/Control/AthToolSupport/AsgTools/AsgTools/AnaToolHandle.icc:240 (StatusCode asg::detail::AnaToolPropertyValueRootCore<Type>::applyPropertyRootCore(asg::AsgTool&, const string&, asg::detail::AnaToolCleanup&) const [with Type = int; std::__cxx11::string = std::__cxx11::basic_string<char>]): Failed to call "tool.setProperty (name, m_value)"
> AsgTools_AnaToolHandle   ERROR   /build1/atnight/localbuilds/nightlies/21.2/athena/Control/AthToolSupport/AsgTools/Root/AnaToolHandle.cxx:429 (StatusCode asg::detail::AnaToolConfig::makeToolRootCore(const string&, asg::IAsgTool*&, asg::detail::AnaToolCleanup&) const): Failed to call "property.second->applyPropertyRootCore (*rawToolPtr, property.first, mycleanup)"
> AsgTools_AnaToolHandle   ERROR   /build1/atnight/localbuilds/nightlies/21.2/athena/Control/AthToolSupport/AsgTools/Root/AnaToolHandle.cxx:515 (StatusCode asg::detail::AnaToolConfig::makeBaseTool(const string&, asg::parentType_t*, ToolHandle<asg::IAsgTool>&, asg::detail::AnaToolCleanup&) const): Failed to call "makeToolRootCore (toolName, baseToolPtr, baseCleanup)"
[...]

PR #1075, opened 2017-10-27
Algorithm::isMC now returns a boolean. If the type cannot be determin…

kkrizka:
isMC should be a yes/no answer. Or unknown, in which case the user should be told to provide the correct answer.

See #1072 for why this is needed.


kkrizka:
Ok, I've added text to the exception message that it comes from `isMC()`. The user can then read documentation about `isMC()` to fix it.

PR #1083, opened 2017-11-06
Moving to the 2017 configuration for Jet hist

johnda102:


kratsg:
Hey @johnda102 this breaks rel20.7. For now, can you wrap the release-21 specific changes in the 

```c++
#ifdef USE_CMAKE 
...
#endif
```

johnda102:
what breaks in 20.7 ?
Is there a way for me to see the compile log

kratsg:
> what breaks in 20.7 ?
> Is there a way for me to see the compile log

You can click "details" for the "travis-ci" to see the rel 21 compile logs. But I was wondering if the moving to 2017 configuration would break release 20.7 analyses, or they should be moving to this configuration anyway?

johnda102:
ok no this should be backward compatible with the 20.7 analyses. 
This is for the trigger performance in 2017.  Which is independent of AB release

PR #1084, opened 2017-11-06
Add minimum and maximum cuts on actual pileup

matthewfeickert:
The cuts on eventInfo->actualInteractionsPerCrossing() allow for
selection of specific pile-up values in a run. This is motivated by a
need case in StudyTrigTracks.


matthewfeickert:
There might be a better place to have put my changes in the files, so if anyone has comments on that please let me know.

Issue #1088, opened 2017-11-08
Electron efficiency scale factors work down to 7 GeV, not 15 GeV


alexandertuna:
15 GeV is hard-coded. But the recommendations nowadays are 7 GeV (if not 4.5 GeV). Does anyone object to me changing this hardcode to 7 GeV?

https://twiki.cern.ch/twiki/bin/view/AtlasProtected/LatestRecommendationsElectronIDRun2

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L642
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L745
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L848
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L957
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L1064

@kratsg @mmilesi @ntadej @MihaMuskinja @jdandoy @kkrizka

alexandertuna:
In case merges affect the line number, just search for "15e3".

ntadej:
Why not make this configurable? The recommendation might change again.

alexandertuna:
New strategy: make it configurable. I'll make the default 15e3 to not annoy anyone.

PR #1093, opened 2017-11-11
Make threshold for electron efficiency scale factors configurable

alexandertuna:
cf https://github.com/UCATLAS/xAODAnaHelpers/issues/1088


alexandertuna:
Tested and successful on 10k W events.

kratsg:
Can you do two things?

- rename `m_threshold` to `m_ptThreshold`
- update the comment (documentation) on the variable to mention that this is for defining the acceptance region for scale factors, or something like that.

PR #1094, opened 2017-11-12
Add MV2c10 as a valid HLTBTagTaggerName

matthewfeickert:
Currently `JetSelector` only selects `MV2c20` and `COMB` as valid strings for `m_HLTBTagTaggerName` for HLT b-tagging. This adds `MV2c10`.


matthewfeickert:
Given @kratsg's comment on Issue #1095, should I change this to just replace `MV2c20`?

kratsg:
This is fine as it is.

PR #1100, opened 2017-11-19
New JetCalibTool option to set CalibArea manually

mattleblanc:
This requires `21.2.10`, so I'm not exactly sure how you want to proceed, @kratsg?

Changes are pretty minor, just adding the new option. Default value is the current recommendation from JetEtMiss.

🍻


kratsg:
> This requires 21.2.10, so I'm not exactly sure how you want to proceed, @kratsg?

Are you saying the option is only available on the tool in 21.2.10+?

mattleblanc:
Yes

kratsg:
Update `.travis.yml` and remove all tags, and only add 21.2.10 in there and make sure we do not allow it to fail. Push that and let it test on that version and we'll push. I suspect we'll just have to stop supporting pre 21.2.0 tags.

kratsg:
Also, wrap this in `#ifdef USE_CMAKE` as well, because this is probably not in the SVN releases at all (2.4.X).

jbossios:
> Also,  wrap this in #ifdef USE_CMAKE as well, because this is probably not in the SVN releases at all (2.4.X).

It's not in the SVN releases indeed.

mattleblanc:
Wrapped the things. I've never modified the `travis.yml` before, so that's probably going to need some fixin' ...

.travis.yml
```diff
@@ -10,11 +10,7 @@ env:
...
     - secure: "NKEq9tEPPQhcEJWgy06LaTRjVObqb57GMllFrWa8zQ7/a6kVr4l0VMSqREeDeLrwIXo3RK3nG01sBFavy1Ge4RqUhPjbSBln/xXSEZiMSA1PI97FX3L+emOum4knVIudR1V3WXwRCxK138Fpwudbf0Hi/zLNh/stmApCStWBHmyCrxbfmnI3n+cTKC6RvINY6DJi9QbagCaZCfmLYGSqiNzoQbMEm5T5EkB8xdPFwSoGD7KTBwpcjkqhpS+lGoxcxdXimyEN5lmoR9+EIw6ZUZJzeaANHui6H2keBDe4tJaDeAqZeWkpfw8ixvJO/2OTtPEU5y4DcPzVW3x0GO4B/97oCLZI9PrYDiEq9inBL3xmUmeqXcBV8PLnUVzbosvd0W1kSWiw9Cude223of5VNpiXP7keCNbsklIwWfMNnlDNOKlrI0k+J5dvdf3gH3E+166FykA78/PNgiBCTiZRACuCY00170LLPnXCHFvDvZcUsDhGx8XbxQGEyPfjKyq5Fgythexg03PGW0LjxYR275RLGxa5o4fd/AoZiLQY+NVs1TZeexMIvgjHk03FpHKbCxSzRalCVeibZymc7pnH3huWMuOfPM2It5cd2MlH3iR/M9wln+RNkU+VUsu205PdUbnXYc3YFCy30vUdvb3B2CQntwfiU1sH23ZB9uVR6rE="
   matrix:
```
kratsg: add in here

`- ABV_CM=21.2.7 ALLOWFAIL=true`

.travis.yml
```diff
@@ -37,8 +33,7 @@ deploy:
 matrix:
   fast_finish: true
   allow_failures:
-    - env: ABV_CM=21.2.3 ALLOWFAIL=true
-    - env: ABV_CM=21.2.4 ALLOWFAIL=true
+    - env: ABV_CM=21.2.10 ALLOWFAIL=false
```
kratsg: We don't want to allow 21.2.10 to fail, so remove this and change it with

`- env: ABV_CM=21.2.7 ALLOWFAIL=true`

PR #1102, opened 2017-11-21
xrootd check is only needed for rel20

### LINKED ISSUE #1101: Why is xrootd required for prun
Since moving to rel. 21.2, the usual `lsetup panda` for prun grid submission is not enough anymore. If I comment out XRD requirement, jobs still seem to be submitted fine. What is the reason for this?

kratsg:
Closes #1101.


ntadej:
If I use scanRucio (which is required for inputs from the grid) I still need this, although it runs fine without it.

kratsg:
> If I use scanRucio (which is required for inputs from the grid) I still need this, although it runs fine without it.

It checks it still if you use scanRucio, please look at the linked issue for the discussion there.

PR #1104, opened 2017-11-21
Update interface paths and fix other warnings

vpascuzz:
Get rid of brutal warnings that flood the screen during compilation with this pull request.

`CMakeLists.txt`: add updated interface paths
`Root/BJetEfficiencyCorrector.cxx`: comment unused parameter
`Root/MuonEfficiencyCorrector.cxx`: comment unused parameter
`xAODAnaHelpers/BJetEfficiencyCorrector.h`: update interface paths
`xAODAnaHelpers/JetSelector.h`: update interface path
`xAODAnaHelpers/MuonSelector.h`: update interface path

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,16 +5,13 @@
 # Declare the package name:
 atlas_subdir( xAODAnaHelpers )
 
-set( release_libs xAODTriggerCnvLib FTagAnalysisInterfacesLib MuonAnalysisInterfacesLib )
-set( release_depends PhysicsAnalysis/Interfaces/FTagAnalysisInterfaces PhysicsAnalysis/Interfaces/MuonAnalysisInterfaces )
+set( release_libs xAODTriggerCnvLib )
 if( ${AnalysisBase_VERSION} VERSION_GREATER 21.0 )
   # 21.X
-  set( release_libs xAODTriggerCnvLib FTagAnalysisInterfacesLib MuonAnalysisInterfacesLib )
-  set( release_depends PhysicsAnalysis/Interfaces/FTagAnalysisInterfaces PhysicsAnalysis/Interfaces/MuonAnalysisInterfaces )
+  set( release_libs xAODTriggerCnvLib )
 else()
   # 2.6.X
   set( release_libs xAODTriggerCnv )
-  set( release_depends PhysicsAnalysis/JetTagging/JetTagPerformanceCalibration/xAODBTaggingEfficiency PhysicsAnalysis/MuonID/MuonSelectorTools )
 endif()
 
 # Declare the package's dependencies:
@@ -54,6 +51,8 @@ atlas_depends_on_subdirs( PUBLIC
                           PhysicsAnalysis/ElectronPhotonID/ElectronPhotonShowerShapeFudgeTool
                           PhysicsAnalysis/ElectronPhotonID/PhotonEfficiencyCorrection
                           PhysicsAnalysis/JetMissingEtID/JetSelectorTools
+                          PhysicsAnalysis/Interfaces/FTagAnalysisInterfaces
+                          PhysicsAnalysis/Interfaces/MuonAnalysisInterfaces
                           PhysicsAnalysis/MuonID/MuonIDAnalysis/MuonMomentumCorrections
                           PhysicsAnalysis/MuonID/MuonIDAnalysis/MuonEfficiencyCorrections
                           PhysicsAnalysis/TauID/TauAnalysisTools
@@ -96,6 +95,7 @@ atlas_add_library( xAODAnaHelpersLib xAODAnaHelpers/*.h Root/*.h Root/*.cxx ${xA
                    ElectronEfficiencyCorrectionLib ElectronPhotonSelectorToolsLib
                    IsolationSelectionLib IsolationCorrectionsLib
                    ElectronPhotonShowerShapeFudgeToolLib
+                   FTagAnalysisInterfacesLib MuonAnalysisInterfacesLib
                    PhotonEfficiencyCorrectionLib METUtilitiesLib METInterface
                    TauAnalysisToolsLib AsgTools xAODMissingET JetResolutionLib
                    AssociationUtilsLib JetEDM JetUncertaintiesLib
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,11 +50,11 @@ atlas_depends_on_subdirs( PUBLIC
                           PhysicsAnalysis/ElectronPhotonID/IsolationCorrections
                           PhysicsAnalysis/ElectronPhotonID/ElectronPhotonShowerShapeFudgeTool
                           PhysicsAnalysis/ElectronPhotonID/PhotonEfficiencyCorrection
+                          PhysicsAnalysis/Interfaces/FTagAnalysisInterfaces
+                          PhysicsAnalysis/Interfaces/MuonAnalysisInterfaces
                           PhysicsAnalysis/JetMissingEtID/JetSelectorTools
-                          PhysicsAnalysis/JetTagging/JetTagPerformanceCalibration/xAODBTaggingEfficiency
                           PhysicsAnalysis/MuonID/MuonIDAnalysis/MuonMomentumCorrections
                           PhysicsAnalysis/MuonID/MuonIDAnalysis/MuonEfficiencyCorrections
-                          PhysicsAnalysis/MuonID/MuonSelectorTools
                           PhysicsAnalysis/TauID/TauAnalysisTools
                           Reconstruction/Jet/JetAnalysisTools/JetTileCorrection
                           Reconstruction/Jet/JetCalibTools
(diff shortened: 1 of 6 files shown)


kratsg:
These changes are going to break release 20. You need to wrap this in `#ifdef USE_CMAKE` instead.

vpascuzz:
Ah...I forgot about the back-compat. Will fix.

vpascuzz:
OK, `PhysicsAnalysis/Interfaces/MuonAnalysisInterfaces` was `21.2.5`...

vpascuzz:
Can we assume everyone is using >`21.2.6`? :)

kratsg:
we're on 21.2.10 now, so we can migrate in rel21.

vpascuzz:
OK.

vpascuzz:
Not sure what test failed for `RootCore`, since all checks reported as passed.

vpascuzz:
Something holding up this PR?

CMakeLists.txt
```diff
@@ -5,13 +5,16 @@
 # Declare the package name:
 atlas_subdir( xAODAnaHelpers )
 
-set( release_libs xAODTriggerCnvLib )
+set( release_libs xAODTriggerCnvLib FTagAnalysisInterfacesLib MuonAnalysisInterfacesLib )
```
kratsg: why is this added like this?

CMakeLists.txt
```diff
@@ -5,13 +5,16 @@
...
+set( release_depends PhysicsAnalysis/Interfaces/FTagAnalysisInterfaces PhysicsAnalysis/Interfaces/MuonAnalysisInterfaces )
 if( ${AnalysisBase_VERSION} VERSION_GREATER 21.0 )
   # 21.X
-  set( release_libs xAODTriggerCnvLib )
+  set( release_libs xAODTriggerCnvLib FTagAnalysisInterfacesLib MuonAnalysisInterfacesLib )
+  set( release_depends PhysicsAnalysis/Interfaces/FTagAnalysisInterfaces PhysicsAnalysis/Interfaces/MuonAnalysisInterfaces )
 else()
   # 2.6.X
   set( release_libs xAODTriggerCnv )
+  set( release_depends PhysicsAnalysis/JetTagging/JetTagPerformanceCalibration/xAODBTaggingEfficiency PhysicsAnalysis/MuonID/MuonSelectorTools )
```
kratsg: we don't really support 2.6.X now.

Root/MuonEfficiencyCorrector.cxx
```diff
@@ -493,7 +493,11 @@ EL::StatusCode MuonEfficiencyCorrector :: histFinalize ()
   return EL::StatusCode::SUCCESS;
 }
 
+#ifdef USE_CMAKE
```
kratsg: this ifdef makes no sense here. The same change should be in both r20 and r21 if you're commenting this out...

xAODAnaHelpers/BJetEfficiencyCorrector.h
```diff
@@ -12,8 +12,13 @@
 #include "PATInterfaces/ISystematicsTool.h"
 
 // external tools include(s):
+#ifdef USE_CMAKE
+#include "FTagAnalysisInterfaces/IBTaggingSelectionTool.h"
+#include "FTagAnalysisInterfaces/IBTaggingEfficiencyTool.h"
+#else // RootCore back-compat
```
kratsg: no need to comment

xAODAnaHelpers/JetSelector.h
```diff
@@ -27,7 +27,11 @@
 #include "AsgTools/AnaToolHandle.h"
 #include "JetJvtEfficiency/IJetJvtEfficiency.h"
 #include "JetInterface/IJetModifier.h"
+#ifdef USE_CMAKE
+#include "FTagAnalysisInterfaces/IBTaggingSelectionTool.h"
+#else // RootCore back-compat
```
kratsg: no need to comment

xAODAnaHelpers/MuonSelector.h
```diff
@@ -18,7 +18,11 @@
 // external tools include(s):
 #include "AsgTools/AnaToolHandle.h"
 #include "IsolationSelection/IIsolationSelectionTool.h"
+#ifdef USE_CMAKE
+#include "MuonAnalysisInterfaces/IMuonSelectionTool.h"
+#else // RootCore back-compat
```
kratsg: no need to comment

CMakeLists.txt
vpascuzz: Because the latter two libraries are used in >21.0. Followed what was done before with `xAODTriggerCnvLib`.

CMakeLists.txt
vpascuzz: Then the first comment (above) isn't needed. I just added to what was there, as you can see from the diff.

Root/MuonEfficiencyCorrector.cxx
vpascuzz: False. There is an `#ifdef` later in the function that makes use of `eventInfo` if not `USE_CMAKE`.

xAODAnaHelpers/BJetEfficiencyCorrector.h
vpascuzz: See above.

xAODAnaHelpers/JetSelector.h
vpascuzz: Followed suit (comment for 2.6 in `CMakeLists.txt`).

xAODAnaHelpers/MuonSelector.h
vpascuzz: See above.

CMakeLists.txt
kratsg: No, but we don't really need it. You're also defining it twice which doesn't make sense.

xAODAnaHelpers/BJetEfficiencyCorrector.h
kratsg: The `// RootCore back-compat` comment needs to be removed. We are planning to remove these lines later, and anything after the `#else` will screw up the removal.

xAODAnaHelpers/JetSelector.h
kratsg: The `// RootCore back-compat` comment needs to be removed. We are planning to remove these lines later, and anything after the `#else` will screw up the removal.

xAODAnaHelpers/MuonSelector.h
kratsg: The `// RootCore back-compat` comment needs to be removed. We are planning to remove these lines later, and anything after the `#else` will screw up the removal.

CMakeLists.txt
vpascuzz: (I wonder how the original duplicate `set` ended up merged then...)
Again, I was following what was already done so as to not change the workflow. Also not sure what you mean by "No, but we really don't need it."

xAODAnaHelpers/JetSelector.h
vpascuzz: Fine.

xAODAnaHelpers/MuonSelector.h
vpascuzz: Fine.

xAODAnaHelpers/BJetEfficiencyCorrector.h
vpascuzz: Fine.

CMakeLists.txt
kratsg: Yes, but the `set( release_libs )` is only for the differences between 2.6 and rel21. The ones your adding could've just been put into the list of dependencies rather than creating a new variable in CMake which needs to be unset.

vpascuzz: The `set`s are for backward-compatibility, as you had originally requested. So I don't know what you expect. The duplicated `set`s, I agree, can go.

kratsg: Backwards compatibility is for release 20 only.

vpascuzz: Sorry, I'm confused. Why is this done for `xAODTriggerCnvLib` then?
The `set( release_libs )` is for differences between 2.6 and rel21, yes. There is no `FTagAnalysisInterfacesLib` or `MuonAnalysisInterfacesLib` in 2.6, is there?

vpascuzz: `MuonAnalysisInterfacesLib` and `FTagAnalysisInterfacesLib` and were introduced in 21.2.5 and 21.2.6, respectively.

kratsg: > Sorry, I'm confused. Why is this done for xAODTriggerCnvLib then?
> The set( release_libs ) is for differences between 2.6 and rel21, yes. There is no FTagAnalysisInterfacesLib or MuonAnalysisInterfacesLib in 2.6, is there?

2.6.X isn't supported at all. Rel20 and Rel21 are the only things that are supported. The flag was left there from when we supported it before.

vpascuzz: So you suggest adding the two libraries to the depends list, even though they aren't present in rel20?

kratsg: > So you suggest adding the two libraries to the depends list, even though they aren't present in rel20?

CMakeLists.txt is only used in release 21. https://github.com/UCATLAS/xAODAnaHelpers/blob/master/cmt/Makefile.RootCore is used in release 20.

vpascuzz: Right. Thanks for the reminder...

kratsg:
@vpascuzz lemme know when you get these changes in.

vpascuzz:
Hopefully this settles it.

Thanks for the discussion, @kratsg.

vpascuzz:
#highfive

PR #1106, opened 2017-11-27
Fix xsec for private samples with existing DSIDs.

kkrizka:
This bug affects reading datasets from txt files (rucio lists are ok) that have an DSID with an entry in the xsec database. A default value of 1 was set for cross-section. This is then added by EventLoop to any value read from the SUSY cross-section database, if an entry there already exists.

The fix sets the cross-section information only if the corresponding .config file exists. It is up to the user to make sure that there is no conflict with the SUSY database. All official samples should have an entry there. The only time a .config file should be present is if a new dataset is being created and then an existing DSID should not be reused.

### DIFF
--- a/scripts/xAH_run.py
+++ b/scripts/xAH_run.py
@@ -362,14 +362,10 @@ if __name__ == "__main__":
                 if len(parts)!=2: continue
                 config[parts[0].strip()]=parts[1].strip()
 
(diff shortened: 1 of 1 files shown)


kratsg:
This could also be fixed through the python configuration using the existing `c.sample()`, no? See #1056 for the details here.

PR #1110, opened 2017-12-01
Small xAH_run Updates

kkrizka:
* Remove xAH_run xrd requirement from prun. See #1101 .
* Make --optGridNGBPerJob a sring (option MAX is valid).

### DIFF
--- a/scripts/xAH_run.py
+++ b/scripts/xAH_run.py
@@ -251,7 +251,7 @@ if __name__ == "__main__":
         raise OSError('Output directory {0:s} already exists. Either re-run with -f/--force, choose a different --submitDir, or rm -rf it yourself. Just deal with it, dang it.'.format(args.submit_dir))
 
     # they will need it to get it working for rel 20
-    needXRD = ( args.use_scanDQ2|args.use_scanRucio ) & args.driver!='prun'
+    needXRD = ( args.use_scanDQ2|args.use_scanRucio ) & (args.driver!='prun')
     if needXRD:
       if os.getenv('XRDSYS') is None and os.getenv('RUCIO_HOME') is None:
         raise EnvironmentError('xrootd client is not setup. Run localSetupFAX or equivalent.')
--- a/scripts/xAH_run.py
+++ b/scripts/xAH_run.py
@@ -175,7 +175,7 @@ prun.add_argument('--optGridDestSE',           metavar='', type=str, required=Fa
 prun.add_argument('--optGridSite',             metavar='', type=str, required=False, default=None)
 prun.add_argument('--optGridCloud',            metavar='', type=str, required=False, default=None)
 prun.add_argument('--optGridExcludedSite',     metavar='', type=str, required=False, default=None)
-prun.add_argument('--optGridNGBPerJob',        metavar='', type=int, required=False, default=2)
+prun.add_argument('--optGridNGBPerJob',        metavar='', type=str, required=False, default=2)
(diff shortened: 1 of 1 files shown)


ntadej:
@kkrizka, do you use optGridNGBPerJob? I'm having problems with it even when it's set as a string. Tasks randomly ignore this setting.

kkrizka:
I've started using it now to mitigate errors when the grid site decides to copy more inputs than it can fit in the work directory. I am having some success with it. Not with the "MAX" option though. I'm using a value of 10, which is some magical number that I found in the dist-analysis-help mailing list. It seems to work.

kratsg:
> I've started using it now to mitigate errors when the grid site decides to copy more inputs than it can fit in the work directory. I am having some success with it. Not with the "MAX" option though. I'm using a value of 10, which is some magical number that I found in the dist-analysis-help mailing list. It seems to work.

Should we set the default to 10 perhaps?

ntadej:
For some of my tasks it didn't work (set to 15), for others the value 15 also sounds fine. I have to do more investigation, what might be wrong that the flag didn't work, as the following resulting `prun` command looks fine to me:
```
prun --destSE=SIGNET_LOCALGROUPDISK --rootVer=6.08/06 --cmtConfig=x86_64-slc6-gcc62-opt --nGBPerJob=15 --mergeOutput --useAthenaPackages --exec="runjob.sh mc16_13TeV.410501.PowhegPythia8EvtGen_A14_ttbar_hdamp258p75_nonallhad.deriv.DAOD_EXOT12.e5458_s3126_r9364_r9315_p3263" --inDS=mc16_13TeV.410501.PowhegPythia8EvtGen_A14_ttbar_hdamp258p75_nonallhad.deriv.DAOD_EXOT12.e5458_s3126_r9364_r9315_p3263 --outDS=user.tadej.SeeSaw.EXOT12MCFakes.v0.fakes.410501.PowhegPythia8EvtGen_A14_ttbar_hdamp258p75_nonallhad.r1 --outputs=hist:hist-output.root,cutflow.root,metadata.root,duplicates.root,tree.root --writeInputToTxt=IN:input.txt --match="*" --mergeScript="elg_merge jobdef.root %OUT %IN" --inTarBall=jobcontents.tgz
```

ntadej:
@kkrizka, sorry for the noise. I also found one of your jobs that is not working: https://bigpanda.cern.ch/job?pandaid=3738485426

I already reported this in a thread in offline help about the reaching working directory size limit, I hope we get some feedback what could be wrong. Probably I should also make a new thread to DAST.

kkrizka:
Yes, I think xAH is doing everything correctly. The problem (unsurprisingly) seems to be with panda.

I've just looked at some of my straggling jobs and nGBPerJob does not actually seem to be respected. Consider this task:
https://bigpanda.cern.ch/task/12734787/
You can see `--nGBPerJob=10` in cliParams. Now look at one of the jobs it contains:
https://bigpanda.cern.ch/job?pandaid=3738198951
That's definitely more than 10 inputs, each roughly 1Gb in size, defined!

I got your reply as I was typing this. My plan was to also ask about this behaviour on the forums. I'll tack on to your thread to make it clear several of us are seeing this problem.

ntadej:
Maybe just add DAST mailing list to thread?

kkrizka:
Ah, I misread your message. I thought it was already on DAST. Yes, I'll add them to it.

PR #1115, opened 2017-12-18
Update CMakeList & xAH_run to allow for AnalysisTop

jdandoy:
Our CMakeList & xAH_run.py files currently require AnalysisBase environmental variables to be set, excluding the possibility to run in other ASG analysis frameworks.  These changes would allow AnalysisTop to be used (perhaps others can be added in the future if needed). 

This also swaps the environmental variable ${AnalysisBase_VERSION} for the cmake variable AnalysisBase_VERSION in CMakeList.txt, which seems cleaner.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,12 +6,19 @@
 atlas_subdir( xAODAnaHelpers )
 
 set( release_libs xAODTriggerCnvLib )
-if( ${AnalysisBase_VERSION} VERSION_GREATER 21.0 )
+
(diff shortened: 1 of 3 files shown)


jdandoy:
I'm not sure how much is gained by this, but I've made the change.  I'd prefer to keep the list of acceptable ASG analysis types in xAH_run rather than burying it in utility scripts.

jdandoy:
One sec, need to fix the error thrown

PR #1116, opened 2017-12-24
Update tool interfaces to be compatible with 21.2.13

kkrizka:
Updates to the tool interfaces to make them compatible with the latest AnalysisBase release:
* `ElectronSelection` tool only wants pointers to electrons, not references.
* Trig::MatchingTool has a new non-default option to [allow rerun decision](https://gitlab.cern.ch/atlas/athena/merge_requests/6515). This does have a default value if you use `Trig::IMatchingTool` and `asg::AnaToolHandle`. This updates `TrigMatcher` to use the `asg::AnaToolHandle` interface.

I am marking this as a WIP, because 21.2.13 does not like the way that we name our tools in BasicEventSelection: `<ToolType>/<ToolType>_<AlgorithmName>::<AlgorithmAddress>`. More speficially, the GRL tool hangs on retrieve if `::<AlgorithmAddress>` is included in the name. I reported it here:
https://groups.cern.ch/group/hn-atlas-PATHelp/Lists/Archive/Flat.aspx?RootFolder=%2fgroup%2fhn-atlas-PATHelp%2fLists%2fArchive%2fAnaToolHandleIGoodRunsListSelectionToolretrieve%20hangs%20in%2021%2e2%2e13&FolderCTID=0x0120020084D7E80CB0C3394A8D54EF07C309B044

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   matrix:
     - ABV_CM=21.2.7 ALLOWFAIL=true
     - ABV_CM=21.2.10 ALLOWFAIL=false
+    - ABV_CM=21.2.13 ALLOWFAIL=false
 script:
   - envsubst '\$ABV_CM \$TRAVIS_COMMIT \$TRAVIS_JOB_ID' < ci/Dockerfile > Dockerfile
   - docker build --build-arg ABV_CM=${ABV_CM}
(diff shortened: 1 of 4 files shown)


kratsg:
Yeah, so this isn't too terrible to fix. We can change the delimiter here (http://xaodanahelpers.readthedocs.io/en/master/xAHAlgorithm.html?highlight=setToolName#_CPPv2NK3xAH9Algorithm11setToolNameERN3asg13AnaToolHandleI1TEENSt6stringE) and that won't break any compatibility.

kkrizka:
I think we should just switch to using private tools. Should work with `EL::Algorithm` in the latest release. Release 20.7 compatibility might been another `USE_CMAKE`.

I will prepare another MR with this changes, since it will be less trivial.

kkrizka:
@kratsg , travis seems to be having trouble unpacking uncrustify. Is there some way to fix it, or bypass it?

See: https://travis-ci.org/UCATLAS/xAODAnaHelpers/jobs/321136689

PR #1117, opened 2017-12-25
Use Private Tools In Algorithms

kkrizka:
Uses the "private tool" functionality of the `asg::AnaToolHandle` to make unique tools per algorithm. This replaces adding the algorithm address to the end of the tool name, which seems to be broken starting in at least 21.1.11.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -36,10 +36,8 @@ using HelperClasses::ToolName;
 ClassImp(BJetEfficiencyCorrector)
 
 
-BJetEfficiencyCorrector :: BJetEfficiencyCorrector ()
-: Algorithm("BJetEfficiencyCorrector"),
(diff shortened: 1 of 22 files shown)


kkrizka:
Should close #1117 .

kratsg:
Whoa, hold on, TDT is special and needs to have the configurable name. It can't be a private tool.

kkrizka:
It's not a private tool.

Why does it need to have a configurable name, as long as it is the same
everywhere? I set it to "TrigDecisionTool", as used by the atlassoftwaredocs
and all other pieces of code I've seen that use it.

kratsg:
We have some folks who use it with CxAODFramework (they don't use the default name). The way we had it before with a configurable (but default set) was what worked. I suspect we might have a complaint or two from someone who gets hit by this again (multiple instantiations of TDT) and it'll end up being that they're using a piece of code that isn't up to date..

PR #1125, opened 2018-01-04
Add analysis top

kratsg:


kratsg:
@jdandoy will appreciate this, as it'll build both AnalysisBase and AnalysisTop docker images -- as well as test multiple releases in both simultaneously.

jdandoy:
Awesome!

PR #1130, opened 2018-01-10
METConstructor tools had no names

### LINKED ISSUE #1128: Error with OR and METConstructor
Hi all (@kratsg ),
while developing a configuration for taus I stumbled in a weird issue. If I run the METConstructor before the OR tool I get the following error

`asg::ToolStore::get WARNING Tool with name "" not found worker finished: Real time 0:00:16, CP time 13.720 [2018-01-10 05:05:22,042][ERROR ] -------------------- An exception was caught! (xAH_run.py:671) Traceback (most recent call last): File "/home/fscutti/workdir/build/x86_64-slc6-gcc62-opt/bin/xAH_run.py", line 653, in <module> driver.submit(job, args.submit_dir) Exception: void EL::Driver::submit(const EL::Job& job, const string& location) => Couldn't find tool with name "" (C++ exception of type runtime_error)`

If I run one tool at a time or reverse the order all seems to be fine. 

My configuration is [here](https://www.dropbox.com/s/80sttx1dva553go/config.py?dl=0)
I am running with 21.2.12 on a SUSY3 DxAOD. At first I thought it was an issue of the ORToolbox
initialisation using empty strings at the init stage but all seems to be fine there. 

Have you stumbled in something like this before?

Cheers,
Federico

kratsg:
This closes #1128.

### DIFF
--- a/Root/METConstructor.cxx
+++ b/Root/METConstructor.cxx
@@ -138,17 +138,15 @@ EL::StatusCode METConstructor :: initialize ()
   ANA_MSG_DEBUG( "Is MC? " << isMC() );
 
   //////////// IMETMaker ////////////////
-  ASG_SET_ANA_TOOL_TYPE(m_metmaker_handle, met::METMaker);
-  m_metmaker_handle.setName("METMaker");
   if ( m_dofJVTCut ) {
     ANA_CHECK(m_metmaker_handle.setProperty("JetRejectionDec", "passFJVT"));
   }
-  m_metmaker_handle.retrieve();
+  ANA_MSG_DEBUG("Retrieved tool: " << m_metmaker_handle);
+  ANA_CHECK(m_metmaker_handle.retrieve());
 
   ///////////// IMETSystematicsTool ///////////////////
-  ASG_SET_ANA_TOOL_TYPE(m_metSyst_handle, met::METSystematicsTool);
-  m_metSyst_handle.setName("METSyst");
-  m_metSyst_handle.retrieve();
+  ANA_MSG_DEBUG("Retrieved tool: " << m_metSyst_handle);
+  ANA_CHECK(m_metSyst_handle.retrieve());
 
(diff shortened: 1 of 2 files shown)


ntadej:
@kratsg, is this just a "friendly" name or is it related to the actual tool?

Should it be unique? We have two instances of `CP::IJetJvtEfficiency` in `JetSelector.h` and I'm wondering if it will still work as intended.

kratsg:
> Should it be unique? We have two instances of CP::IJetJvtEfficiency in JetSelector.h and I'm wondering if it will still work as intended.

Read the documentation here: https://xaodanahelpers.readthedocs.io/en/master/Development.html#adding-and-initializing-tools 

We're - by default - using private tools where possible. Because of this, we don't need to worry about name conflicts. If two tools have the same type and name, then it's the same tool used for both. If you need to share tools, this is the case.

ntadej:
Thanks, then `JetSelector` needs fixing. I will take care of it.

ntadej:
I'm now getting
```
AsgTools_AnaToolHandle   ERROR   /build1/atnight/localbuilds/nightlies/21.2/athena/Control/AthToolSupport/AsgTools/Root/AnaToolHandle.cxx:148 (StatusCode asg::detail::makeToolRootCore(const string&, const string&, asg::AsgTool*&)):name "met.met::METMaker" does not match format expression
```

kratsg:
I'm seeing this sort of error with SUSYTools as well -- I'm wondering if there's a change in AnaToolHandle that breaks this sort of thing.

kratsg:
Sent an email to pathelp -- tool names can't contain a ":". I'll fix this I suppose...

kratsg:
Fixed here: https://gitlab.cern.ch/atlas/athena/merge_requests/7891/diffs#note_516003 but we'll fix it now in xAH.

PR #1132, opened 2018-01-11
Add some cross-sections

kkrizka:
* High pT filtered VH->bb samples
* Low mass Z'->qq + jet samples
* Z'->qq + gamma samples with lower photon pT filter


kratsg:
Just wondering if we should switch to the ST crosssection database? It's in GroupData/dev

kkrizka:
Sure, but this will do for now.

Issue #837, opened 2017-03-11
xAH and submission of multiple datasets


urania277:
Not having modified xAH_run yet (and unfortunately not having that much time to do it myself), I'm looking to see if anyone would be interested to implement this: 

https://twiki.cern.ch/twiki/bin/view/AtlasProtected/EventLoop#Processing_multiple_datasets_in

Thanks!
Caterina

kratsg:
This would be somewhat easily implemented in `xAH_run.py` but I guess it's a question of how we support this option since it's a very-very-very specific under `prun`. @kkrizka might have an idea.

kkrizka:
How about adding an option to prun called `--inputSingleTask name`? Then the `SampleHandler` part of `xAH_run.py` (https://github.com/UCATLAS/xAODAnaHelpers/blob/83a5669d2385cdc0e4947409612001167a3188fb/scripts/xAH_run.py#L315) will use the `SH::SampleGrid(name)` stuff instead of `scanRucio`.

This should work quite nicely for data at least. I have to think a bit how to deal with splitting the MC. We can either automatically create a separate output DS for each input DS, or add another option to xAH_run.py to enable the behaviour. @urania277, do you have a preference?

I will experiment with this tomorrow.

kkrizka:
I put an example implementation at https://github.com/kkrizka/xAODAnaHelpers/tree/singlejeditask . It crates a list of datasets per input filelist when the `--singleTask` option is added to the prun command.

It almost works, except the first list of datasets is tries to submit crashes with the following.

`Traceback (most recent call last):
  File "/cvmfs/atlas.cern.ch/repo/ATLASLocalRootBase/x86_64/pyAmi/pyAMI-5.0.3.2/lib/argparse/__init__.py", line 1194, in __eq__
    return vars(self) == vars(other)
TypeError: vars() argument must have __dict__ attribute
TypeError: an integer is required`

All subsequent lists are submitted correctly. More debugging is still needed.

beojan:
Has there been any progress on this? The lack of period containers for many (all?) derivations in release 21 is making the one-task-per-dataset workflow quite painful, due to retries, and worse, resubmits if the pilot jobs for a task spuriously fail.

ntadej:
Period containers should be available for all data periods after the re-derivation completes. Note that you usually have to request them by your derivation contacts.

kratsg:
[...]

PR #1140, opened 2018-01-21
xAH and submission of multiple datasets

kkrizka:
Implements submitting several datasets to the grid in a single JEDI task. Should address #837.
https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/EventLoop#Processing_multiple_datasets_in

If you use `--singleTask` with `--inputList` option, the datasets in each filelist are grouped into one sample. The name of the input filelist is used as the sample name. Also `--singleTask` only works with rucio.

There are two hacks in this commit. 
* Adding `SampleGrid` does not seem to work in Python and throws the error below. Unclear why. I added a `xAH::addRucio` function that implements creating and adding the `SampleGrid` to `SampleHandler`.
> TypeError: vars() argument must have __dict__ attribute TypeError: an integer is required
* Adding `addRucio` to `HelperFunctions` does not seem to generate the dictionary correctly. Also unclear why. I added it to the `xAH` namespace instead.

Also removes `--inputDQ2`, since DQ2 does not exist anymore.

### DIFF
--- a/Root/HelperFunctions.cxx
+++ b/Root/HelperFunctions.cxx
@@ -3,6 +3,11 @@
 
 #include "xAODBase/IParticleContainer.h"
 
+// samples
+#include <SampleHandler/SampleGrid.h>
+#include <SampleHandler/MetaFields.h>
+#include <SampleHandler/MetaObject.h>
+
 // jet reclustering
 #include <fastjet/PseudoJet.hh>
 #include <fastjet/ClusterSequence.hh>
@@ -11,6 +16,13 @@
 #include <fastjet/tools/Filter.hh>
 #include <JetEDM/JetConstituentFiller.h>
 
+void xAH::addRucio(SH::SampleHandler& sh, const std::string& name, const std::string& dslist)
+{
+  std::unique_ptr<SH::SampleGrid> sample(new SH::SampleGrid(name));
+  sample->meta()->setString(SH::MetaFields::gridName, dslist);
+  sample->meta()->setString(SH::MetaFields::gridFilter, SH::MetaFields::gridFilter_default);
+  sh.add(sample.release());
+}
 
 MsgStream& HelperFunctions::msg( MSG::Level lvl ) {
   static MsgStream msgStream( "HelperFunctions" );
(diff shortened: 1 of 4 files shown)


kratsg:
Ok, I understand the changes being made here. The main question I have is what the use case for mapping N files in a file list to 1 sample is? Most people prefer to be able to have an output for each input and then merge them manually after some post-processing. I suspect that this isn't currently possible?

The hack is also curious and should be reported. I'm not 100% in favor of adding C++ code to make python code work... [especially because we'll have to go back and fix it when the python code eventually works again?]

kkrizka:
My original plan was to use this to manage the submission of large number of datasets (ie: with mc16c, the number of datasets doubles...). The hope was that the `--addNthFieldOfInDSToLFN` prun option would also add the suffix to the output dataset for a file in addition to the output file name. For example, [user.kkrizka.Pythia8_dijet.fatjet.20180116-test-01_tree.root](https://bigpanda.cern.ch/datasetList/?containername=user.kkrizka.Pythia8_dijet.fatjet.20180116-test-01_tree.root/&jeditaskid=13022556) would become [user.kkrizka.Pythia8_dijet.fatjet.20180116-test-01_mc16_13TeV.361020.Pythia8EvtGen_A14NNPDF23LO_jetjet_JZ0W.tree.root](https://bigpanda.cern.ch/fileList/?datasetid=171528438). Sadly that is not the case, so I don't have a good use case for it anymore.

Unless @urania277 thinks that this meets her request in #837, I think I will remove this PR.

As for the bug, I've notified PATHelp a while ago.

beojan:
As I stated in #837, this would be rather helpful to deal with missing period containers.

kratsg:
> As I stated in #837, this would be rather helpful to deal with missing period containers.

Hi @beojan, you're welcome to try and get this working. This is in the `singleTask` branch in xAH so you can fork and submit a pull request if you get it working. I don't know if @kkrizka has time to work on this.

ntadej:
Also, at this point, you should probably just request the period containers. They are just not created automatically.

beojan:
I have, but @alexandertuna is waiting for the reprocessing to finish (EXOT8 didn't get reprocessed with the other derivations).

beojan:
It seems to work just fine for my use case.

kratsg:
@beojan if it works for you, we will merge it in.

kkrizka:
If this way of combining tasks is OK, then this should be ready to merge.

PR #1141, opened 2018-01-21
Cleanup MuonEfficiencyCorrector of per-year tools

ntadej:
As we no longer support release 20.7, per-year trigger eff. tools are no longer necessary.

### DIFF
--- a/Root/MuonEfficiencyCorrector.cxx
+++ b/Root/MuonEfficiencyCorrector.cxx
@@ -214,45 +214,27 @@ EL::StatusCode MuonEfficiencyCorrector :: initialize ()
 
   // 3.
   // Initialise the CP::MuonTriggerScaleFactors tool
-  // several years can be configured
   //
 
-  m_YearsList.push_back("all");
+  m_trigEffSF_tool_name = "MuonTriggerScaleFactors_effSF_Trig_Reco" + m_WorkingPointRecoTrig + "_Iso" + m_WorkingPointIsoTrig;
 
-  // Initialize vector of names
-  //
-  for(auto yr : m_YearsList) {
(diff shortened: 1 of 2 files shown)


jdandoy:
Hey @ntadej,

Can you clarify a bit why this is no longer necessary?  Looking at the [muon trigger efficiency twiki](https://twiki.cern.ch/twiki/bin/view/Atlas/MuonTriggerPhysicsRecommendationsRel212017#Supported_triggers), it seems the recommendation is still `HLT_mu20_iloose_L1MU15_OR_HLT_mu50` for 2015 but `HLT_mu26_ivarmedium_OR_HLT_mu50` for 2016+2017.  Can we cover all data years in 1 go?

Thanks,

Jeff

lschaefer:
Hi @jdandoy, @ntadej, 
So my understanding is: the muon trigger SF tool has been updated to accommodate all years at once, but even so we should only pass it the triggers corresponding to the year from the random run number. So a single mc16a sample requires both 2015 and 2016 triggers, but individual events require one or the other. So in my local version of xAH, I added a new property to MuonEfficiencyCorrector, `m_MuTrigLegs2015` which has `HLT_mu20_iloose_L1MU15_OR_HLT_mu50` (while the standard `m_MuTrigLegs` has `HLT_mu26_ivarmedium_OR_HLT_mu50`) and then xAH decides whether to pass the 2015 or 2016+ 2017 triggers to the SF tool. Does this seem like a reasonable way to proceed? If so I can create a new PR after some further validation. 
Best,
Leigh

ntadej:
Hi @jdandoy, @lschaefer,

Since release 21, muon trigger efficiency tool does all year-dependent logic internally. When a trigger is not supported, you will get a zero SF (when `m_AllowZeroSF` is True). If you want to avoid it, we can re-introduce some protection.

Cheers, Tadej

lschaefer:
Thanks!

PR #1143, opened 2018-01-27
Make muon quality inclusive (closes #995)

### LINKED ISSUE #995: Muon quality inconsistency
Muon quality is written out "strictly" - Medium quality muon is marked only as Medium and not also as Loose and VeryLoose. This is not consistent with electrons where all branches are always inclusive.

I can fix muon selector but this is a breaking change so maybe it needs to be discussed.

ntadej:
`xAOD::Muon::Tight` is the lowest enum. Always compare with `<=`.


ntadej:
Travis is failing due to stalling.

kratsg:
restarted. problem is with docker hub.

Issue #1109, opened 2017-12-01
Ntuples size reduction proposal


ntadej:
To reduce output ntuple size it would be useful not to output:
 - [x] muon isolation status for unused WPs
 - [x] electron isolation status for unused WPs
 - [x] muon quality status for unused WPs
 - [x] electron ID status for unused WPs
 - [x] particle counts - Is there a use case for those? Usually they are wrong if you do extra cuts on ntuple processing and you can always check a size of vectors in branches.

Most of upper branches contain -1 anyway if not used so it's useless to store them. I propose to propagate some of the settings to final tree algo to avoid complicated output configuration. What do you think?

I can go forward with implementing those changes but they might be breaking (especially particle counts). Let me know if this sounds rights so I can proceed.

kratsg:
How do you check if it's not needed? As it currently is, with my massive changes to how Electrons are handled -- you can specify the individual isolation that you want.

ntadej:
For the SFs, yes but not for "status" branches (el_isIsolated, muon_isMedium, ...). I counted at least 25 branches that I do not need and just take up space. When doing systematics this can account to quite a lot of size I guess.

Let's look at the electron isolation example:

All `el_isIsolated_*` branches are still present all the time. In ElectronSelector you define `m_IsoWPList` of isolations you want to be decorated. This is usually one or two WPs. Others that get filled with -1s just take space.

I realised now that just storing this vector to TStore will not work as branches need to be connected before filling the variables, am I correct? Is there another way to propagate algorithms configuration to the final TreeAlgo?

Another option would be to just use your code also for those branches. Then you would always need SFs and "status branches".

kratsg:
I suspect it would be nicer to define the status in the same way as we do for SFs.

ntadej:
Ok, I will come up with a proposal. Thanks for the feedback.

jbossios:
noMultiplicity flag added to remove particle count: https://github.com/UCATLAS/xAODAnaHelpers/pull/1500

Closing since everything seems to have been implemented

PR #1144, opened 2018-01-27
Add configurable electron PID/isolation branches

ntadej:
This is the first part of #1109 implementation.

What was changed:
 - added `PID_<wp>` switch token and renamed the existing one to `PIDSF_<wp>` as scale factors do not use all PID working points
 - used existing `ISOL_<wp>` for isolation branches
 - isolation counters were removed

This is breaking: branches names stayed the same but isolation ones will be removed. PID can get the same defaults as they were before. If you want also `ISOL` can be split into `ISOL` and `ISOLSF`, but I don't think it's necessary - one would need SFs anyway.

I plan to do the same thing for muons if this is reasonable.

### DIFF
--- a/Root/ElectronContainer.cxx
+++ b/Root/ElectronContainer.cxx
@@ -21,16 +21,7 @@ ElectronContainer::ElectronContainer(const std::string& name, const std::string&
   }
 
   if ( m_infoSwitch.m_isolation ) {
-    m_isIsolated_LooseTrackOnly              = new std::vector<int>   ();
-    m_isIsolated_Loose                       = new std::vector<int>   ();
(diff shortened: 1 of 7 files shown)


ntadej:
@kratsg, this has been fully implemented. The defaults can be updated to retain the same branches as before. Please let me know, what you think about this change.

kratsg:
Please add more documentation on this, and then I'll be happy.

ntadej:
Ok, can you please point me to the best location for this, as I can not find the part about info switches (I can create one if needed).

kratsg:
Header file `HelperClasses.h` contains the info about info switches.

PR #1146, opened 2018-01-31
Implement PRW autoconfiguration (stolen from SUSYTools)

kkrizka:
Implements automatic association of PRW config files with the PRW tools using code copied from SUSYTools. Set `m_autoconfigPRW` to use.

* The campaign is automatically determined from the tags in the dataset name
* The campaign can be overridden by setting `m_mcCampaign` to a comma separated list of campaigns (ie: if one wants to treat mc16a+mc16c as a super-dataset).

I am marking this WIP since I am still testing it.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -30,6 +30,8 @@
 #include "xAODCore/tools/IOStats.h"
 #include "xAODCore/tools/ReadStats.h"
 
+#include "xAODMetaData/FileMetaData.h"
+
 
 // this is needed to distribute the algorithm to the workers
 ClassImp(BasicEventSelection)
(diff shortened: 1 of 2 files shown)


Root/BasicEventSelection.cxx
```diff
@@ -950,6 +982,115 @@ EL::StatusCode BasicEventSelection :: execute ()
   return EL::StatusCode::SUCCESS;
 }
 
+// "Borrowed" from SUSYTools
```
kratsg: Can you add a link to where you got this from (the GitLab location)?

kkrizka: done

kkrizka:
@kratsg, what are the protections on master? I did `git push`, expecting my master (with some of the changes in this PR) to fail with some errors. However it did not and master is now updated..

kratsg:
> @kratsg, what are the protections on master? I did git push, expecting my master (with some of the changes in this PR) to fail with some errors. However it did not and master is now updated..

screenshot 2018-01-31 08 15 58

Unclear, perhaps you're an administrator. I can fix this with a force push to remove the top commit.

kratsg:
Fixed.

kratsg:
@kkrizka is this still a WIP or ready to go?

kkrizka:
I think it should be ready to go. Still checking some things, but at least the old behavior shouldn't be broken.

PR #1148, opened 2018-02-01
Do not hardcode PRW scale factors

ntadej:
Do not hardcode PRW scale factors. These are defined in the tool and currently there is no reason to edit them by hand.


ntadej:
Travis stalling on docker step again :)

kratsg:
> Travis stalling on docker step again :)

restarted build. You should also have access to restart the builds as well on travis, I think, if you log-in with your github account.

ntadej:
I don't think I have them. I guess I need to be a project member to be able to do it (I tried it before and I didn't see any button).

kratsg:
I see, it's probably admin-only for right now.

fscutti:
Hi @kratsg, can this be merged? It seems pretty urgent and I'm not able to merge myself.

PR #1150, opened 2018-02-02
Add ability to disable counter branches

ntadej:
As some of us do not rely on the counter branches, it would be nice to add ability to remove them using `noCounter` switch.

### DIFF
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -142,6 +142,7 @@ namespace HelperClasses{
       }
 
     m_useTheS   = has_exact("useTheS");
+    m_noCounter = has_exact("noCounter");
   }
 
   void MuonInfoSwitch::initialize(){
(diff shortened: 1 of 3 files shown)


kkrizka:
Out of curiosity, what exactly do you gain by removing this branch?

ntadej:
I gain some size, it might not be much, but you remove it ~40 times when doing systematics.

kratsg:
I'm ok with this if @kkrizka is fine with it. But this is a "yet another configuration" which may or may not be useful...

kkrizka:
I would agree with @kratsg here about this being "yet another configuration" which may or may not be useful".

kkrizka:
@ntadej , do you have some statistics on how much you save on size when running with systematics?

ntadej:
@kkrizka, I will try to do them. Until then, feel free to leave this open (I understand your opinion).

ntadej:
I'm closing this for now. Will reopen if we will find it important again.

PR #1152, opened 2018-02-05
Trigger matching for jets

fscutti:
This PR introduces trigger matching for jets and corrects the trigger match setting for taus. 
No DR info is provided for jets or for taus as opposed to say muons and electrons. The class JetTriggerInfoSwitch has been removed as all jet info has been transferred to JetInfoSwitch, like is done for leptons

### DIFF
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -230,7 +230,6 @@ namespace HelperClasses{
   void JetInfoSwitch::initialize(){
     std::string tmpConfigStr; // temporary config string used to extract multiple values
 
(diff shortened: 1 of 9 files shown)


Root/HelperClasses.cxx
```diff
@@ -235,6 +235,8 @@ namespace HelperClasses{
   void JetInfoSwitch::initialize(){
     std::string tmpConfigStr; // temporary config string used to extract multiple values
 
+    m_kinematic     = has_exact("kinematic");
```
ntadej: This is inherited from ParticleInfoSwitch so it should not be added here.

PR #1156, opened 2018-02-08
Allow multiple trigger jet collections in treeAlgo and sort L1 RoIs

kalderon:
1. Based on the logic of truthJets, extend the triggerJets sections to allow multiple trigger jet collections to be run at once, each with its own detailStr.
2. Add an option to sort L1  jet RoIs by et8x8, since they are currently unsorted, leave default false so as to leave default behaviour unchanged

This should only be three commits, but ended up as 6 since I made a tiny extra one (ac4c701) and subsequently tried to squash it onto the former, which obviously didn't work...

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -542,15 +542,34 @@ void HelpTreeBase::AddL1Jets()
 
 }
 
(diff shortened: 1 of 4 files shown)


kratsg:
Don't worry about not squashing commits. We do this for you as part of merging the PR.

PR #1159, opened 2018-02-09
Protect from zero random run number

jdandoy:
It is possible for the random run number from the pileup tool to be zero, and reflects cases where the pileup conditions of that MC event do not match any in data.  Currently the default was Data16, which does not work if you're running on Data17 only, causing a crash when the non-existent instance of m_muonCalibrationAndSmearingTools[randYear] is accessed.  I've set the default randYear to be the first instance in m_YearsList, ensuring it exists.

Also some nearby retabbing.

### DIFF
--- a/Root/MuonCalibrator.cxx
+++ b/Root/MuonCalibrator.cxx
@@ -252,10 +252,8 @@ EL::StatusCode MuonCalibrator :: execute ()
   const xAOD::EventInfo* eventInfo(nullptr);
   ANA_CHECK( HelperFunctions::retrieve(eventInfo, m_eventInfoContainerName, m_event, m_store, msg()) );
 
-  std::string randYear = "Data16";
-
-  // a default 2016 run
-  int runNumber = 296939;
+  std::string randYear = m_YearsList.at(0);
+  int runNumber = 0;
 
   if ( !isMC() && m_forceDataCalib ) {
     runNumber = eventInfo->runNumber();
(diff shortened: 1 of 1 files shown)


Root/MuonCalibrator.cxx
```diff
@@ -339,16 +337,15 @@ EL::StatusCode MuonCalibrator :: execute ()
...
 
-	ANA_MSG_DEBUG( "  uncailbrated muon " << idx << ", pt = " << muSC_itr->pt()*1e-3 << " GeV");
-	if(muSC_itr-> primaryTrackParticle()){
-	  if ( m_muonCalibrationAndSmearingTools[randYear]->applyCorrection(*muSC_itr) == CP::CorrectionCode::Error ) {  // Can have CorrectionCode values of Ok, OutOfValidityRange, or Error. Here only checking for Error.
-	    ANA_MSG_WARNING( "MuonCalibrationAndSmearingTool returned Error CorrectionCode");		  // If OutOfValidityRange is returned no modification is made and the original muon values are taken.
-	  }
-	}
-
-	ANA_MSG_DEBUG( "  corrected muon pt = " << muSC_itr->pt()*1e-3 << " GeV");
+	      ANA_MSG_DEBUG( "  uncailbrated muon " << idx << ", pt = " << muSC_itr->pt()*1e-3 << " GeV");
```
kratsg: "cailbrated" -> "calibrated"

kratsg: Since this is inside `execute()`, we should use `ANA_MSG_VERBOSE` for this level, rather than DEBUG.

Root/MuonCalibrator.cxx
```diff
@@ -339,16 +337,15 @@ EL::StatusCode MuonCalibrator :: execute ()
...
+	      if(muSC_itr-> primaryTrackParticle()){
+	        if ( m_muonCalibrationAndSmearingTools[randYear]->applyCorrection(*muSC_itr) == CP::CorrectionCode::Error ) {  // Can have CorrectionCode values of Ok, OutOfValidityRange, or Error. Here only checking for Error.
+	          ANA_MSG_WARNING( "MuonCalibrationAndSmearingTool returned Error CorrectionCode");		  // If OutOfValidityRange is returned no modification is made and the original muon values are taken.
```
kratsg: Is this an error that should stop execution or is it ok if there is an error and continue execution?

Root/MuonCalibrator.cxx
```diff
@@ -339,16 +337,15 @@ EL::StatusCode MuonCalibrator :: execute ()
...
+	        }
+	      }
 
-	++idx;
+	      ANA_MSG_DEBUG( "  corrected muon pt = " << muSC_itr->pt()*1e-3 << " GeV");
```
kratsg: `ANA_MSG_VERBOSE` here

ntadej:
How often do you get invalid run number returned? This sounds like a bug and should probably reported.

jdandoy:
Hi @ntadej ,

It is very infrequent, and I believe this is to be expected on a very limited basis.  If you have an MC event simulated with a <mu> of 2, but there are no data events that had a <mu> of 2, then the pileup weight should be zero, and I believe the tool appropriately returns a random run number of zero.  I asked about this during MC15 and it was expected.  MC16c does not perfectly match 2017,  so there are some cases like this.

ntadej:
Hi @jdandoy, thank you for the explanation. So if the weight is zero, we do not really care for which year the event is calibrated, just that it does not crash?

jdandoy:
Exactly, when pileup reweighted the event will have zero weight anyway.  But we don't reject the event earlier (still useful perhaps for people not pileup-reweighting), so we still need to make sure an available muon calibrator tool is selected.  Before it defaulted to 2016, which does not work when running only on 2017, so now it just defaults to the first year provided by the user.  The bias here should be very small.

PR #1160, opened 2018-02-11
Add support for multiple data unprescaling weights

ntadej:
This add support for multiple data unprescaling weights. This is useful if you have same ntuples for different studies (different trigger selection).

Personally I would remove the old implementation which takes an OR of all triggers in `m_triggerSelection` but this is breaking. Let me know how to proceed.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -303,6 +303,7 @@ EL::StatusCode BasicEventSelection :: initialize ()
     ANA_MSG_INFO( "Truth only! Turn off trigger stuff");
     m_triggerSelection      = "";
     m_extraTriggerSelection = "";
+    m_triggerUnprescale     = "";
     m_applyTriggerCut = m_storeTrigDecisions = m_storePassL1 = m_storePassHLT = m_storeTrigKeys = false;
     ANA_MSG_INFO( "Truth only! Turn off GRL");
     m_applyGRLCut = false;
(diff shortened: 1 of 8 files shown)


kratsg:
I'm ok with breaking backwards compatibility here as long as it's documented. THis is something hat needs to be cleaned up.

ntadej:
I did some changes:
 - prescales info is now optional in the ntuples
 - you have prescale info from trigger DB for every trigger and lumi-based prescale from PRW for the ones you choose
 - set in trigger info switch as it is more suitable there
 - disabled for MC as MC does not contain prescaled triggers
 - added description to info switch to know which switch corresponds to which weights

ntadej:
@kratsg, I got a comment from @fscutti that we should unify how we store prescales. Prescales from the trigger tool are now stored in a vector with a separate vector of strings acting as the key for the contents. I stored the lumi-based prescales in a `map<string, float>`.

Should this be unified and how? Note that lumi-based prescales will not necessarily be for the same triggers.

kratsg:
> @kratsg, I got a comment from @fscutti that we should unify how we store prescales. Prescales from the trigger tool are now stored in a vector with a separate vector of strings acting as the key for the contents. I stored the lumi-based prescales in a map<string, float>.
> 
> Should this be unified and how? Note that lumi-based prescales will not necessarily be for the same triggers.

It should be unified. Might consider thinking about creating a typedef for `prescales` so that we don't necessarily create a new special object for this, but make it easier to keep track of that particular special object in the code? (Move this to an issue).

Issue #1161, opened 2018-02-12
Automatic PRW file setup for mc16c failing on old p-tags (mc16a)


kpachal:
Hi folks,

A recent update from @kkrizka awesomely created a way to automatically set up the right PRW file for mc16a vs mc16c using the information from xAOD::FileMetaData::amiTag. Sadly this value only exists in my mc16c files (*s3126_r9781_r9778_p3340) and not in my mc16a files (*s3126_r9364_r9315_p3309). Guessing it was added with the later p-tag?

Anyway, running on an mc16a file with the above tags gives the following error:

`BasicEventSelection.Ba...INFO    Adding Pileup files for CP::PileupReweightingTool:
BasicEventSelection.Ba...INFO    Adding LumiCalc files for CP::PileupReweightingTool:
	 GoodRunsLists/data15_13TeV/20170619/PHYS_StandardGRL_All_Good_25ns_276262-284484_OflLumi-13TeV-008.root
	 GoodRunsLists/data16_13TeV/20170605/PHYS_StandardGRL_All_Good_25ns_ignore_TOROID_STATUS_297730-311481_OflLumi-13TeV-008.root
	 GoodRunsLists/data17_13TeV/20170901/physics_25ns_Triggerno17e33prim.lumicalc.OflLumi-13TeV-001.root
PathResolver             WARNING Locating dev directory dev/SUSYTools/PRW_AUTOCONGIF/files/. Do not let this propagate to a release
BasicEventSelection.Ba...ERROR   /afs/cern.ch/work/k/kpachal/DijetISR/Resolved2017/CodeBase/DijetISRCode_Organizer/xAODAnaHelpers/Root/BasicEventSelection.cxx:1054 (StatusCode BasicEventSelection::autoconfigurePileupRWTool()): autoconfigurePileupRWTool(): access to FileMetaData failed, but don't panic. You can try to manually set the 'mcCampaign' BasicEventSelection property to 'mc16a', 'mc16c' or 'mc16d' and restart your job. If you set it to any other string, you will still incur in this error.
BasicEventSelection.Ba...ERROR   /afs/cern.ch/work/k/kpachal/DijetISR/Resolved2017/CodeBase/DijetISRCode_Organizer/xAODAnaHelpers/Root/BasicEventSelection.cxx:525 (virtual EL::StatusCode BasicEventSelection::initialize()): Failed to call "autoconfigurePileupRWTool()"
[...]

PR #1162, opened 2018-02-13
Altered mc campaign parsing & added lumicalc automatic PRW setup

kpachal:
Hi,

This addresses the issue discussed with @kratsg in [1161](https://github.com/UCATLAS/xAODAnaHelpers/issues/1161). It does two things:

1. Alters the way that the mc campaign is being picked up, taking it from the run number in the MC rather than from the metadata since that quantity does not exist in older p-tags (like mc16a). This matches Giordon's method from the multi-b code.
2. Adds automatic selection of lumicalc files for the pileup reweighting tool. This is following the recommendation for the PRW that only the lumicalc files for 2015+2016 data be used for processing mc16a samples and only those for 2017 for mc16c samples.

mc16d is not supported yet as I didn't know the expected run number.

Cheers,
Kate

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -249,26 +249,26 @@ EL::StatusCode BasicEventSelection :: fileExecute ()
   //
   // Metadata for the MC campaign to be used by PRW autoconfiguration.
   if(isMC() && m_autoconfigPRW)
-    {
-      std::string amiTag = "";
-      m_mcCampaignMD = "";
-      const xAOD::FileMetaData* fmd = 0;
-      if (m_event->retrieveMetaInput(fmd, "FileMetaData"))
-	{
-	  fmd->value(xAOD::FileMetaData::amiTag, amiTag);
-	  if( amiTag.find("r9364")!=std::string::npos ) m_mcCampaignMD = "mc16a";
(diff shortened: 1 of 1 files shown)


kratsg:
> mc16d is not supported yet as I didn't know the expected run number.

Interesting - I would think this is detailed in the [mc16d production twiki](https://twiki.cern.ch/twiki/bin/view/AtlasProtected/AtlasProductionGroupMC16d) but it seems to not be. I guess the next best thing is to grab one of the p-tags listed in the [derivation production twiki](https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/DerivationProductionTeam) and dump the run number... that seems annoying.

kpachal:
So I just asked James our lovely data prep coordinator and he says it has the same run number as mc16c. He also says not to worry about PRW for mc16d yet as several "issues still need to be discussed" -- it seems that this is not necessarily worked out yet on the other end. Maybe we wait and double back to this?

kratsg:
Ok, then I suggest just adding a comment above the mc16c campaign line mentioning to switch this to `mc16d` when the time comes. I suppose it's a hard-switch and it's not easy for xAH to support both (nor would we want to...).

kpachal:
Good point, comment added.

kpachal:
oops :P should type more carefully

PR #1164, opened 2018-02-19
Update MuonCalibrator with the latest recommendations

ntadej:
This updates muon calibrator to latest recommendations. As far as I know it requires AB 21.2.19.


kratsg:
> As far as I know it requires AB 21.2.19.

Then update the .travis.yml to account for those versions?

ntadej:
@kratsg, should I keep the old versions there or bump everything to 21.2.19?

kratsg:
> @kratsg, should I keep the old versions there or bump everything to 21.2.19?

You can see if it works in the old version.. but we probably just need to bump entirely, unfortunately. There's no way to support old releases and new releases at the same time.

ntadej:
I know that `Data17` Year string only works since 21.2.19 so I guess yes, everything should be bumped. Unfortunately Docker images for this release are not available yet so I will wait until they are available.

kratsg:
@lukasheinrich - when do you expect 21.2.19 docker images to come out for Base and Top releases?

kratsg:
@ntadej new docker images should be coming out in a few hours for 21.2.19.

kratsg:
images are up @ntadej

ntadej:
Bumped the release. Sorry for the delay.

kratsg:
@ntadej remove analysistop for now -- it's not getting built.

ntadej:
Finally, this seems to be green :)

PR #1166, opened 2018-02-19
Update JES uncertainties with the latest recommendations

ntadej:
JES uncertainties also require CalibArea. Requires 21.2.18.

### DIFF
--- a/xAODAnaHelpers/JetCalibrator.h
+++ b/xAODAnaHelpers/JetCalibrator.h
@@ -72,7 +72,9 @@ public:
   /// @brief config for JES Uncertainty Tool
   std::string m_JESUncertConfig = "";
   /// @brief JetUncertaintiesTool parameter
-  std::string m_JESUncertMCType = "MC15";
+  std::string m_JESUncertMCType = "MC16";
+  /// @brief JES CalibArea tag
+  std::string m_JESCalibArea = "CalibArea-02";
   /** @rst
     If you do not want to use SampleHandler to mark samples as AFII, this flag can be used to force run the AFII configurations.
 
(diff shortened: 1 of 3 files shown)


xAODAnaHelpers/JetCalibrator.h
```diff
@@ -72,7 +72,9 @@ class JetCalibrator : public xAH::Algorithm
   /// @brief config for JES Uncertainty Tool
   std::string m_JESUncertConfig = "";
   /// @brief JetUncertaintiesTool parameter
-  std::string m_JESUncertMCType = "MC15";
+  std::string m_JESUncertMCType = "MC16";
+  /// @brief JES CalibArea tag
+  std::string m_JESCalibArea = "CalibArea-01";
```
jbossios: Maybe you can put the latest one (CalibArea-02), which is needed for p-flow jets:
https://twiki.cern.ch/twiki/bin/view/AtlasProtected/JetUncertaintiesRel21Moriond2018SmallR

kratsg:
> Requires 21.2.18.

Update the .travis.yml if that's the case.

kratsg:
@ntadej a quick ping? Can you fix this up soon?

ntadej:
I can fix this one, but not the muons (still waiting for 21.2.19 docker image).

kratsg:
> I can fix this one, but not the muons (still waiting for 21.2.19 docker image).

That's fine. Have this one fixed. Then merge in the latest master to the new one which will update from 21.2.18 to 21.2.19 -- if that makes sense?

ntadej:
Done.

kratsg:
Can you change CalibArea to 02 like @jbossios mentioned?

ntadej:
I'm a bit confused about that as these seem to be two sets of different calibrations. I can change it, sure.

kratsg:
> I'm a bit confused about that as these seem to be two sets of different calibrations. I can change it, sure.

The CalibArea has the same as 01 -- but with fixes in 02. Think of it like CVMFS GroupData but for Jet calibrations alone. It makes it easier to fast-track some things and fixes.

ntadej:
@kratsg, I think this one should be merged in before muons, to avoid conflicts with version bumps.

PR #1172, opened 2018-02-21
Edits for DataScouting compatibility

kalderon:
A few changes are needed to run on DataScouting AODs, this pull request contains three sets of changes to allow this - "allEventsCBK protection", "Option to ignore lack of Primary Vertices" and "Add new jet cleaning infoSwitch".

I seem to have caused problems by reusing the branch used for a previous pull request (having rebased to upstream/master), and so this pull request encompasses the desired changes plus the edits from a previous pull request, as well as splitting each commit into a separate one per file. I only want to include the ones from today, ie commit 847615a onwards. If it would be better to make a new branch from master and cherry-pick the relevant commits to it, that is fine and I can cancel this pull request.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -99,35 +99,32 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
 
   m_decorSF = m_decor + "_SF";
 
-  bool allOK(false);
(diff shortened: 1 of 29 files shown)


kratsg:
Don't worry about the commits. We squash it. Double-check that the changes listed in this PR are the changes you expect and I'll merge/sign off.

kalderon:
Yes the changes are what I expect. Thanks.

PR #1184, opened 2018-03-28
Save Trigger Prescales Independent Of Decision

kkrizka:
This changes how trigger prescales are saved in an output ntuple. Instead of being saved for only triggers that passed, they are now saved for all triggers whose status is checked. This is needed for studying turn-on curves of triggers that are disabled for parts of a run. See:
https://groups.cern.ch/group/hn-atlas-TriggerHelp/Lists/Archive/Flat.aspx?RootFolder=%2fgroup%2fhn-atlas-TriggerHelp%2fLists%2fArchive%2fTurn-on%20Curves%20for%20Triggers%20Disabled%20For%20Parts%20Of%20Run&FolderCTID=0x012002001FA16C589111D24CA634A8138BBC0A5C

Changes:
* `isPassedBitsNames` is replaced by `triggerNames` 
* `isPassedBits`, `triggerPrescalesLumi` and `triggerPrescales` now reference `triggerNames` instead of `passedTriggers`

Tagging @ntadej, since you seem to care about prescales.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -901,26 +901,27 @@ EL::StatusCode BasicEventSelection :: execute ()
     if ( m_storeTrigDecisions ) {
 
       std::vector<std::string>  passTriggers;
+      std::vector<std::string>  triggerNames;
       std::vector<float>        triggerPrescales;
       std::vector<float>        triggerPrescalesLumi;
-      std::vector<std::string>  isPassedBitsNames;
       std::vector<unsigned int> isPassedBits;
 
       // Save info for the triggers used to skim events
       //
       for ( auto &trigName : triggerChainGroup->getListOfTriggers() ) {
         auto trigChain = m_trigDecTool_handle->getChainGroup( trigName );
-        if ( trigChain->isPassed() ) {
+        if ( trigChain->isPassed() )
           passTriggers.push_back( trigName );
-          triggerPrescales.push_back( trigChain->getPrescale() );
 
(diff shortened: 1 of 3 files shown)


ntadej:
So if I understand the linked discussion correctly, the only way to find out if the trigger was disabled, is using the prescale information? I would prefer you adding a new branch (which is disabled by default) specifying the info you need (e.g. a list of prescaled/disabled triggers in the event). This is a very specific use case and your changes might cause confusion.

kkrizka:
Yes, a disabled trigger has a prescale of 0. Only way to check.

To me it seems like adding another branch just for this is adding duplicate
information.

kratsg:
Is this good to go?

ntadej:
I’m mildly against the change as it is a very corner case requirement and it complicates the code needed to process the ntulles.

I leave it up to you to decide, but personally it would introduce a new branch and not make this change a breaking one.

kratsg:
> I’m mildly against the change as it is a very corner case requirement and it complicates the code needed to process the ntulles.
> 
> I leave it up to you to decide, but personally it would introduce a new branch and not make this change a breaking one.

Well, this is making the use case a bit more general, rather than limiting to triggers that are passed. How significant is this breaking change, because as you state, it is a corner case?

kkrizka:
It does not complicate the code too much. Instead of using the `triggersPassed` branch, you use the `triggerNames`. This is the breaking change.

But since it is a breaking change, I will update it to add a "disabledTriggers" branch instead.

PR #1186, opened 2018-04-05
Keep OR of triggers together for extra chains

ntadej:
We use trigger tool to split OR of triggers. For base triggers this is probably desirable, but sometimes one can also be interested in the actual OR of triggers (e.g. for prescale). This disables OR splitting of extra triggers.

In most cases this should not be breaking, except if one always passes OR of triggers instead of a comma separated list. If desired, I can add an extra configuration for this.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -473,6 +473,13 @@ EL::StatusCode BasicEventSelection :: initialize ()
     ANA_CHECK( m_trigDecTool_handle.setProperty( "OutputLevel", msg().level() ));
     ANA_CHECK( m_trigDecTool_handle.retrieve());
     ANA_MSG_DEBUG("Retrieved tool: " << m_trigDecTool_handle);
+
+    // parse extra triggers list and split by comma
+    std::string token;
+    std::istringstream ss(m_extraTriggerSelection);
+    while (std::getline(ss, token, ',')) {
+      m_extraTriggerSelectionList.push_back(token);
+    }
   }//end trigger configuration
 
   // 3.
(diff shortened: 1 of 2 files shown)


jbossios:
Hi @ntadej,

I always do something like this (OR of multiple triggers):
"m_triggerSelection": "HLT_j420 | HLT_j360 | HLT_j260 | HLT_j175 | HLT_j110 | HLT_j85 | HLT_j60 | HLT_j45 | HLT_j35 | HLT_j25 | HLT_j15 | L1_RD0_FILLED",

I'm just wondering if this will still be supported.

Best,
Jona

ntadej:
``m_triggerSelection`` did not change, but ``m_extraTriggerSelection`` did.
I guess then my change is possibly breaking.

kkrizka:
Can this be made backwards compatible by making a loop over all `m_extraTriggerSelection` elements split by a `,`?

ntadej:
This is how it is done, the problem are other ways of splitting triggers like @jbossios showed.

kratsg:
Updates? Are people ok with this?

kkrizka:
Yes, I'm happy with the important parts.

Although would it be possible to readd the result of the trigger tool to the "Extra Trigger Info Saved are" debugging information? I've used that output before.

Root/BasicEventSelection.cxx
```diff
@@ -645,10 +652,8 @@ EL::StatusCode BasicEventSelection :: execute ()
 
   if ( m_eventCounter == 0 && !m_extraTriggerSelection.empty() ) {
     ANA_MSG_INFO( "*** Extra Trigger Info Saved are :\n");
```
ntadej: @kkrizka, the info is still here, isn't it? Or do you want something else?

kratsg:
@ntadej lemme know the status of this

ntadej:
@kratsg, I have to check what @kkrizka recommended, but I'm swamped with getting the CONF note published. If this can wait another week or two, it would be nice.

Root/BasicEventSelection.cxx
ntadej: Added `TrigDecisionTool` evaluation for each comma separated string.

ntadej:
Any updates on this?

PR #1188, opened 2018-04-10
Some fixes to the autoprw code

kkrizka:
* Move campaign determination to `autoconfigurePileupRWTool()`. I had a few jobs fail when the MC run number could not be retrieved in `fileExecute()`.
* Fix combining of multiple campaigns. `prwConfigFile` needs to be reset each time in the for loop over campaigns.
* PRW_AUTOCONGIF to PRW_AUTOCONFIG. SUSY realized that they aren't reddit.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -251,34 +251,6 @@ EL::StatusCode BasicEventSelection :: fileExecute ()
 
   }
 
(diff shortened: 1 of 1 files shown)


kratsg:
> PRW_AUTOCONGIF to PRW_AUTOCONFIG. SUSY realized that they aren't reddit.

😆

PR #1194, opened 2018-04-23
WIP: move btag decorator outside of systematics loop

lschaefer:


kratsg:
@jdandoy I'm not sure this is correct, still. Shouldn't the systematic loop contain nominal?

jdandoy:
Hi @kratsg ,

The systematic loop does contain the empty string signifying nominal already, but it is not yet setup to run on the "Nominal" string.  More work will be ongoing to add it.

kratsg:
> The systematic loop does contain the empty string signifying nominal already, but it is not yet setup to run on the "Nominal" string. More work will be ongoing to add it.

In which case, moving the b-tagging out of it won't work as you'll still want to be able to decorate the systematic-varied jets differently, rather than b-tagging once.

jdandoy:
Yes you're right, the systematics that alter the jet pt will need to have the b-tagging decision re-evaluated.  We'll put this back in while we update to HelperFunction's getListofSystematics

PR #1198, opened 2018-05-03
WIP: making TTree memory flushing more frequent

kpachal:
WIP because I haven't tested that this consistently fixes my problem yet, it just looks promising from one job. Just want to get the discussion out there and collect ideas before going too much into it.

I found that when I added more systematics to my jobs they began to fail. The memory management plots showed the memory usage suddenly ballooning some time into the job: see [this one](https://bigpanda.cern.ch/memoryplot/?pandaid=3907846163) as a classic example. Even in cases where jobs succeeded the same behaviour was visible ([example](https://bigpanda.cern.ch/memoryplot/?pandaid=3907442946)). 

Someone helpfully forwarded a thread on the flavour tagging list looking at a very similar issue, which they found was due to TTree memory flushing intervals essentially being too long. (Sorry I'd link the mail archive here but it's protected for that list, for some reason...) I am trying the same fix here which they used, and on a test job it seemed improved. However, I feel like there's got to be a better way of dealing with this ... 

Tagging @kratsg and @kkrizka for thoughts!

Cheers,
Kate

### DIFF
--- a/Root/TreeAlgo.cxx
+++ b/Root/TreeAlgo.cxx
@@ -196,7 +196,6 @@ EL::StatusCode TreeAlgo :: execute ()
     if(systName.empty()) treeName = "nominal";
 
     ANA_MSG_INFO( "Making tree " << m_name << "/" << treeName );
-    std::cout << "Making tree " << m_name << "/" << treeName << std::endl;
     TTree * outTree = new TTree(treeName.c_str(),treeName.c_str());
     if ( !outTree ) {
       ANA_MSG_ERROR("Failed to instantiate output tree!");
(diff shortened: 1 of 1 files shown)


kratsg:
I've generally known about this sort of ~hack~ fix which helps here. If this helps reduce the memory usage of the jobs with no obvious downfalls, this is a fine addition.

kkrizka:
Does this really help? The [default value](https://root.cern.ch/doc/v608/classTTree.html#ad4c7c7d70caf5657104832bcfbd83a9f) is 30MB, which is already quite small.... Definitely not the GBs that are causing the jobs to fail.

PR #1199, opened 2018-05-04
Fix BJetEfficiencyCorrector loop for uncalibrated btaggers

kkrizka:
The `BJetEfficiencyCorrector` loops over the list of recommended systematics (this includes the nominal "systematic"). For the unsupported taggers, there are no scale factors and hence the recommended systematics list is empty (not even "nominal").

The fix is to invert the double loop in `executeEfficiencyCorrection` to make the outer loop be over jets and apply btagging decision to it and then the inner loop over the systematics (if `m_getScaleFactors = true`, otherwise set a single SF of 1).

This should be done regardless, because the systematics only affect the SF and not the decision. There is no need to rerun the `BTaggingSelectionTool` for every single systematic.

See problem seen by @kpachal at
https://groups.cern.ch/group/atlas-sw-xAODAnaHelpersFW/Lists/Archive/Flat.aspx?RootFolder=%2Fgroup%2Fatlas-sw-xAODAnaHelpersFW%2FLists%2FArchive%2FNon-supported%20taggers%20now%20set%20to%20-1&FolderCTID=0x012002008E97964F804B4E4BBC47A7E29F81D790

Marking WIP since I want to further validate this PR by running over a larger dataset on the grid.


kpachal:
Thank you Karol!!!

kkrizka:
Should be good to merge.

PR #1203, opened 2018-05-08
Introducing tau efficiency corrections

fscutti:
Tau efficiency corrections are introduced for MC using just one branch which is the cumulative weight of all corrections. Instructions on how to include the new branch are provided in HelpTreeBase. Some changes were made for muons as well but I now realise that the changed code is completely outdated legacy code. So I will issue a new MR in the future to remove it.


kratsg:
This looks great. Note that at some point in the future, all copyright/author statements at the tops of files will be removed (in favor of using git log/git blame).

PR #1207, opened 2018-05-15
Updates to automatic PRW configuration

kkrizka:
* Default 2017 MC is now mc16d
* Configs are now in `dev/SUSYTools/PRW_AUTOCONFIG_SIM/files` (new SUSYTools location)
* Differentiate between FS and AFII, decided from sample `metaString`
* Return `FAILURE` when a PRW config file is not found

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -1014,9 +1014,17 @@ StatusCode BasicEventSelection::autoconfigurePileupRWTool()
   const xAOD::EventInfo* eventInfo = 0;
   ANA_CHECK( m_event->retrieve( eventInfo, "EventInfo" ) );
 
+  // Determine simulation flavour
+  const std::string SimulationFlavour = wk()->metaData()->castString("SimulationFlavour");
+  if(SimulationFlavour.empty())
+    {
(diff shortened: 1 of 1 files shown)


kratsg:
Is `SimulationFlavour` a thing automatically/correctly set? Or does the user now have to set it with something like `c.sample("410000", {"SimulationFlavor": "AFII"})`?

kkrizka:
Sadly it has to be manually set. I use the following pattern.
```
c.sample('*_a875_*',SimulationFlavour='AFII')
c.sample('*_s3126_*',SimulationFlavour='FS')
```

kkrizka:
Btw, the same `metaString` is used by the `JetCalibrator`. It would be nice to standardize AFII vs FS determination in all xAH algorithms. Via `c.sample` seems like a nice way.

kratsg:
Hrmm, I suggest we update the error message, in the case of autoconfiguration, to explain what the user needs to do (add `SimulationFlavor` metaString) and point to the docs? https://xaodanahelpers.readthedocs.io/en/master/UsingUs.html#configuring-samples

ecorrigan:
Hi! @kratsg 

I have actually just implemented something similar and was about to issue a PR, when I saw that this has already been done. Is there any reason it is still not merged? These changes are needed to run. 

Thanks!

Eric

kpachal:
+1 :)

PR #1208, opened 2018-05-16
Fix AOD crashes for small and fat jets, more flexible TreeAlgo, add to debug information

kalderon:
The commit history here is a bit of a mess since I waited a long time between making changes and this request, and both I and someone else implemented clusters in TreeAlgo. I've tried to keep the UCATLAS cluster implementation everywhere.

The changes have the following affects:
* safety against non-presence of truth jets in MC: optional m_haveTruthJets=true that, when false, doesn’t try to retrieve truth jets for JVT in MC (since AODs now don’t have truth jets)
* safety against non-presence of “GhostTrackCount” in fat jets
* allow multiple large-R collections with different names and detailStrs
* allow truth fat jet branch to be named
* debug messages in TreeAlgo when jet/cluster containers are not available and hence an event is not written

### DIFF
--- a/Root/TreeAlgo.cxx
+++ b/Root/TreeAlgo.cxx
@@ -299,7 +299,6 @@ EL::StatusCode TreeAlgo :: execute ()
       }
     }
 
-
     if (!m_truthFatJetContainerName.empty() )   { helpTree->AddTruthFatJets(m_truthFatJetDetailStr, m_truthFatJetBranchName);               }
     if (!m_tauContainerName.empty() )           { helpTree->AddTaus(m_tauDetailStr);                               }
     if (!m_METContainerName.empty() )           { helpTree->AddMET(m_METDetailStr);                                }
--- a/Root/TreeAlgo.cxx
+++ b/Root/TreeAlgo.cxx
@@ -296,6 +296,7 @@ EL::StatusCode TreeAlgo :: execute ()
       }
     }
 
+
     if (!m_truthFatJetContainerName.empty() )   { helpTree->AddTruthFatJets(m_truthFatJetDetailStr, m_truthFatJetBranchName);               }
     if (!m_tauContainerName.empty() )           { helpTree->AddTaus(m_tauDetailStr);                               }
     if (!m_METContainerName.empty() )           { helpTree->AddMET(m_METDetailStr);                                }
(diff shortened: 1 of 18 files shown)


kalderon:
It looks like the AnalysisTop,21.2.20 build timed out but everything else worked fine. I will close and reopen the request to trigger a retry.

kratsg:
> The commit history here is a bit of a mess since I waited a long time between making changes and this request, and both I and someone else implemented clusters in TreeAlgo. I've tried to keep the UCATLAS cluster implementation everywhere.

I added the clusters for @fizisist's and my work on gFEX.

Root/TreeAlgo.cxx
```diff
@@ -240,20 +288,34 @@ EL::StatusCode TreeAlgo :: execute ()
...
+	else{ helpTree->AddFatJets       (m_fatJetDetails.at(ll), m_fatJetBranches.at(ll).c_str()); }
+      }
+    }
+
+    if (!m_truthFatJetContainerName.empty() )   { helpTree->AddTruthFatJets(m_truthFatJetDetailStr, m_truthFatJetBranchName);               }
     if (!m_tauContainerName.empty() )           { helpTree->AddTaus(m_tauDetailStr);                               }
     if (!m_METContainerName.empty() )           { helpTree->AddMET(m_METDetailStr);                                }
     if (!m_METReferenceContainerName.empty() )  { helpTree->AddMET(m_METReferenceDetailStr, "referenceMet");       }
     if (!m_photonContainerName.empty() )        { helpTree->AddPhotons(m_photonDetailStr);                         }
     if (!m_clusterContainerName.empty() )       { helpTree->AddClusters(m_clusterDetailStr);                         }
```
kratsg: is this line double-adding clusters?

kalderon: Indeed. This and L540-L546 (duplicated by 564-583, which allows multiple cluster collections) should be removed. I have done so in https://github.com/kalderon/xAODAnaHelpers/commit/bcbb815f3a8f008fe994e6d173e6aa57ff6d828a - what is the best way to proceed?

kratsg: Just push the changes to the same branch that this PR was made from. It'll auto-update.

PR #1216, opened 2018-06-19
protect against prun submissions from cc7

kratsg:
/cc @stephensekula


kratsg:
For similar code used in `atlasexternals` -- see https://gitlab.cern.ch/atlas/atlasexternals/blob/master/Build/AtlasCMake/modules/AtlasInternals.cmake#L214.

A future update would create a nice function in xAH to identify the architecture of the machine.

PR #1220, opened 2018-07-04
allow for multiple variations on slc6 in lsb_release

kratsg:
SLC6 is reported in `lsb_release -d` as one of two possible variations atm:

```

Description:	Scientific Linux release 6.9 (Carbon)
Description:    Scientific Linux CERN SLC release 6.9 (Carbon)

```

this patch allows both to be recognized/identified as SLC6.


kratsg:
Note there are more possible variations and when we run into them, we'll add more into xAH.

PR #1224, opened 2018-07-10
WIP: Use the PMG xsec DB

kkrizka:
This PR implements the `PMGCrossSectionTool` to set the cross-section and filter efficiency sample metadata for official samples. The goal is to replace the SUSYTools style xsec database that we are using now (and our base file is quite outdated). It is done by `xAH_run.py` after all other cross-section databases (SUSYTools, .config files) are loaded, so this value (if it exists) override anything above.

Some other notes:
* Since `xAH_run` does not have access to the sample contents, the DSID for the DB lookup is extracted from the sample name. 
* The `dev/PMGTools/PMGxsecDB_mc16.txt` files is used for the database.
* Additionally contents of `dev/xAODAnaHelpers/xsecdb` are loaded in case a samples does not exist in the official PMG DB.

Marking as WIP, to allow for discussion and to update the `getDSInfo.py` script for this format. Also this needs to be turned into C++ to work on the grid, since one cannot set metadata in Python without the weird argparse "integer is required" error.

Also sneaking in a change I forgot about. `Config.sample` now supports wildcards. Example of usage would have been to set AFII and FS sample types by matching the tags in the sample name.


kratsg:
> Since xAH_run does not have access to the sample contents, the DSID for the DB lookup is extracted from the sample name.

Should move this to [utils.py](https://github.com/UCATLAS/xAODAnaHelpers/blob/master/python/utils.py) through a `get_dsid(sampleName)` function or something like that. Something like this has worked for me: https://github.com/kratsg/optimization/blob/master/root_optimize/utils.py#L135-L148



> Additionally contents of dev/xAODAnaHelpers/xsecdb are loaded in case a samples does not exist in the official PMG DB.

We'll need to document somewhere the order of priority. I suspect we'll want:

- PMG XSec DB
- SUSY Tools ReadMetaDir
- xAH XSec DB

Note that we need to be a bit careful, since I think ST will automatically add up cross-sections instead of replacing the existing value.

kratsg:
@kkrizka is this still a WIP?

PR #1232, opened 2018-08-10
Set missing scale factors to -1 (breaking)

ntadej:
Set missing scale factors to -1, to be sure that they are missing (being 1 is not completely clear, maybe just the SF is not needed).

Also lower the electron efficiencies threshold and add a separate trigger one (as they have it higher, but do not want to hardcode the value in).

I guess there is no need to say this is a breaking change, but it will probably avoid mistakes for many users.


alexandertuna:
👍

PR #1235, opened 2018-08-15
Revert "Make cutflow all bin be the pre-derivation count. (#1226)"

kratsg:
This reverts commit a4f37e018c5eb531a78b98b6f39d89b41088f67c.


kratsg:
See #1226

PR #1236, opened 2018-08-16
Cutflowall

kkrizka:
Same as #1226, but with histogram creation moved back to `initialize()`. The call to `isMC()` now works.

PR #1237, opened 2018-08-19
move histogram creation to histInitialize()

kratsg:
Fixes a bug introduced by #1236. Technically, this was a problem with the original code as the histogram creation was done in `initialize()` and not `histInitialize()`.


fscutti:
Thanks! This seems to fix the problem.

PR #1242, opened 2018-09-12
Tau tracks

fscutti:
Updates for dumping tau tracks information in the ntuples. 

- A new algorithm is introduced to match an arbitrary and configurable collection of jets with the TauJet object. This is useful for cases where the tau seedJet (LC topo jet) is not available in xAOD but one would need to get a handle on some original jet info. The algorithm (TauJetMatching) has a configurable DR matching criterion and for now decorates taus with the matched jet width. It should be executed before OLR. I'm currently using it before tau selection and after tau calibration.

- tau tracks information is now decorated onto the tau in the TauSelector algorithm and retrieved when filling the tau container. This is optional and controlled by the HelperClasses flag trackparams. Performing the decoration is also optional and controlled by the flag m_decorateWithTracks in the TauSelector configuration.

### DIFF
--- a/Root/LinkDef.h
+++ b/Root/LinkDef.h
@@ -51,6 +51,7 @@
 #include <xAODAnaHelpers/HelperFunctions.h>
 #include <xAODAnaHelpers/OverlapRemover.h>
 #include <xAODAnaHelpers/TrigMatcher.h>
+#include <xAODAnaHelpers/TauJetMatching.h>
 #include <xAODAnaHelpers/Writer.h>
 #include <xAODAnaHelpers/MessagePrinterAlgo.h>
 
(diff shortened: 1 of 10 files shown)


fscutti:
Hi @kratsg, can I ping you on this? Is everything ok with the update?

kratsg:
My main concern is that this breaks the meaning of `m_trackparams` for the Taus. Is there a different variable you can add to the info switch for taus that's not also used by muons/electrons -- so it's conceptually the same? E.G. if nobody runs the TauJet matching, they should still want to enable `m_trackparams` and not expect it to break.

fscutti:
yes you're right. I'll add two other variables then. One to enable only variables decorated by the TauJetMatching algorithm and one for the track parameters which are currently decorated, which are not exactly the same parameters decorated for electrons or muons when trackparams is enabled.

kratsg:
> One to enable only variables decorated by the TauJetMatching algorithm

call it like `m_xahTauJetMatching` and can only be added by matching "xAHTauJetMatching" or something.

PR #1245, opened 2018-09-12
Add MC/MC scale factors to b-tagging.

kkrizka:
Added support for [MC/MC scale factors for b-tagging](https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/BTagCalib2017#MC_MC_Scale_Factors_for_Analysis). The MC type can be set via the `m_EfficiencyCalibration`, with `auto` guessing the type from the sample name. The latter was [inspired via SUSYTools](https://gitlab.cern.ch/atlas/athena/blob/c5869763675d1bd421853a58031788c10e9e981e/PhysicsAnalysis/SUSYPhys/SUSYTools/SUSYTools/ISUSYObjDef_xAODTool.h#L157).

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -16,6 +16,8 @@
 #include <EventLoop/StatusCode.h>
 #include <EventLoop/Worker.h>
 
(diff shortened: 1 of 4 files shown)


kkrizka:
Should be ready to go.

PR #1250, opened 2018-09-27
Custom jvt cuts need to use detector-eta

### LINKED ISSUE #1249: Bug in JVT cut in JetSelector
Hello,

I just want to report that the JVT cut should be updated to use DetectorEta instead of eta at EM/Consit scale, since that eta is no longer the detector eta in R21:
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetSelector.cxx#L1109

I never used that option but there might be someone using it.

Best,
Jona

kratsg:
This resolves #1249. Constituents are origin-corrected by default in R21 which means that the eta of a jet at the EM/Constituent scale is the origin-corrected eta. Custom Jvt cuts (and fJvt cuts) need to be applied using DetectorEta which is an available decoration on jets. This is a migration change from R20 where origin-correction was not applied at the constituent level, so that `jet->eta()` returned the DetectorEta if the jet was at the EM-scale.

### DIFF
--- a/Root/JetSelector.cxx
+++ b/Root/JetSelector.cxx
@@ -1096,32 +1096,26 @@ int JetSelector :: PassCuts( const xAOD::Jet* jet ) {
   }//do forward JVT
 
   // JVT pileup cut
-  //
   if ( m_doJVT ) {
-    ANA_MSG_DEBUG("Checking JVT cut...");
-    if ( m_JVTCut > 0 && jet->getAttribute< float >( "Jvt" ) < m_JVTCut ) { ANA_MSG_DEBUG( " pt/eta = "<<jet->pt()<<"/" << jet->eta() ); }
+    // NB: origin-correction is applied at constituent level. Eta for Jvt needs to be the DetectorEta explicitly.
+    float jet_pt = jet->pt();
+    float jet_eta = jet->getAttribute<float>("DetectorEta");
+    float jet_jvt = jet->getAttribute<float>("Jvt");
+    ANA_MSG_DEBUG("Checking Jvt cut for jet pT=" << jet_pt << " MeV, DetectorEta=" << jet_eta <<", and Jvt="<< jet_jvt );
+    bool result = false;
 
-    // Old usage: check manually whether this jet passes JVT cut
-    //
-    if ( m_JVTCut > 0 ) {
-      if ( jet->pt() < m_pt_max_JVT ) {
-        ANA_MSG_DEBUG("Pass JVT-pT Cut");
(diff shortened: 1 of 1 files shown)


Root/JetSelector.cxx
```diff
@@ -1096,32 +1096,26 @@ int JetSelector :: PassCuts( const xAOD::Jet* jet ) {
...
-            ANA_MSG_DEBUG(" upper JVTCut is " << m_JVTCut << " - cutting this jet!!" );
-            return 0;
-          } else {
-            ANA_MSG_DEBUG(" upper JVTCut is " << m_JVTCut << " - jet passes JVT ");
-          }
-        }
-      }
+    // NB: origin-correction is applied at constituent level. Eta for Jvt needs to be the DetectorEta explicitly.
+    float jet_pt = jet->pt();
+    float jet_eta = jet->getAttribute<float>("DetectorEta");
```
mattleblanc: Should we check to make sure it's there? Sometimes this doesn't make it into the DAODs ... maybe have everything crash & burn if it's missing.

kratsg: If the Jvt crashes, so will this. jvt doesn't protect against it does it?

kratsg: I'll leave it as it is. It'll crash if someone's not using it right.

kalderon:
I noticed that this merge request inverts the behaviour of any custom JVT cuts given, ie requires JVT *less* than the value given if m_JVTcut is greater than 0. This is somewhat problematic problematic...

For example, having a JetSelector with:
```
"m_cleanJets"               :   True,
"m_doJVT"                   :   True,
"m_JVTCut"                  :   0.59,
```
which worked fine before, using the default JVT cut for the medium working point, I now get:

> JetSelector.JetSelecto...DEBUG   In pass cuts
> JetSelector.JetSelecto...DEBUG   Checking Jvt cut for jet pT=307842 MeV, DetectorEta=0.294441, and Jvt=1
> JetSelector.JetSelecto...DEBUG   Custom JVT working point with pT<60000, |eta|<2.4, Jvt<0.59
> JetSelector.JetSelecto...DEBUG    ... jet does not pass Jvt cut

which is definitely counter to my intentions...

If I remove the "m_JVTCut" and pick up the default Medium working point without trying to specify exactly that, then it works as expected. If I set the cut value to -0.59, just out of curiosity, then again it has behaviour that makes sense (JVT<-0.59 fails, JVT>-0.59 passes). 

This seems a much bigger change than just swapping detector eta for jet eta. Was this intentional?

PR #1254, opened 2018-10-01
Merge TLA fork (no substantive change)

kalderon:
I developed a package here but then moved it elsewhere, so there are a large number of commits that should be squashed into one. The only actual change is a correction to a debug message and some whitespace additions/removal.


kratsg:
FYI, unless it says there are conflicting changes, you never need to merge into master/update.

PR #1255, opened 2018-10-01
make PrimaryVertices configurable, not hard-coded

### LINKED ISSUE #1251: Inconsistency b/w NPV calculation & PV cut
Hello,

In BasicEventSelection it's possible to select a different vertex container than "PrimaryVertices" for the PVNTrack cut but then in EventInfo the name of the vertex container for the NPV calculation is hardcoded as "PrimaryVertices". Can we fix this? How? Which algorithm should be used to set a vertex container variable for the NPV calculation?

Best,
Jona

kratsg:
This resolves #1251.

The tl;dr is that setting `m_vertexContainerName` on any algorithm will make sure that the algorithm uses that for the vertex container when doing primary-vertex-related calculations -- including `TreeAlgo` which retrieves the vertex container and passes it through to `EventInfo`.

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -137,6 +137,11 @@ void HelpTreeBase::FillEvent( const xAOD::EventInfo* eventInfo, xAOD::TEvent* /*
 
   this->ClearEvent();
 
+  // only retrieve the vertex container if it's not set and the user asks for that information
+  if( m_eventInfo->m_infoSwitch.m_pileup && !vertices ) {
+    HelperFunctions::retrieve( vertices, m_vertexContainerName, m_event, 0);
+  }
+
   m_eventInfo->FillEvent(eventInfo, m_event, vertices);
 
   this->FillEventUser(eventInfo);


kratsg:
@jbossios - just need a thumbs up from you on this and I'll merge it in.

Root/EventInfo.cxx
```diff
@@ -297,13 +297,8 @@ void EventInfo::FillEvent( const xAOD::EventInfo* eventInfo,  xAOD::TEvent* even
...
 
-    if ( event ) {
-      const xAOD::VertexContainer* vertices(nullptr);
-      HelperFunctions::retrieve( vertices, "PrimaryVertices", event, 0 );
-      m_npv = HelperFunctions::countPrimaryVertices(vertices, 2);
-    } else {
-      m_npv = -1;
-    }
+    m_npv = -1;
+    if(vertices) m_npv = HelperFunctions::countPrimaryVertices(vertices, 2);
```
jbossios: @kratsg This breaks the backward compatibility. There are codes based on xAH which are using FillEvent(eventInfo) and getting NPV from PrimaryVertices. This update will get them always NPV=-1. So I suggest keeping as default PrimaryVertices but allowing to use a custom PV container.

An example is IsolatedJetTree:
https://gitlab.cern.ch/atlas-jetetmiss-jesjer/MCcalibrations/IsolatedJetTree/blob/master/Root/IsolatedJetAlgo.cxx#L488

I will update IsolatedJetAlgo to pass the PV container but there might be other packages which will be affected by this.

kratsg: I at least updated it so that if the vertices received is a nullptr, and if they're looking at information that requires the vertices (e.g. pileup), then it'll retrieve it and pass it through the HelpTreeBase code

PR #1260, opened 2018-10-12
Grid submission

lschaefer:
I've had a few jobs go into the exhausted state because of the nGBPerJob default of 2. I think this is something that can be set by panda so the default can safely be None. 

(I'm not sure why so many changes that are already in master are showing up as my changes, but I only changed one line in xAH_run.py)


kratsg:
You most likely need to rebase because you started your commit from a different point than master so things diverged.

kratsg:
Note that the default of 2 GB per job is actually a grid limitation really. Normally, if you don't set a default, you'll run into a lot of scouting issues because some jobs will take > 2 GB of memory. I don't think there are enough grid sites that can handle more than 2 GB of memory...

ntadej:
@kratsg, this is not about memory but how many jobs to make. It will use about 2GB of inputs per job. This is surely too low, but I would not make it automatic as it can have negative consequences.

@lschaefer, I use 12 GB for normal jobs and 4 GB for systematics. At least at the time of optimising this value automatic settings would be to large and would cause scouting jobs being too long or fail.

Note that generally grid jobs should not take more than a day. You can use this for time to optimise.

lschaefer:
Hi @kratsg , @ntadej , 
Thanks for the comments. I did rebase before committing and I checked a few changes which seem to agree with what's already in master, but it's very possible I did something else wrong. But regarding the nGBPerJob, so the recommendation is for the user to optimize for their own jobs? So you would like to leave the default at 2?

mswiatlo:
IMO we shouldn't leave the default at 2, since that will almost certainly lead to exhaustion. @ntadej and @lschaefer are right and this default should be raised imo.

ntadej:
Personally I would remove the default but give a warning if user does not override it. Using an optimised version of the setting would give you the best throughput, but having an ability for scouting to decide is also nice to have.

kratsg:
> Personally I would remove the default but give a warning if user does not override it. Using an optimised version of the setting would give you the best throughput, but having an ability for scouting to decide is also nice to have.

Let's remove the default. I'll add a warning in another PR.

lschaefer:
Thanks everyone for the discussion!

cohm:
I agree with the conclusion. But just to check: part of the reason to have the scout jobs in the first place is exactly this, isn't it? To check how much output is produced for a given amount of input. I think they typically run on one file each, and then assume the output scales linearly with input. Based on the amount of output produced by the scout jobs the nGBPerJob parameter is tuned to the optimal value for the site in question by grid magic. It could also be that I'm just naive and optimistic, it has happened before.

kratsg:
> I agree with the conclusion. But just to check: part of the reason to have the scout jobs in the first place is exactly this, isn't it? To check how much output is produced for a given amount of input. I think they typically run on one file each, and then assume the output scales linearly with input. Based on the amount of output produced by the scout jobs the nGBPerJob parameter is tuned to the optimal value for the site in question by grid magic. It could also be that I'm just naive and optimistic, it has happened before.

I mean, so to explain what happened -- this option was set to this default 4 years ago in xAH. We're mostly discussing (a) my code rot [I set this to 2 GB a long time ago based on ASG recs at the time] and (b) no easy way to test these options to make sure they're still valid. So I'm fine with removing the default as the grid should be smarter 4 years later.

ntadej:
@cohm, to clarify it a little more: Scouting takes quite a large amount of input by default (I think it's 200 files or 20 GB whatever is acceptable). When you run systematics, this means your scouting jobs might run 2 or 3 days if you are unlucky. This means that the majority of your jobs will have to wait that time for processing to really start. Once you know your jobs' requirements, helping panda broker your jobs can make them finish much faster.

PR #1266, opened 2018-10-26
Add new egamma charge-flip support

ntadej:
This MR adds support to:
 - central charge-flip SFs
 - charge-flip killer variables and recalculation

Note that charge-flip SFs for some reason use a different tool and need a file explicitly set. I try to autogenerate the filename for now and one only needs to specify the path but this introduces some hardcoded stuff - let me know if this is acceptable.

### DIFF
--- a/xAODAnaHelpers/ElectronEfficiencyCorrector.h
+++ b/xAODAnaHelpers/ElectronEfficiencyCorrector.h
@@ -19,6 +19,7 @@
 
 // external tools include(s):
 #include "ElectronEfficiencyCorrection/AsgElectronEfficiencyCorrectionTool.h"
+#include "ElectronEfficiencyCorrection/ElectronChargeEfficiencyCorrectionTool.h"
 
 // algorithm wrapper
 #include "xAODAnaHelpers/Algorithm.h"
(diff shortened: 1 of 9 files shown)


xAODAnaHelpers/ElectronEfficiencyCorrector.h
```diff
@@ -108,6 +116,8 @@ class ElectronEfficiencyCorrector : public xAH::Algorithm
   std::string m_TrigEffSF_tool_name;                                  //!
   AsgElectronEfficiencyCorrectionTool  *m_asgElEffCorrTool_elSF_TrigMCEff = nullptr; //!
   std::string m_TrigMCEff_tool_name;                                  //!
+  CP::ElectronChargeEfficiencyCorrectionTool *m_elChflipEffCorrTool = nullptr; //!
```
kratsg: This should be using the AnaToolHandle interface.

ntadej: How do I dynamically set the tool name as we have multiple instances for each working point?

kratsg: `tool_handle.setName()`

PR #1267, opened 2018-10-29
Fix ROOT batch mode

ntadej:
It seems that ROOT 6.14 changed something so we need to set batch mode earlier.


ntadej:
X is spawning over ssh and it was not before.

kratsg:
> X is spawning over ssh and it was not before.

It would be good to figure out what part is spawning that. The xAH python package isn't importing or using ROOT at all, so it's not clear what libraries we use that do that. Instead of moving the `import ROOT` up higher, what if you just explicitly set batch mode where it stands? Does that fix things?

PR #1268, opened 2018-10-29
Update MET significance interface

ntadej:
Update MET significance interface that changed in 21.2.50.

**Note: requires 21.2.50 (updated in travis)!**


ntadej:
It is new, you can wait with merging it but it should compile and run now (also if someone else needs it).

kratsg:
We should update this to 21.2.51 due to issues with MET Sig lately.

ntadej:
Ok, let's merge this in today (will update it now) and then I'll rebase my other MRs on this one.

PR #1272, opened 2018-11-02
Update and cleanup JetCalibrator

ntadej:
This updates JetCalibrator to the new recommendations available since 21.2.41. JES and JER tools are now merged.

This MR:
 - removes all JER references
 - calibration areas are now optional
 - MC type for uncertainties is automatically set
 - default config files are updated

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -36,8 +36,6 @@
 // tools
 #include "JetCalibTools/JetCalibrationTool.h"
 #include "JetUncertainties/JetUncertaintiesTool.h"
-#include "JetResolution/JERTool.h"
-#include "JetResolution/JERSmearingTool.h"
 #include "JetSelectorTools/JetCleaningTool.h"
 #include "JetMomentTools/JetVertexTaggerTool.h"
 #include "JetTileCorrection/JetTileCorrectionTool.h"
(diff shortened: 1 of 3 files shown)


Root/JetCalibrator.cxx
```diff
@@ -163,13 +163,17 @@ EL::StatusCode JetCalibrator :: initialize ()
...
       ANA_MSG_WARNING( "Could not access simulation flavour from EL::Worker. Treating MC as FullSim by default!" );
     } else {
       m_isFullSim = (stringMeta == "AFII") ? false : true;
     }
     if ( !m_isFullSim ) {
       m_calibConfig = m_calibConfigAFII;
+      m_uncertMCType = "AFII";
+    } else {
+      // Insitu should not be applied to the trimmed jets, per Jet/Etmiss recommendation
+      if ( m_forceInsitu && m_calibSequence.find("Smear") == std::string::npos) m_calibSequence += "_Smear";
```
jbossios: @ntadej  Can we use a m_forceSmear instead of m_forceInsitu?

ntadej:
Forgot that .50 does not work due to MET interface changes. Will remove it.

mswiatlo:
At this point the jet calibrations are physically meaningless in anything below 51 (see Steven Schramm's recent emails). I would purge compatibility with pre-51, or else wrap around the MET tool in such a way that this can be updated.

kratsg:
> At this point the jet calibrations are physically meaningless in anything below 51 (see Steven Schramm's recent emails). I would purge compatibility with pre-51, or else wrap around the MET tool in such a way that this can be updated.

I guess we should just bump straight to 21.2.51. @ntadej can you do that?

kratsg:
I've updated this to 21.2.51. when tests pass, ill merge in.

xAODAnaHelpers/JetCalibrator.h
```diff
@@ -57,24 +57,22 @@ class JetCalibrator : public xAH::Algorithm
...
-  std::string m_calibSequence = "JetArea_Residual_Origin_EtaJES_GSC";
-  /// @brief CalibArea tag on CVMFS (for pointing at an area outside of the current release)
-  std::string m_calibArea = "00-04-81"; // current as of 19/11/17
-  /// @brief config for JES Uncertainty Tool
-  std::string m_JESUncertConfig = "";
-  /// @brief JetUncertaintiesTool parameter
-  std::string m_JESUncertMCType = "MC16";
-  /// @brief JES CalibArea tag
-  std::string m_JESCalibArea = "CalibArea-02";
+  std::string m_calibConfigAFII = "JJES_MC16Recommendation_AFII_EMTopo_April2018_rel21.config";
```
lschaefer: I think the double J is a typo and it should rather be "JES_MC16Recommendation_AFII_EMTopo_April2018_rel21.config"

lschaefer:
Hi all,
I put a comment below an individual line, but I'm not sure it will show up in the overall discussion so I'll add it here too: I think there is a typo in the JES config file name ("JJES..." instead of "JES"). Thanks!

PR #1274, opened 2018-11-13
Bring fork into line - no net changes

kalderon:
Essentially, I have all these commits where I added something and then moved it elsewhere, mixed in with a few merge / rebase conflicts, the end result of which is to do precisely nothing with respect to the current UCATLAS master. If they can be squelched into one commit and merged into UCATLAS master, such that they don't keep popping up and being annoying every time I want to make a merge request / rebase / etc, that would be excellent. If this isn't possible / isn't the way to go about that / is a pain for you, then I'll see if this is something I can do myself, and sorry for the spam.

Oh, and maybe I should have done this from a non-master branch? I can delete the request and try again if so.

Will

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -637,24 +637,29 @@ void HelpTreeBase::FillL1Jets( const xAOD::JetRoIContainer* jets, bool sortL1Jet
       m_nL1Jet++;
     }
   }
-
+  
   else {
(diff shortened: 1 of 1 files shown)


ntadej:
@kalderon, it seems that there are no changes (at least if we believe GitHub). Probably it's safe for you to just start using the current master.

kalderon:
@kratsg sorted me out, thanks both!

PR #1275, opened 2018-11-14
Modify minor JES uncertainty and smearing changes

balunas:
This is a super simple request which changes 2 things that were recently done in JetCalibrator (in #1272 ):

- Revert `m_uncertMCType` to be configurable. It was recently changed to automatically be set to `MC16` or `AFII` depending on the type of simulation, but this breaks compatibility with existing fat jet calibrations for example.
- Change the default value of `m_forceSmear` to false. This is to avoid confusing default behavior that people have no way of knowing about without reading through the source code (I for one spent a fair amount of time trying to figure out why the tool was attempting to smear my fat jets when I didn't didn't configure it to). Anyone who wants this feature should turn it on in their configuration, or simply include smearing in their CalibSequence.

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -179,6 +179,8 @@ EL::StatusCode JetCalibrator :: initialize ()
     }
   }
 
+  if(m_uncertMCType.empty()) m_uncertMCType = m_isFullSim ? "MC16" : "AFII";
+
   // initialize jet calibration tool
   ANA_CHECK( ASG_MAKE_ANA_TOOL(m_JetCalibrationTool_handle, JetCalibrationTool));
   ANA_CHECK( m_JetCalibrationTool_handle.setProperty("JetCollection",m_jetAlgo));
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -143,8 +143,6 @@ EL::StatusCode JetCalibrator :: initialize ()
     return EL::StatusCode::FAILURE;
   }
 
-  m_uncertMCType = "MC16";
-
   if ( !isMC() ) {
     // Insitu should not be applied to the trimmed jets, per Jet/Etmiss recommendation
     if ( m_forceInsitu && m_calibSequence.find("Insitu") == std::string::npos) m_calibSequence += "_Insitu";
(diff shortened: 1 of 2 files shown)


ntadej:
I don't mind the second change, but I would keep `m_uncertMCType` to be auto configurable. Can you add a check that if the value is not set, you do the autoconfiguration, if it is set, then you do not do it?

Just for clarification, do you do insitu calibration for data also for fat jets?

balunas:
Sure, sounds reasonable enough to me, I'll go ahead and put that feature back for cases where the user doesn't specify anything. For fat jets, no in-situ calibration is applied, I think because the CP group hasn't derived it yet.

ntadej:
Then I would put both `forceSmear` and `forceInsitu` to `false` by default, just for consistency.

balunas:
Good catch, I didn't notice the in-situ one because I hadn't tested on data. That's changed as well in the last commit, as well as autoconfiguring the MC type if the user doesn't do it manually.

kalderon:
I'd like to second the change to make forceSmear to default to false, that also broke things for my students and me and took some tracking down.

balunas:
Are we waiting for anything before we merge this? I think we're in agreement now about the changes.

mattleblanc:
Hi, I strongly disagree with changing the defaults away from the standards for R=0.4 jets. The tool should be configured to work for these as out-of-the-box as possible, as it is the most common use case. The default is to apply the nominal JER smearing and the in situ calibration to data for these jets.

If this breaks your large-R jets, it is possible to disable things by hand. It should be obvious from the output of configuring the tool that something isn't set up properly. Trying to apply jet cleaning to large-R jets already breaks it, because it's not applicable: this isn't any different in my opinion.

kratsg:
I clicked too soon before seeing @mattleblanc's message. My bad.

balunas:
Hi @mattleblanc - The way I've done it here is to keep the default behavior as it was before if the user doesn't specify anything in the config (that is, it'll be "MC16" for fullsim and "AFII" for fastsim). I've just made it configurable again in case someone wants to use something other than the standard R=0.4. That was Tadej's point as well (and I agree with you both), which is why I changed it to that in the second commit.

balunas:
Actually I think I misunderstood; I thought you were referring to `m_uncertMCType`. Wouldn't the correct in-situ calibration or smearing be correctly applied anyway if the user just uses the calibration sequence you guys recommend? Why should in-situ or smearing be different from, say, the GSC? It would be very confusing if xAH always tried to apply GSC even if you didn't include it in your calibration sequence, unless you manually set some secret flag which you need to read the source code to know about.

PR #1280, opened 2018-11-18
Add more pile-up info

ntadej:
This adds more pile-up information for data: corrected + corrected and scaled values of both average and actual mu.

One branch was renamed for clarity. As MC has now 6 branches that are the same, should we just drop 4 or 5 of them to save space (I guess for per-event variables it's not that critical)?


kratsg:
> As MC has now 6 branches that are the same, should we just drop 4 or 5 of them to save space (I guess for per-event variables it's not that critical)?

Which are the same? I suppose you just want to drop them for MC and not for data? That would make things harder to analyze -- since you usually want the same output structure for both MC and data.

ntadej:
MC has all pileUp branches the same, but yes, I agree we should keep the same structure for data and MC.

ntadej:
@kratsg, could this be merged now?

PR #1281, opened 2018-11-19
Adjust default jet calibration behavior, allow uncertainty configuration

balunas:
The last pull request for this (#1275) was reverted (#1277) after being merged, so I believe this new one is required to continue the discussion and get the changes back in. To recap, these changes are to `JetCalibrator` and are:
 - Allow the user to configure `m_uncertMCType` (which is required for fat jets, for example). If the user does not set this in their config, the default behavior is the same as it was before (`"MC16"` for fullsim and `"AFII"` for fastsim). This change is uncontroversial, since it just adds another option which the user is free to ignore.
 - Change the default values of `m_forceSmear` and `m_forceInsitu` to false. The smearing and in-situ calibration steps are (or should be, according to Jet/Etmiss recommendations) already included in the CalibSequence specified by the user. The purpose of this change is to avoid cases (of which at least 2 have already happened) of users explicitly excluding this step from their CalibSequence but the tool still mysteriously tries to apply it. Currently there's no way of knowing how to disable smearing/in-situ without reading the source code.

The only objection was from @mattleblanc, who made the point (which I agree with) that it should work out-of-the-box for R=0.4 EMTopo jets. But even in this case, the user has to specify a CalibSequence anyway, and if they use the recommended ones from the CP group, it will already include the smearing/in-situ step.

Personally I care more about making `m_uncertMCType` configurable because that's actually preventing my analysis code from working at the moment, so I'm OK with it if we insist on not changing the smearing/in-situ thing. But I do think it's confusing to have what are effectively manual overrides that default to "on" and are hidden from the user.

### DIFF
--- a/xAODAnaHelpers/JetCalibrator.h
+++ b/xAODAnaHelpers/JetCalibrator.h
@@ -69,7 +69,7 @@ public:
   /// @brief config for Jet Uncertainty Tool
   std::string m_uncertConfig = "";
   /// @brief MC type for Jet Uncertainty Tool
-  std::string m_uncertMCType = "MC16";
+  std::string m_uncertMCType = "";
   /// @brief Override CalibArea tag (default recommended)
   std::string m_overrideCalibArea = "";
   /// @brief Override uncertainties CalibArea tag (default recommended)
@@ -89,7 +89,7 @@ public:
   @endrst */
   bool m_setAFII = false;
   /// @brief when running data "_Insitu" is appended to calibration sequence
-  bool m_forceInsitu = true;
+  bool m_forceInsitu = false;
   /// @brief when running FullSim "_Smear" is appended to calibration sequence
   bool m_forceSmear = false;
   /// @brief when using DEV mode of JetCalibTools
--- a/xAODAnaHelpers/JetCalibrator.h
+++ b/xAODAnaHelpers/JetCalibrator.h
@@ -64,8 +64,8 @@ public:
   std::string m_calibConfigFullSim = "JES_data2017_2016_2015_Recommendation_Aug2018_rel21.config";
   /// @brief config for JetCalibrationTool for AFII MC
   std::string m_calibConfigAFII = "JES_MC16Recommendation_AFII_EMTopo_April2018_rel21.config";
-  /// @brief List of calibration steps. "Insitu" added automatically if running on data and "Smear" if running on FullSim MC
-  std::string m_calibSequence = "JetArea_Residual_EtaJES_GSC";
+  /// @brief List of calibration steps. Auto-configured to the Jet/Etmiss recommendation if left blank.
+  std::string m_calibSequence = "";
   /// @brief config for Jet Uncertainty Tool
   std::string m_uncertConfig = "";
   /// @brief MC type for Jet Uncertainty Tool
--- a/xAODAnaHelpers/JetCalibrator.h
+++ b/xAODAnaHelpers/JetCalibrator.h
@@ -68,6 +68,8 @@ public:
   std::string m_calibSequence = "JetArea_Residual_EtaJES_GSC";
   /// @brief config for Jet Uncertainty Tool
   std::string m_uncertConfig = "";
+  /// @brief MC type for Jet Uncertainty Tool
+  std::string m_uncertMCType = "MC16";
   /// @brief Override CalibArea tag (default recommended)
   std::string m_overrideCalibArea = "";
   /// @brief Override uncertainties CalibArea tag (default recommended)
@@ -89,7 +91,7 @@ public:
   /// @brief when running data "_Insitu" is appended to calibration sequence
   bool m_forceInsitu = true;
   /// @brief when running FullSim "_Smear" is appended to calibration sequence
-  bool m_forceSmear = true;
+  bool m_forceSmear = false;
   /// @brief when using DEV mode of JetCalibTools
   bool m_jetCalibToolsDEV = false;
 
@@ -136,7 +138,6 @@ private:
   bool m_isFullSim;       //!
 
   std::string m_calibConfig; //!
-  std::string m_uncertMCType; //!
 
   std::vector<CP::SystematicSet> m_systList; //!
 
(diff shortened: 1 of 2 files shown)


kratsg:
@mattleblanc

ntadej:
This is actually a mistake on my side when I was updating `JetCalibrator`. Originally I intended to drop one of the `m_calibConfigData` and `m_calibConfigFullSim` as they are basically the same (with the exception of those suffixes that were autoconfigured).

Anyway, I'm indifferent to this change.

jdandoy:
Just to jump in, historically we added "_Insitu" to the jet calib sequence as you should not run on data without it, unless you are an expert and explicitly set the flag to keep it off.  Then users could run with a single jet calib sequence and it would be edited to meet the default needs.  The same would now be true for MC with the _Smear addition.

I think we want to at least keep the functionality where if the R=0.4 user does not specify a m_calibSequence, then they get by default JetArea_Residual_EtaJES_GSC_Insitu for Data and JetArea_Residual_EtaJES_GSC_Smear for MC.  There's probably a couple ways to do this, but they all require some special action by expert users to not get the full string.  

Unfortunately it is no longer the case that the official MC m_calibSequence is a substring of the official one for data, which means the code for expert's will break as they can't just take the default.

It also sounds like part of the issue is the default for R=0.4 is not the default for R=1.0, which means we're calling any use of R=1.0 jets "expert" functionality.  Maybe we can gleam the jet type from the config file and determine the default that way.

Jeff

kratsg:
> It also sounds like part of the issue is the default for R=0.4 is not the default for R=1.0, which means we're calling any use of R=1.0 jets "expert" functionality. Maybe we can gleam the jet type from the config file and determine the default that way.

Some combination of case/switch with jet type + data-type should get you all the way there? As well as additionally doing something ala SUSYTools where the user specifies the string they want, and xAH deletes the portions that obviously don't make sense for the jet type or data type? (e.g. if Insitu is specified, delete it if running on MC).

mswiatlo:
Ok, so are we converging that we actually want this "auto-configured" in the C++? Arguably one can do write an auto-configuration in the python configuration file already, but I'd rather we just settle on a course and get this done so we can move forward.

mattleblanc:
Hi, I'm back -- sorry for being a pain! I've been thinking about this a bit since the last MR.

My main objection is that I don't think that xAH should behave differently than the CP tool does. 

Personally, I also find the automatic appending of these components (`_smear` and `_Insitu`) of the CalibString to also be shady -- especially if the user has explicitly set what they thought was the CalibString by hand. So I understand why you want to have those settings disabled by default in xAH. 

I maintain that we should mirror the normal `JetCalibTool` behaviour here, but I would encourage you to raise this point on the JSV mailing list, and I will raise it with S&C as well. I believe the rationale was originally to protect users from forgetting that the config should be different for data & MC, but perhaps we should think again about whether we do things under their noses.

In the context of large-R jets, I guess having either of these things appended will cause problems. It might be good to see if we can veto it on our end for this case. I would maybe just check the input collection for the substring `AntiKt4` or something similar (we only have a JER & insitu for those jets right now), but that's probably too crude for @kratsg's liking.

MLB

ntadej:
To add to @mattleblanc's comments: I really like the recent philosophy changes towards "users should just use defaults of the tool". I we can propose at the P&P or S&C that Jet tools get updated in this way, it would be really nice. Ideally user would just set the type of jets and "calibration release", everything else should be configured by the tool.

mswiatlo:
Ok, so I see two votes for "auto-configure" from @ntadej and @cohm and two votes for "pass the string cleanly" from @balunas and @mattleblanc, is that right? I agree Matt in principle we should update the official tool to be kinder to use, but since you're responsible for this on the ATLAS side I'll let you get that started, and in the meantime we can develop an auto-configuration like SUSYTools does?

Another option is that we have doJetAutoconfig as an option, which could be true/false by default as desired...

jbossios:
As long as we keep the expert configuration option available (i.e. no autoconfiguration) it's fine with me. I use xAH for other situations where the autoconfiguration will not work (R-scan, HLLHC, HLT jets).

balunas:
Alright, how about this proposal:
 - If the user sets m_calibSequence manually, it gets sent to the CP tool exactly as-is.
 - If the user doesn't set m_calibSequence, we have it autoconfigure to the default recommended string based on the input collection name and whether it's data or MC.
- If the user doesn't set m_calibSequence and the input collection name is unrecognized, we throw an error and exit.

The code already throws an error if you try to apply smearing to data or in-situ to MC, so it's already impossible for people to accidentally forget and use the wrong one. Nobody should be manually passing in a partial calibSequence and expecting to get the full calibration anyway.

mswiatlo:
Love the idea, @balunas . We'd have to detect if it's AFII or not as well (smear is only applied to fullsim). But I think your proposal is clear and provides good defaults, while leaving expert options.

mattleblanc:
Sounds good to me!

jdandoy:
Sounds great to me too!  It would also would be good to include warnings for non-standard cases but allow them, such as not including Insitu but running on data.  There may be some previous users who set the calib sequence to the default in their config but depend on the automatic addition of Insitu, or new users not yet aware of the differences, but a loud warning should be sufficient.

balunas:
Alright, I just implemented the auto-configuration (and the warnings Jeff suggested) in this last commit, give it a look, and hopefully everyone should be happy with it now.

xAODAnaHelpers/JetCalibrator.h
```diff
@@ -68,6 +68,8 @@ class JetCalibrator : public xAH::Algorithm
   std::string m_calibSequence = "JetArea_Residual_EtaJES_GSC";
```
jbossios: @balunas I think it should be empty m_calibSequence by default. Right?

balunas: Good catch, you're right about that. Fixed.

mswiatlo:
I think this looks good! Can we get this merged please? We need it for a new production shortly, if possible.

PR #1282, opened 2018-11-23
Update Root/BJetEfficiencyCorrector.cxx: add/remove new/old taggers

vrpascuzzi:
Taggers have been updated:
https://gitlab.cern.ch/atlas/athena/commit/e36de3cc8d4b65e065e3f09c3e45cec64bd64442
and use of deprecated ones now result in `StatusCode::FAILURE`.

This commit removes from the list of "known" `xAH` taggers and adds the
new ones. Also added was the tagger/WP combination that is unknown to `xAH`, which I think is useful.

N.B. as of the date of this commit, weights are not
available in old(er) derivations for the new taggers!

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -118,15 +118,17 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
   if (m_operatingPt == "HybBEff_85")  { opOK = true; m_getScaleFactors =  true; }
 
   // Only DL1 and MV2c10 are calibrated
-  if (m_taggerName == "MV2c10")    { taggerOK = true; m_getScaleFactors =  true; }
-  if (m_taggerName == "MV2c10rnn") { taggerOK = true; m_getScaleFactors =  false; }
-  if (m_taggerName == "MV2c10mu")  { taggerOK = true; m_getScaleFactors =  false; }
-  if (m_taggerName == "DL1")       { taggerOK = true; m_getScaleFactors =  true; }
-  if (m_taggerName == "DL1rnn")    { taggerOK = true; m_getScaleFactors =  false; }
-  if (m_taggerName == "DL1mu")     { taggerOK = true; m_getScaleFactors =  false; }
+  // Available taggers (>=21.2.51):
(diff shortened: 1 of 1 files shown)


mswiatlo:
Because this PR breaks compatibility on older derivations I'm not sure what the best way to proceed is. My guess is that we should maintain backwards compatibility with a configuration flag that changes the supported modes (in principle this can be auto-detected with the p-tag metadata?). This is definitely a useful update but I think some more protection for older datasets would be necessary, at least for a while.

vpascuzz:
Hi, @mswiatlo.
Good point. I think using an automated mechanism would be a nice solution to maintain backwards compatibility.
I wonder what @kratsg thinks.

kratsg:
> Good point. I think using an automated mechanism would be a nice solution to maintain backwards compatibility.
> I wonder what @kratsg thinks.

There's no easy way here. I've spent some time thinking about this, but I don't have a good answer how to handle the older derivations. This fix only enables/disables boolean flags for which taggers are ok or not. We can just support the older ones by just not removing them, and only adding new ones.. but this still gets to the problem that we don't have a trivial way to look up availability of information corresponding to a given p-tag.

vpascuzz:
What about an additional override in the config? Something like:
```
bool m_ptagOverride
```
Yes, I see this undoubtedly opening a whole new can of worms but could also be useful support-wise.

kkrizka:
The `BTaggingSelectionTool` tool already checks what taggers are supported by it. The only advantage to including the tagger check is determing whether scale factors should be calculated. That is a more stable check, since the supported taggers don't change often.

I propose we remove the failure if an "unsupported" tagger is requested and leave the tool to fail. Then apply scale factors only for MV2c10 and DL1, and add a `m_forceScaleFactors` flag for other taggers.

mswiatlo:
Hi,

What's the status here, @vpascuzz ? If you'd like, some of us from hh4b can help out on this as well.

Best,
Max

vpascuzz:
Hi, @mswiatlo.
Being no expert on b-taggers, I'm happy to adopt someone else's, e.g. @kkrizka's, solution. Or just close this PR altogether until there's better support for new taggers.

kratsg:
I would suggest @kkrizka 's solution if someone can implement it.

kratsg:
Any updates guys?

mswiatlo:
Not yet, but I intend to work on this soon.

PR #1287, opened 2018-11-30
BasicEventSelection: Allow m_storeTrigDecisions even if no trigger selection is specified

lawrenceleejr:
In BasicEventSelection, it would be great to be able to get the list of passed triggers (if `m_storeTrigDecision` is true) written to the store even if there is no trigger selection is specified.

Is there any reason to not make the proposed change here?

Thanks!
-Larry

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -879,7 +879,7 @@ EL::StatusCode BasicEventSelection :: execute ()
   // Trigger decision cut
   //---------------------
 
-  if ( !m_triggerSelection.empty() ) {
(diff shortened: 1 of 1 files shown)


mswiatlo:
Won't this increase the size considerably for people who didn't ask for trigger decisions?

lawrenceleejr:
Not if you set `m_storeTrigDecision` to false, right? If you really want nothing to do with the trigger, then just make sure they're both set to false and nothing will change? Also the "storing" aspect is just putting it in the store, so this would be a change to the size of the TStore and not anything else, yeah?

mswiatlo:
Fair enough~

PR #1293, opened 2018-12-05
HelpTreeBase: Allow non-default object names for user-defined branches

lschaefer:
This commit adds functionality to allow for a non-default name for object containers, like those that can already be set with the Add\[Objects\](...) functions. 

So for example if the user wanted two separate muon containers, they could create them by calling AddMuons(m_muDetailStr, "MuonContainer1") and AddMuons(m_muDetailStr, "MuonContainer2"). The standard branches (those filled by HelpTreeBase) would have names corresponding to each container. But the user-added branches would still be restricted to the standard "muon_" branch name. This commit corrects that, so now AddMuonsUser can also pass the object name.

Additionally, I noticed that AddTruthParts and AddTrackParts (and the corresponding User functions) are inconsistent with the rest of the Add(User) functions in the ordering of the "name" and "detailStr" arguments, but I didn't change this to avoid incompatibilities. But it could certainly lead to confusion on the user's side.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -413,6 +413,7 @@ EL::StatusCode BJetEfficiencyCorrector :: executeEfficiencyCorrection(const xAOD
     dec_sfBTag( *jet_itr ) = sfVec;
   }
 
+
   //
   // Store list of available systematics
   //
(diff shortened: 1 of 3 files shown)


kratsg:
You should definitely clean up your fork or re-update so you don't have a bunch of spurious commits.

lschaefer:
Hi Giordon,
Thanks. I know, this is really annoying. I struggled with this every time, although my fork is up to date, somehow the old commits are left lying around. With @jdandoy's help I eventually used commands similar to what's here https://stackoverflow.com/questions/41283955/github-keeps-saying-this-branch-is-x-commits-ahead-y-commits-behind. Maybe the documentation could be updated to show something similar to this, if I'm not the only one who had this problem, or at least add a few lines like "if you are seeing this problem then do these steps"?
Thanks!

PR #1295, opened 2018-12-07
BasicEventSelection::finalize(): Remove TDT from ToolStore

lawrenceleejr:
Simply calling finalize of the TDT does some nasty things under the hood in the TDT itself. So say if you have two algs (say, `xAH` and `SUSYTools` -- don't hate. Let's be friends.) grabbing the TDT from the tool store, if they both try to finalize, it barfs a bit.

```
ToolSvc.TrigDecisionTool ERROR   /build2/atnight/localbuilds/nightlies/21.2/AnalysisBase/athena/Trigger/TrigAnalysis/TrigDecisionTool/Root/TrigDecisionTool.cxx:246 (virtual StatusCode Trig::TrigDecisionTool::finalize()): could not find instance name in instance list, but must have been added in ::initialize(). Name: ToolSvc.TrigDecisionTool
```

The proposed change from Hass ensures that after calling `finalize()` on the TDT, it's removed from the `ToolStore` so that nobody else can try to call `finalize()` again.

Just tested on my side and we no longer see this error with this change! Yay.

-L

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -1269,7 +1269,12 @@ EL::StatusCode BasicEventSelection :: finalize ()
 
   m_RunNr_VS_EvtNr.clear();
 
-  if ( m_trigDecTool_handle.isInitialized() )  m_trigDecTool_handle->finalize();
+  if ( m_trigDecTool_handle.isInitialized() ){
+    if (asg::ToolStore::contains<Trig::TrigDecisionTool>("ToolSvc.TrigDecisionTool") ){
+      m_trigDecTool_handle->finalize();
+      asg::ToolStore::remove("ToolSvc.TrigDecisionTool").ignore();
+    }
+  }
 
   //after execution loop
   if(m_printBranchList){


ntadej:
I just wonder, would this affect downstream xAH algorithms using TDT (just thinking out loud)?

lawrenceleejr:
I suppose only if they want to do something to the TDT in their finalize calls that come after `BasicEventSelection::finalize()`, right? Which presumably doesn't happen? Maybe it does, so please check me on this! 

-L

kratsg:
> I suppose only if they want to do something to the TDT in their finalize calls that come after `BasicEventSelection::finalize()`, right? Which presumably doesn't happen? Maybe it does, so please check me on this!

This shouldn't happen. Unfortunately, I think we're going to have to both patch it in xAH and loudly complain to ASG. This is clearly a problem with the tool design because tools can be retrieved multiple times and should simply be finalized multiple times. For some reason, it's not doing that...

Additionally, I see why SUSYTools doesn't have the capability of doing it naively (but it does! if it inherits from AsgMetaDataTool, it can register a callback/hook that runs when the `finalize()` occurs, but that's a lot of work). I think I'm ok with this, provided we complain upstream.... annoyingly, we're supposed to avoid `asg::toolstore::contains<>` calls... So.... hrmm..

Yeah, this is meant to be a shared_ptr and should only go away when all references are removed...

PR #1298, opened 2018-12-27
Adding jet cleaning and isBadBatman cleaning to BES

### LINKED ISSUE #449: Do jet cleaning without calibrating jets
I would like to decouple jet calibration and jet cleaning, for trigger jet purposes. Anyone else would like to have this functionality? If not, I may just wait until we have a calibration for trigger jets and modify JetCalibrator only.

### LINKED ISSUE #1283: Event-level jet cleaning flags
We should implement the ability to use the event-level jet cleaning flags in the `BasicEventSelection`. These `eventInfo` flags, specifically:

```
  |       |     &->getAttribute<Char_t>("DFCommonJets_eventClean_LooseBad")
  |       |     &->getAttribute<Char_t>("DFCommonJets_isBadBatman")
```

which will allow people using jets other than EMTopo jets for cleaning. Currently, you would technically need to run two jet calibration algorithms to clean jets if you wanted to use PFlow for your analysis, as the jet cleaning is only defined for EMTopo.

Might as well do the BadBatMan / Frog-Eyes cleaning while we're at it.

This will address an old issue from @urania277, #449 

I'll get this in over the weekend.

:beers: MLB

mattleblanc:
I finally got around to it! This seems to be working, as the flags removes a small handful of events in my local tests. A reminder of the recommendations --

* The event-level jet cleaning flag is the recommended way to clean *all* jets, but especially collections other than EMTopo.

* The isBadBatman cleaning is an optional flag which should only ever be used in 2015 and 2016 data, for analyses which may be of interest for analyses where fake MET can be an issue. It will remove 1.66% of the 2016 data and a small amount from 2015.

See also https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/HowToCleanJets2017

I'm not actually sure how to get things into the cutflow histograms properly, @kratsg feel free to change or tell me what to do if you want that done differently.

This closes #1283, and I would say also closes #449 and #412  ... but I would maybe confirm with the original people who opened those issues first.

🍻 MLB

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -412,6 +412,10 @@ EL::StatusCode BasicEventSelection :: initialize ()
       m_cutflow_grl  = m_cutflowHist->GetXaxis()->FindBin("GRL");
       m_cutflowHistW->GetXaxis()->FindBin("GRL");
     }
+    if ( m_applyIsBadBatmanFlag ) {
+      m_cutflow_isbadbatman =  m_cutflowHist->GetXaxis()->FindBin("IsBadBatman");
+      m_cutflowHistW->GetXaxis()->FindBin("IsBadBatman");
+    }
     m_cutflow_lar  = m_cutflowHist->GetXaxis()->FindBin("LAr");
     m_cutflowHistW->GetXaxis()->FindBin("LAr");
     m_cutflow_tile = m_cutflowHist->GetXaxis()->FindBin("tile");
@@ -427,7 +431,10 @@ EL::StatusCode BasicEventSelection :: initialize ()
     m_cutflow_trigger  = m_cutflowHist->GetXaxis()->FindBin("Trigger");
     m_cutflowHistW->GetXaxis()->FindBin("Trigger");
   }
-
+  if ( m_applyJetCleaningEventFlag ) {
+    m_cutflow_jetcleaning = m_cutflowHist->GetXaxis()->FindBin("JetCleaning");
+    m_cutflowHistW->GetXaxis()->FindBin("JetCleaning");
+  }
 
   ANA_MSG_INFO( "Histograms set up!");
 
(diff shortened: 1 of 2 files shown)


xAODAnaHelpers/BasicEventSelection.h
```diff
@@ -129,6 +129,10 @@ class BasicEventSelection : public xAH::Algorithm
     bool m_applyEventCleaningCut = false;
     bool m_applyCoreFlagsCut = false;
 
+    // Jet Cleaning
```
kratsg: definitely add more documentation for what these flags are doing (e.g. based on the comment in the PR) e.g.

```
// Jet Cleaning (see also https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/HowToCleanJets2017)
/// recommended way to clean all jets, but especially collections other than EMTopo
bool m_applyJetCleaningEventFlag = false;
/// should only ever be used in 2015 and 2016 data, for analyses which may be of interest for analyses where fake MET can be an issue
bool m_applyIsBadBatmanFlag = false;
```

kratsg:
> I'm not actually sure how to get things into the cutflow histograms properly, @kratsg feel free to change or tell me what to do if you want that done differently.

This looks like you're adding in the cutflow histogram correctly?

xAODAnaHelpers/BasicEventSelection.h
mattleblanc: Done!

mattleblanc:
> > I'm not actually sure how to get things into the cutflow histograms properly, @kratsg feel free to change or tell me what to do if you want that done differently.
> 
> This looks like you're adding in the cutflow histogram correctly?

Yes, but the bins seem to be out of order? Or maybe they are not ordered in a particular way?

Root/BasicEventSelection.cxx
```diff
@@ -858,6 +865,27 @@ EL::StatusCode BasicEventSelection :: execute ()
 
   }
 
+  // more info: https://twiki.cern.ch/twiki/bin/viewauth/AtlasProtected/HowToCleanJets2017
+  if ( m_applyJetCleaningEventFlag && eventInfo->isAvailable<Char_t>("DFCommonJets_eventClean_LooseBad") ) {
+    if(eventInfo->auxdataConst<Char_t>("DFCommonJets_eventClean_LooseBad")<1) {
```
ntadej: A naive question, but do we still need to use deprecated ROOT types like `Char_t`?

mattleblanc: This is the type of the attribute on the `eventInfo`, so it's what I used --

```
http://atlas-computing.web.cern.ch/atlas-computing/links/nightlyDocDirectory/xAODEventInfo/html/
xAOD::EventInfo
  |     xAOD::EventInfo* EventInfo
  |       |     &->getAttribute<Char_t>("DFCommonJets_eventClean_LooseBad")
  |       |     &->getAttribute<Char_t>("DFCommonJets_isBadBatman")
```

ntadej: Sure, but `Char_t` is now just a typedef to `char`. You can have it either way, I just wanted to bring up, that we should probably try to move away from custom ROOT types if possible.

kratsg: I agree with @ntadej , we should use `char` and not `Char_t`.

kratsg:
> Yes, but the bins seem to be out of order? Or maybe they are not ordered in a particular way?

They're not usually ordered in a particular way. Because users can have algorithms in different orders, it's up to them to extract out the order they want by bin name.

mattleblanc:
pis j'ai parker mon char a la source code

PR #1299, opened 2019-01-02
WIP: Use electron LooseBL information from derivations

ntadej:
Electron `LooseBL` information is now available in derivations. I propose we remove the hardcoded override as I observed some differences between the current information and information from derivations.

I kept `m_doBLTrackQualityCut` and it should be probably made configurable.

### DIFF
--- a/Root/ElectronSelector.cxx
+++ b/Root/ElectronSelector.cxx
@@ -279,12 +279,6 @@ EL::StatusCode ElectronSelector :: initialize ()
     }
 
     if ( m_readIDFlagsFromDerivation ) {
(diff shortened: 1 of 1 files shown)


kratsg:
This is still a WIP?

ntadej:
If you do not want any extra configuration support, then this is not WIP anymore.

kratsg:
> If you do not want any extra configuration support, then this is not WIP anymore.

I wonder if this will break on older derivations? But then I suppose people should just set `m_LHOperatingPoint = Loose`?

ntadej:
Probably ... will make it configurable.

PR #1301, opened 2019-01-04
Remove MsgStream name growth

mswiatlo:
Alex Melzer found this but can't get a GitHub setup, so I'm making the PR for him.

msg() returns a static MsgStream object for messaging. This means that every time getPrimaryVertex( ... msg() ) was being called, ".getPrimaryVertex" was being appended to the static object's name. When running over enough events (easy to happen when you run with systematics), there would be a huge amount of memory and time spent around managing this single gigantic string. This PR fixes the three instances of this stream name change in HelperFunctions.

We found that without this, time to finish an algorithm was exponentially dependent on the number of events. After the fix, we see a linear dependence, as expected.

There are probably alternate fixes for this (changing how msg() works?) but I think this is the simplest and cleanest solution.


cohm:
Whoa. Nice catch!

PR #1302, opened 2019-01-10
Random updates

kkrizka:
Random updates I had laying around. Mostly updating code to latest recommendations.

* Remove JER smearing tool inclues from JetCalibrator since they are no longer used.
* Change autoPRW location to the new common PileupReweighting one.
* Update list of supported taggers in BJetEfficiencyCorrector and improve error output.
* Speed up particle loading in `ParticleContainer` by reusing particle objects in memory.
 * Add `ParticleContainer::particles()` for direct access to the particles list.

All should be backward compatible. The removed btaggers will fail in the btagging tool anways. Maybe we should remove this check inside xAH anyway, since it is also done in `BTaggingSelectionTool`?


kratsg:
> All should be backward compatible. The removed btaggers will fail in the btagging tool anways. Maybe we should remove this check inside xAH anyway, since it is also done in `BTaggingSelectionTool`?

Yes. Let's have another PR for this.

PR #1303, opened 2019-01-10
Per-user defaults for xAH_run (aka a dotfile)

kkrizka:
To run on my institution's cluster, I need to specify a lot of options (ie: queue, setup commands...) to xAH_run.py. This is not very practical if I want to run one-off tests.

This PR solves this by adding a dotfile (`~/.xah`) that is read by `xAH_run.py` at start and overrides any defaults passed to the argument parser. It is an INI file with a section per sub-command and a `general` section for the general options. The keys are the options being set.

As an example, my .xah looks as follows:
```
[slurm]
optBatchSharedFileSystem=1
optBatchWait=1
optSlurmRunTime=5:00:00
optSlurmExtraConfigLines=#SBATCH --qos=shared --tasks-per-node=1 --constraint=haswell --image=centos:centos7 --export=NONE
optSlurmWrapperExec=export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/global/project/projectdirs/atlas/scripts/extra_libs_180822; hostname; shifter --module=cvmfs /bin/bash \
```

### DIFF
--- a/docs/UsingUs.rst
+++ b/docs/UsingUs.rst
@@ -171,6 +171,8 @@ API Reference
 
     eval "$(register-python-argcomplete xAH_run.py)"
 
+.. note:: All of the following properties can be set in a user-specific dotfile located at ``${HOME}/.xah``. It is a simple ``key=value`` per-line format, where the `key` is the argument being set without the preceeding dashes. There is no separation of the arguments to the different drivers.
+
 
 .. argparse::
    :ref: xAH_run.parser
--- a/docs/UsingUs.rst
+++ b/docs/UsingUs.rst
@@ -171,8 +171,20 @@ API Reference
 
     eval "$(register-python-argcomplete xAH_run.py)"
 
-.. note:: All of the following properties can be set in a user-specific dotfile located at ``${HOME}/.xah``. It is a simple ``key=value`` per-line format, where the `key` is the argument being set without the preceeding dashes. There is no separation of the arguments to the different drivers. All lines starting with # are ignored.
+All of the following properties can be set in a user-specific dotfile located at ``${HOME}/.xah``. It is an `INI file <https://en.wikipedia.org/wiki/INI_file>`_, with the `general` section used for the generic options and other sections named after sub-commands. The keys in each section are the options without the preceeding dashes.
 
+The following example configures the Slurm driver for NERCS' Cori and records usage statistics:
+::
+   
+   [general]
+   stats=1
+
+   [slurm]
+   optBatchSharedFileSystem=1
+   optBatchWait=1
+   optSlurmRunTime=5:00:00
+   optSlurmExtraConfigLines=#SBATCH --qos=shared --tasks-per-node=1 --constraint=haswell --image=centos:centos7 --export=NONE
+   optSlurmWrapperExec=export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/global/project/projectdirs/atlas/scripts/extra_libs_180822; hostname; shifter --module=cvmfs /bin/
 
 .. argparse::
    :ref: xAH_run.parser
--- a/docs/UsingUs.rst
+++ b/docs/UsingUs.rst
@@ -171,7 +171,7 @@ API Reference
 
     eval "$(register-python-argcomplete xAH_run.py)"
 
-.. note:: All of the following properties can be set in a user-specific dotfile located at ``${HOME}/.xah``. It is a simple ``key=value`` per-line format, where the `key` is the argument being set without the preceeding dashes. There is no separation of the arguments to the different drivers.
+.. note:: All of the following properties can be set in a user-specific dotfile located at ``${HOME}/.xah``. It is a simple ``key=value`` per-line format, where the `key` is the argument being set without the preceeding dashes. There is no separation of the arguments to the different drivers. All lines starting with # are ignored.
 
 
 .. argparse::
--- a/python/cli_options.py
+++ b/python/cli_options.py
@@ -4,6 +4,16 @@ from __future__ import absolute_import
 from __future__ import print_function
 import logging
 
+#
+# Update the default fields of an argument definition dictionary
+#
+# argdict: reference to the argument definitions
+# newvalues: dictionary with the argument name as key and new default value as value
+def update_defaults(argdict, newvalues):
+    for option,optdata in argdict.iteritems():
+        if option in newvalues: optdata['default']=newvalues[option]
+
+
 logger = logging.getLogger("xAH.cli_options")
 
 
@@ -327,7 +337,7 @@ drivers_slurm = {
         "metavar": "",
         "type": str,
         "required": False,
-        "default": "shared-chos",
+        "default": "",
         "help": "the name of the partition to use",
     },
     "optSlurmRunTime": {
@@ -341,7 +351,7 @@ drivers_slurm = {
         "metavar": "",
         "type": str,
         "required": False,
-        "default": "1800",
+        "default": "",
         "help": "the maximum memory usage of the job (MB)",
     },
     "optSlurmConstrain": {
--- a/python/cli_options.py
+++ b/python/cli_options.py
@@ -193,6 +193,10 @@ drivers_common = {
     },
 }
 
+# define arguments for direct driver
+drivers_direct = {}
+drivers_direct.update(copy.deepcopy(drivers_common))
+
 # define arguments for prooflite driver
 drivers_prooflite = {}
 drivers_prooflite.update(copy.deepcopy(drivers_common))
--- a/python/cli_options.py
+++ b/python/cli_options.py
@@ -4,16 +4,6 @@ from __future__ import absolute_import
 from __future__ import print_function
 import logging
 
-#
-# Update the default fields of an argument definition dictionary
-#
-# argdict: reference to the argument definitions
-# newvalues: dictionary with the argument name as key and new default value as value
-def update_defaults(argdict, newvalues):
-    for option,optdata in argdict.iteritems():
-        if option in newvalues: optdata['default']=newvalues[option]
-
-
 logger = logging.getLogger("xAH.cli_options")
 
 
--- a/python/cli_options.py
+++ b/python/cli_options.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import
 from __future__ import print_function
 import logging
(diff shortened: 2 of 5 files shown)


kratsg:
don't move the `import ROOT` because it's how we can build the documentation.

scripts/xAH_run.py
```diff
@@ -17,11 +17,33 @@
 import argparse
 try: import argcomplete
 except: pass
-import os
+import os, os.path
```
kratsg: this change isn't needed. you already have `os.path`

kratsg:
@kkrizka , I've been thinking a little about this, and I want to take the opportunity to make this easier for you, actually. As it's been pretty clear, more especially so in this PR, the configurations available (their types, the default values, etc...) should be maintained in some particular list that you can directly access or use to automatically generate the dot-file config as well.

Let me submit a PR to rearrange some of this for you, so that this is a bit easier to deal with.

-------

On a second note, I would encourage you to use the *exact* name of the argument in the config file as well, instead of using a different name, e.g. `optSlurmAccount` instead of `Account`. This makes grep'ing much easier for users to do.

kratsg:
When #1304 is merged in, you should be able to do something like this

```python
config.add_section('slurm')
for optName,optConfig in xAODAnaHelpers.cli_options.drivers_slurm.iteritems():
  config.set('slurm', optName, optConfig.get('default'))

 config.read(userconfigpath)

# update the cli_options based on the read-in config file
for optName in xAODAnaHelpers.cli_options.drivers_slurm.keys():
  xAODAnaHelpers.cli_options.drivers_slurm[optName]['default'] = config.get('slurm', optName)
...
...
...
xAODAnaHelpers.utils.register_on_parser(xAODAnaHelpers.cli_options.drivers_slurm, slurm)
```

and then in portions when we build the argument parser, it should see the updated defaults.

kkrizka:
I prefer the dotfile. It is a standard way to configure applications on Linux.

kratsg:
> I prefer the dotfile. It is a standard way to configure applications on Linux.

This still uses the dotfile.

kkrizka:
Ok, I don't see any difference by moving all of the options to a separate file then. So go ahead.

kratsg:
> Ok, I don't see any difference by moving all of the options to a separate file then. So go ahead.

This way, you're not hard-coding the `config.set()` options twice -- once for argument parser, and one for config parser.

python/cli_options.py
```diff
@@ -4,6 +4,16 @@
 from __future__ import print_function
 import logging
 
+#
+# Update the default fields of an argument definition dictionary
+#
+# argdict: reference to the argument definitions
+# newvalues: dictionary with the argument name as key and new default value as value
+def update_defaults(argdict, newvalues):
```
kratsg: isn't `argdict.update(newvalues)` just enough? Why do we need a special function?

scripts/xAH_run.py
```diff
@@ -17,11 +17,12 @@
...
 import subprocess
 import sys
 import datetime
 import time
+import ConfigParser
```
kratsg: This will break in python3 FYI -- so we will need to think about how to deal with the switch.

scripts/xAH_run.py
```diff
@@ -35,6 +36,30 @@
...
+        dotconfig[key]=value
+
+# Apply the configuration defaults
+xAH_cli_options.update_defaults(xAH_cli_options.standard         , dotconfig)
+xAH_cli_options.update_defaults(xAH_cli_options.drivers_common   , dotconfig)
+xAH_cli_options.update_defaults(xAH_cli_options.drivers_prooflite, dotconfig)
+xAH_cli_options.update_defaults(xAH_cli_options.drivers_prun     , dotconfig)
+xAH_cli_options.update_defaults(xAH_cli_options.drivers_condor   , dotconfig)
+xAH_cli_options.update_defaults(xAH_cli_options.drivers_lsf      , dotconfig)
+xAH_cli_options.update_defaults(xAH_cli_options.drivers_slurm    , dotconfig)
```
kratsg: all of this could be written as...

(python2)
```python
cli_options = {}
cli_options.update(xAH_cli_options.standard)
cli_options.update(xAH_cli_options.drivers_common)
cli_options.update(xAH_cli_options.drivers_prooflite)
cli_options.update(xAH_cli_options.drivers_prun)
cli_options.update(xAH_cli_options.drivers_condor)
cli_options.update(xAH_cli_options.drivers_lsf)
cli_options.update(xAH_cli_options.drivers_slurm)
cli_options.update(dotconfig)
```

(python3)
```python
cli_options = {**xAH_cli_options.standard, 
               **xAH_cli_options.drivers_common,
               **xAH_cli_options.drivers_prooflite,
               **xAH_cli_options.drivers_prun,
               **xAH_cli_options.drivers_condor,
               **xAH_cli_options.drivers_lsf,
               **xAH_cli_options.drivers_slurm}
cli_options.update(dotconfig)
```

or something similar to that. Although I understand why you might want to update the defaults there instead, so you can just as easily do

```python
for c in ['standard', 'drivers_common', 'drivers_prooflite', 'drivers_prun', 'drivers_condor', 'drivers_lsf', 'drivers_slurm']:
  getattr(xAH_cli_options, c).update(dotconfig)
```

or something like iterating over everything inside the `xAH_cli_options` module.

kratsg:
Also, any thoughts on using the existing dotfiles from asetup -- and adding in configurations there -- instead of creating another dotfile `.xAH`? I suppose at the moment, there's no harm -- but one can imagine this would be useful for others to use as well in different frameworks and then we start seeing hundreds of different dotfiles...

scripts/xAH_run.py
kkrizka: The syntax no longer uses ConfigParser for simplicity. I forgot to remove the import.

scripts/xAH_run.py
kkrizka: Many ways to do things. I did one.

kkrizka:
The configuration is very specific to xAH. I don't see a reason to mix its configuration with asetup, which is a separate application.

kratsg:
Additional, how do we feel about using something like https://github.com/theskumar/python-dotenv? We can package it up with xAH as part of a CMake installation (and it's relatively small) and might be perfect for something like this? If it's too much, that's also fine.

kratsg:
> The configuration is very specific to xAH. I don't see a reason to mix its configuration with asetup, which is a separate application.

Right, so the reason I ask is -- for example -- pyproject.toml is becoming (in python3) a centralized, sort-of dotfile for various python packages like flake8 and pytest -- as well as to put setup.py specific things in it as well. One could imagine the same thing here, where xAH plays nicely with *other* frameworks that depend on xAH like the Xhh folks (so it's less confusing for them to have a ... `.atlas` or `.analysis` file with all configurations -- than a `.xah`).

Also I might request that no matter what we pick, we keep the filename lowercase.

scripts/xAH_run.py
kratsg: I'm still not convinced it needs a special function though.

kratsg: I now realize why. Never mind.

python/cli_options.py
kratsg: Never mind.

kratsg:
Can we move the parsing of dotfile into utils, and the updating of cli_options into utils as well? Then I think I'm happy with this. Main reason I ask is so that other people who depend on xAH (Xhh) can take advantage of the nice utility functions for parsing/updating.

kratsg:
And I think one last thought, and maybe a nice thing. I can imagine that you might want to have multiple `.xah` files -- one in `$HOME/.xah` and one in `$HOME/analysis_wrapper/.xah`. E.G. `$HOME/.xah` contains global settings specific to the machine, while the local one would have settings specific to the analysis. This still follows standard precedence of local overriding `$HOME` if they share the same keys. This would be, presumably, as simple as

```
home_dotfile.update(local_dotfile)
```

before updating anything else. Or just adding another line, so that the local dotfile updates all defaults last.

kkrizka:
I've moved the dotfile loading into a function in utils. That way if someone requests analysis-specific dotfiles, this can be added.

python/cli_options.py
kkrizka: No, the dictionaries have different formats.

python/utils.py
```diff
@@ -112,11 +112,31 @@ def register_on_parser(cli_options, parser):
...
+
+def read_dotfile(dotpath):
+  """Return the contents of dotpath configuration file as a dictionary."""
+
+  dotconfig={}
+  if os.path.exists(dotpath):
+    with open(dotpath) as fh :
+      for line in fh:
+        line=line.strip()
+        if line=='': continue
```
kratsg: `if line.empty() or line.startswith('#'): continue`

python/utils.py
```diff
@@ -112,11 +112,31 @@ def register_on_parser(cli_options, parser):
...
+        if line[0]=='#': continue
+        parts=line.split('=')
```
kratsg: parts = line.split('=', 1)

docs/UsingUs.rst
```diff
@@ -171,6 +171,8 @@ API Reference
 
     eval "$(register-python-argcomplete xAH_run.py)"
 
+.. note:: All of the following properties can be set in a user-specific dotfile located at ``${HOME}/.xah``. It is a simple ``key=value`` per-line format, where the `key` is the argument being set without the preceeding dashes. There is no separation of the arguments to the different drivers.
```
kratsg: mention that comments can be added to file with `#` and perhaps give a short example of a file.

kkrizka:
This PR does not cause #1316 , but the error does highlight an issue with the current proposal. There are certain "common" driver options that I don't want to have the same defaults for. For example, I want `optBatchDriver=True` for the Slurm driver (local batch, mostly for making histograms that I want automatically merged at the end) and I don't want it for grid (long jobs, make ntuples that I don't want downloaded/merged).

The PR has been changed back to use INI files, with section per sub-command and `general` for the general options. The documentation is also updated.

I'm going to keep it as WIP for a day or two to further test the latest changes.

kratsg:
This is fine. Remove the WIP when you're happy and ping me.

kkrizka:
I've been running with this for a while and I'm happy. So I think it can be merged.

PR #1305, opened 2019-01-14
Test compatibility with 21.2.56

beojan:
This PR is just to run the CI with AB 21.2.56.


ntadej:
I just want to point out that AFAIK only compilation is tested in the CI.

kratsg:
We can still add this version to list of version to be tested/automatically build docker images for.

beojan:
Nothing else actually needs to be done for that, right? I.e. you just have to merge this PR?

PR #1309, opened 2019-01-28
add flag to treat MC as pseudoData

rhankache:
This is needed when we want to propagate the JER uncertainties using MC which should be treated as a pseudo-Data

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -285,7 +285,12 @@ EL::StatusCode JetCalibrator :: initialize ()
     ANA_CHECK( ASG_MAKE_ANA_TOOL(m_JetUncertaintiesTool_handle, JetUncertaintiesTool));
     ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("JetDefinition",m_jetAlgo));
     ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("MCType",m_uncertMCType));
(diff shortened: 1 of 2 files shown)


Root/JetCalibrator.cxx
```diff
@@ -285,7 +285,12 @@ EL::StatusCode JetCalibrator :: initialize ()
     ANA_CHECK( ASG_MAKE_ANA_TOOL(m_JetUncertaintiesTool_handle, JetUncertaintiesTool));
     ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("JetDefinition",m_jetAlgo));
     ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("MCType",m_uncertMCType));
-    ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("IsData",!isMC()));
+    if(m_pseudoData) {
+      ANA_MSG_INFO("Input treated as pseudo-data");
+      ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("IsData",true));
+    }
+    else
```
kratsg: add braces here

xAODAnaHelpers/JetCalibrator.h
```diff
@@ -126,6 +126,9 @@ class JetCalibrator : public xAH::Algorithm
   /// @brief jet tile correction
   bool m_doJetTileCorr = false;
 
+  // needed in case want to treat MC as pseudoData for JER uncertainty propagation
```
kratsg: you need to follow the commenting scheme of the other variables to get it to show up in the documentation properly.

PR #1310, opened 2019-02-01
Custom jet flavour composition files for JetUncertainties

mattleblanc:
Adding an option to `JetCalibrator` which lets you specify an analysis-specific jet flavour composition for use with `JetUncertainties`. This basically amounts to setting the `AnalysisFile` option of `JetUncertainties`.

See the JES/JER recommendations for documentation of this functionality: https://twiki.cern.ch/twiki/bin/view/AtlasProtected/JetUncertaintiesRel21Summer2018SmallR#Configuring_and_initializing_the

I've put an example file which gives a dijet flavour composition in the following location, in case people are interested in what these should look like:

```
/afs/cern.ch/user/m/mleblanc/afs/public/JetEtMiss/DijetFlavourComp_Run2.root
```

🍻 MLB


kratsg:
Guessing that #1311 is used instead here.

PR #1314, opened 2019-02-06
Grabbing TEvent and TStore in BES::initialize()

lawrenceleejr:
Because of what looks like a change in execution order in 21.2.61, the TEvent and TStore being grabbed in the `fileExecute()` doesn't make it available for the `initialize()`. So proposing that these are also grabbed in the `initialize()` to protect against this. Maybe a bit redundant, but would make this less sensitive to function call order. Can we do something like this?

Without this, the current HEAD gives us the error:

```
Both TEvent and TStore objects are null. Cannot retrieve anything.
```

Thanks!

-L

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -346,6 +346,11 @@ EL::StatusCode BasicEventSelection :: initialize ()
     m_doPUreweightingSys = false;
   }
 
+  // get TEvent and TStore
+  //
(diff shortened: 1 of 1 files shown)


kratsg:
I've got a bunch of changes/refactors that will be coming shortly tonight (on my todo list). We can merge this in for now as a temporary stop-gap which will get changed later tonight.

Issue #1315, opened 2019-02-07
Fail Harder?


mattleblanc:
Hi all,

I just spent a set of grid jobs making some empty trees. It's not the end of the world, but I've seen others make this mistake in the past as well: if any of the detail strings in the `TreeAlgo` has a typo in it, jobs will proceed and appear to work, but produce only empty trees at the end of the day.

It seems to me that it would be better for users and from a resource-management point of view if this algorithm crashed and burned immediately, rather than silently proceeding ...

🍻  MLB

kratsg:
> I just spent a set of grid jobs making some empty trees. It's not the end of the world, but I've seen others make this mistake in the past as well: if any of the detail strings in the `TreeAlgo` has a typo in it, jobs will proceed and appear to work, but produce only empty trees at the end of the day.
> 
> It seems to me that it would be better for users and from a resource-management point of view if this algorithm crashed and burned immediately, rather than silently proceeding ...

Good idea. We can do some quick (and nice) parsing of the C++ source files for special algorithms.

kkrizka:
A better solution would be for the info switch classes to throw an error if it detects a string it cannot parse.

kratsg:
> A better solution would be for the info switch classes to throw an error if it detects a string it cannot parse.

We can't do this during config time (or at least not before the job is submitted), can we?

PR #1317, opened 2019-02-08
Feature: info switch python parsing

kratsg:
This provides a JSON file that will be checked for changes in Travis-CI builds for helping with info switch parsing. A later PR will parse this to check against a provided xAH configuration for specific info-switches.

This PR also moves a lot of the declarations back into the header file for info switches.

/cc @mattleblanc


kkrizka:
Can you try to keep the initializations in the cpp file? This will cause everything to recompile anytime we try to tweak the code.

kratsg:
> Can you try to keep the initializations in the cpp file? This will cause everything to recompile anytime we try to tweak the code.

Clang won't parse it. If you can figure out a way to get around the fact that clang only parses one file at a time...

kkrizka:
By not using clang for this check. I would run the check at run time as I suggested in #1315 (user should run a test job before submitting to grid anyways...) or somehow force the info switches to initialize at submission time.

Putting code in the header also increases compile time. Any class that includes the `HelperFunctoins` (aka any class) will have to recompile all this code.

kratsg:
> By not using clang for this check. I would run the check at run time as I suggested in #1315 (user should run a test job before submitting to grid anyways...) or somehow force the info switches to initialize at submission time.

I've tried to figure out a way to force info switches to initialize at submission time, but I cannot. So the easier way I've found is to do the parsing from within the xAH configuration instead -- but it won't be perfect since we can't handle all use-cases -- and this will only be used to check coverage.

> Putting code in the header also increases compile time. Any class that includes the `HelperFunctoins` (aka any class) will have to recompile all this code.

I've fixed this now by creating a tempfile that's the concatenated inputs of the header/source and feeding that into Clang's single-file parser.

kratsg:
@kkrizka any updates? Should I merge this in?

kkrizka:
Some things came up this week that I had to deal with urgently. But an alternative is on the way.

PR #1324, opened 2019-02-18
fix HighPt and LowPt parsing for MuQuality

lschaefer:
This PR fixes the attempt to add HighPt and LowPt WP functionality to the MuonSelector in PR #1308, which was buggy and was assigning Loose rather than HighPt or LowPt.

Because the HighPt and LowPt WPs are not in xAOD::Muon::Quality, they cannot be added to the HelperClasses muon quality parser, so we have to check by hand.

The second commit updates the way the JVT SF vectors are filled. Currently if the JVT needs to be done twice (i.e. once in baseline jet selection, so the JVT decision is available for MET, and once in signal jet selection, to do the JVT veto), these vectors are filled each time --> twice for every signal jet, so there is a misalignment between this and other jet vectors. This commit checks if the vector was already filled for this jet, and if so, skips the SF calculation/ vector filling.

### DIFF
--- a/Root/JetSelector.cxx
+++ b/Root/JetSelector.cxx
@@ -592,55 +592,55 @@ bool JetSelector :: executeSelection ( const xAOD::JetContainer* inJets,
 
     // Do it only if a tool with *this* name hasn't already been used
     //
-    if ( m_JVT_tool_handle.isInitialized() ) {
+    if ( m_JVT_tool_handle.isInitialized() && !m_jvtUsedBefore) {
+      //
+      //  If SF decoration vector doesn't exist, create it (will be done only for the 1st systematic for *this* jet)
+      //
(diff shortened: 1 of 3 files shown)


lschaefer:
update: I tried to make a new PR for a totally different feature, but the PR got thrown into this one, sorry for the confusion. I'll edit the description to explain the new PR as well.

kratsg:
Are we merging this in?

lschaefer:
I'm not sure why an old PR got overwritten when I tried to make a new one, but from my end the changes are ready to be merged, if you approve of them.

kratsg:
> I'm not sure why an old PR got overwritten when I tried to make a new one, but from my end the changes are ready to be merged, if you approve of them.

PRs are by branch.

PR #1328, opened 2019-03-04
Choose the scale at which the selection is performed

jbossios:
These changes allow choosing the scale at which the selection is performed in JetSelector. The default still remains to be the final (default) jet momentum scale.

This functionality allows, for instance, to derive and apply a calibration to the same set of jets.

This doesn't change the scale at which the jets are saved in the output container.


jbossios:
Small clarification, this doesn't change the scale at which the jets are saved in the new container.

PR #1329, opened 2019-03-04
Adding functionality in Jet and Jetcontainer to handle newest btaggers.

d-abbott:
These classes needed some upkeep to handle the newest btaggers DL1r, DL1rmu, MV2r, MV2rmu. I have tested the changes as part of my work with the hh4b analysis.


kratsg:
Looks pretty standard -- thanks!

PR #1342, opened 2019-04-18
Extend TruthSelector and TruthContainer

jbossios:
These changes allow to select truth muons/electrons from
STDMTruthMuons/Electrons and do not change the previous behaviour.

We can now select on *_dressed, classifierParticleType and classifierParticleOrigin in TruthSelector (these quantities are calculated with MCTruthClassifier).

We can now store pT_dressed, eta_dressed, phi_dressed and e_dressed into
the TTrees.

This will be used in the SM Z+HF analysis.


jbossios:
@kratsg Would be possible to merge this soon?

PR #1349, opened 2019-06-14
Add trigger jet cleaning

kalderon:
Changes by @etolley to add HLT jet cleaning. Also change accessors from char to int.


jbossios:
Wouldn't be better to add a specific string for jet trigger cleaning? Otherwise, now everybody gets that branch even for not trigger jets, and even in the cleanLight option which is meant to avoid having things which are mostly never used (like jet trigger cleaning).

kalderon:
Jona you're quite right, and the char -> int caused other problems (it was needed for HLT jets, but breaks non-HLT jets). So I will close this for now, please don't merge, and will reopen when fixed

kratsg:
@kalderon you can just mark as WIP.

kalderon:
Emma's changes from ~ a month ago fix this

kratsg:
All fixed?

jbossios:
Wouldn't be better not to set branches that then are not filled?

kratsg:
> Wouldn't be better not to set branches that then are not filled?

You would prefer the branches to stay fixed? Usually, the design pattern we do is to only include a branch if we intend to fill it. That at least indicates to the user that if the branch exists, it was filled in.

jbossios:
I meant exactly what you said. If I'm not mistaken with this MR we would be creating branches than then are not filled.

kalderon:
Yes, good point. I've tweaked things now so that there are no more "phantom" cleaning branches created (commit incoming).

kratsg:
Looks good to me.

jbossios:
Me too. Thank you!

PR #1350, opened 2019-06-20
WIP: Secondary Vertices

aroepe:
Hello!

We would like to integrate Displaced Vertices into the official repository for all DV analyses to use. Colleen Treado wrote these functions for her analysis, and we would like to generalize it for all DV analyses. NB: This code is not able to compile just yet! There are a lot of internal dependencies that need to be removed before it can compile. Before we put in a lot of effort on this, we would like to get some feedback from the developers. With some effort on our part, could this code be merged in? What changes would be required to match the xAODAnaHelpers coding philosophy?  

Also, there is currently some matching between truth and reco vertices based on hardcoded pdgIDs for dark pions.  Should this be left to analyzers and removed from the selector entirely, or generalized to take a set of PDG IDs in the config file for purposes of truth matching?


aroepe:
Adding @ctreado

kratsg:
Had a quick look. Seems fine, roughly. Big concerns: there are often loops over `i` instead of using `foreach` like loops, e.g.

```
for(int i=0; i < getN(); ++i){
  auto object = container->get(i);
}
```

instead of

```
for(const auto& object: container){
  ...
}
```

Secondly, documentation needs to be added pretty heavily. We follow doxygen style (triple back-slash and `/* ... */` blocks). Look at some of the other header files which have documentation. To make sure the algorithms show up in the generated docs at the end, you just need to make sure the algorithm names are included in the index/api files, etc... Again, pick a header file and just look for instances of that header file in the docs and mirror it.

Root/SecondaryVertexSelector.cxx
```diff
@@ -271,9 +271,12 @@ int SecondaryVertexSelector :: PassCuts ( const xAOD::Vertex* vtx, const xAOD::V
   std::vector<const xAOD::TrackParticle*> filteredTracks;
   std::vector<const xAOD::TrackParticle*> droppedTracks;
 
-  for ( size_t i = 0; i != vtx->nTrackParticles(); ++i ) {
+
+  std::vector< ElementLink< xAOD::TrackParticleContainer > > trackLinks = vtx->trackParticleLinks();
+
+  for(auto trkIter = trackLinks.begin(); trkIter != trackLinks.end(); ++trkIter) {
+	const xAOD::TrackParticle* trk = **trkIter;
```
kratsg: this isn't quite right.

```
for(const auto& trkLink: trackLinks){
  if(!trkLink.isValid()) break; // error out somehow
  const xAOD::TrackParticle* trk = *trkLink;
  ...
}
```

is a little bit cleaner.

jburzy01: Ah ok, didn't know an ElementLink can be dereferenced like that. Good to know!

jburzy01: FYI, that previous code sample was adapted from JetHists.cxx, if you would like I can update it there as well

kratsg: The `**trkIter` is basically `*(*trkIter)` with the inner parens being an ElementLink so...

kratsg: > FYI, that previous code sample was adapted from JetHists.cxx, if you would like I can update it there as well

Ahh, is it? We can update it there as well - but not in this PR (too out of scope).

kratsg:
Info/Error/Warning -> ANA_MSG_INFO and similar please.

aroepe:
Hi Giordon,

We are working out a couple of offline issues at the moment, so if you could please not merge this yet I would appreciate it. We might need to make more changes, but I will keep you updated. Thank you for all of your help!

kratsg:
@aroepe - updates here?

ctreado:
@aroepe -- I'm not sure how things have evolved on your end, but I've made quite a few changes in our EmergingJetsAnalysis repo, including adding a dedicated VertexMatcher algorithm. Not sure how time-consuming it will be to include these updates here, as well, but it may be a good idea...

jburzy01:
@kratsg @ctreado Sorry for not following up sooner about this. Since there are dependencies in this code that exist in other GitLab repositories (VsiBonsaiTool.cxx) we have instead structured our code with xAH as a submodule rather than integrating our tools into the main repo. We can close this PR.

PR #1353, opened 2019-07-10
HLT jet duplicate removal and sorting

kalderon:
Add options to JetSelector to sort jets and remove duplicates, for the benefit of HLT jets. HLT jets are not sorted in (D)AODs, and do not get calibrated so cannot be sorted in JetCalibrator. HLT "GSC" jets in some MC are duplicated (identical eta and phi, very close pT) thanks to how they are constructed, removing duplicates allows the relevant triggers to be properly emulated. Duplicates are removed by ordering in eta and removing any that have identical eta. @jbossios 

P.S. The commits from jet cleaning made their way here since I branched from that feature branch accidentally, but there are no actual changes to those parts here thanks to rebasing so if commits are squashed it should all be OK.


kratsg:
@jbossios - if you thumbs up/approve -- we'll merge.

kalderon:
@kratsg It looks like Jona approved

PR #1356, opened 2019-07-22
Save vertex information

jbossios:
When "m_vertexDetails" from TreeAlgo is "primary" or "all", vertex information (x, y and z) for, respectively, primary/hard-scatter vertex or all vertices is saved.
Similarly to other objects, it is possible to save more than one vertex container by listing them in "m_vertexContainerName" (space separated). It is also possible to choose the branch name(s) with "m_vertexBranchName".

When saving more than one vertex collection, the first vertex collection in "m_vertexContainerName" is the one used elsewhere, for instance for track-based jet moments.

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -640,9 +640,7 @@ void HelpTreeBase::FillL1Jets( const xAOD::JetRoIContainer* jets, const std::str
 void HelpTreeBase::ClearL1Jets(const std::string jetName) {
 
   xAH::L1JetContainer* thisL1Jet = m_l1Jets[jetName];
-  std::cout << "Before thisL1Jet->clear()" << std::endl;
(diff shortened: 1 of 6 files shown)


kratsg:
Looks backwards compatible? I guess we're not going for a new variable here or a rename.

jbossios:
It is backwards compatible, "m_vertexDetails" is empty by default, and in that case, no vertex info is saved into the TTrees (no branch is created). If anyone ever wants to save vertex info, they just have to set that string, unless they want info from more than one vertex collection. It is only that case where they should be careful (first collection is the one which is used elsewhere). I found this way simpler than having to create another very similar string specifically for saving vertex info. But that is a very simple change if you would prefer that way. We could call it: "m_vertexContainerNameToWrite" or something like that which makes it obvious is the one to set up to save those vertices to the TTree.

PR #1360, opened 2019-09-22
Extend xAH::VertexContainer to also handle truth vertices

taehyounpark:
Hello everyone,

I was wondering if it would also be possible to save truth vertex information in HelpTreeBase using the xAH::VertexContainer. This branch is a "minimal extension" that I've used for myself. Also, I've used the code logic used in [IsolatedJetTree](https://gitlab.cern.ch/atlas-jetetmiss-jesjer/MCcalibrations/IsolatedJetTree/blob/master/Root/IsolatedJetAlgo.cxx#L269) to get the hard-scatter truth vertex.

I don't know if this is in line with the rest of the framework (i.e. Should there be explicitly separate xAH::TruthVertexContainers with Add/FillTruthVertices methods?). Whichever way is the best, it would be great if this functionality is added!


kratsg:
Looks like a regular operator overload to switch between reco VC and truth VC -- which seems fine to me.

jbossios:
It would be maybe good to be able to save reco and truth vertices simultaneously, it is a bit confusing that if both FillVertices() functions are used then only the last call will work.

jbossios:
> It would be maybe good to be able to save reco and truth vertices simultaneously, it is a bit confusing that if both FillVertices() functions are used then only the last call will work.

even worst, reco and truth vertices will be saved into the same branches

taehyounpark:
Hi Jona,

How about having the methods be named FilledTruthVertices then? This won’t stop users from saving truth and reco vertices into the same branches, but at least it will be explicit when they write their own code?

jbossios:
As you said, that would not fix the issue, people might use both calls w/o knowing both are saved into the same branches. I would care more about creating dedicated branches than changing the function's name, although that would be nice too.

taehyounpark:
Then it sounds like having a separate xAH::TruthVertexContainer would be the better way?

kratsg:
> Then it sounds like having a separate xAH::TruthVertexContainer would be the better way?

That seems to be the current suggestion, yes. Especially because one could imagine doing specific things in truth that you wouldn't do in reco...

taehyounpark:
Hi again,

I've moved the truth vertices out of xAH::VertexContainer into a new xAH::TruthVertexContainer. How does this look?

jbossios:
Looks good to me. Thanks!

kkrizka:
I would be against creating a new class for the truth vertex container. This duplicates code that I don't see a reason to be duplicated.

Maybe take a look at how large R jets and truth large R jets are stored:
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelpTreeBase.cxx#L835

taehyounpark:
Hi @kkrizka,

So this approach would be to have xAH::VertexContainer handle both reco & truth vertices, but differentiate them by method names (AddVerticies vs. AddTruthVerticies) and explicitly have two instances of the xAH::VertexContainer in HelpTreeBase, one for each kind. Is this right?

kkrizka:
Yes, that's how I would do it.

taehyounpark:
I've implemented the suggestions, please take a look and see if these changes are ready for merging. Thank you!

kratsg:
I'm ok with these changes, @kkrizka ?

PR #1361, opened 2019-09-24
Histogram booking mods

kalderon:
This borrows some functionality from https://gitlab.cern.ch/atlas/athena/blob/master/Control/AthenaBaseComps/AthenaBaseComps/AthHistogramming.h and .cxx to make use of the existing m_allHists vector to avoid having to declare each histogram in the header file, simplifying development.

See examples of use in https://gitlab.cern.ch/atlas-phys-exotics-dijet-tla/TLAAlgosFullRun2/blob/master/Root/SelectAndHistogram.cxx. But a snippet would be:

.h:
```cpp
  HistogramManager* m_histManager = nullptr; //!
```

.cxx
```cpp
  // histInitialize()
  m_histManager = new HistogramManager(m_name, "");
  ANA_CHECK( m_histManager -> initialize() );
  m_histManager->book("myDir/h_jetPt", "", "jet pT [GeV]", 100, 0, 2000);
  m_histManager -> record( wk() );
  //
  // execute()
  m_histManager->fillHist("myDir/h_jetPt", jet->pt() * 0.001);
```

I'm not sure why commit 64cf7f4 has made it in here, it should already be in master, so perhaps I need to do a new rebase and cherry-pick just ffd5016, but I thought I'd create the pull request now so any discussion can start since this is a slightly bigger change than I normally make (albeit no existing workflows should be affected beyond some tiny extra overhead in record().

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -1007,8 +1007,10 @@ EL::StatusCode BasicEventSelection :: execute ()
 
     }
 
-    static SG::AuxElement::Decorator< float > weight_prescale("weight_prescale");
-    weight_prescale(*eventInfo) = triggerChainGroup->getPrescale();
+    if ( m_storePrescaleWeight ) {
+      static SG::AuxElement::Decorator< float > weight_prescale("weight_prescale");
+      weight_prescale(*eventInfo) = triggerChainGroup->getPrescale();
+    }
 
     if ( m_storePassL1 ) {
       static SG::AuxElement::Decorator< int > passL1("passL1");
--- a/Root/HistogramManager.cxx
+++ b/Root/HistogramManager.cxx
@@ -151,18 +151,16 @@ void HistogramManager::Sumw2(TH1* hist, bool flag /*=true*/) {
 void HistogramManager::record(TH1* hist) {
   m_allHists.push_back( hist );
 
-  // Check if the hash for this histName already exists, i.e., if we have a hash collision
+  // Check if this histName already exists
   std::string histName = hist->GetName();
-  const hash_t histHash = this->makeHash(histName);
-  HistMap_t::const_iterator it = m_histMap.find( histHash );
+  HistMap_t::const_iterator it = m_histMap.find( histName );
   if ( it != m_histMap.end() ) // It does exist!
     {
-      ANA_MSG_WARNING( "Detected a hash collision. The hash for the histogram with name=" << histName
-                       << " already exists and points to a histogram with name=" << it->second->GetName()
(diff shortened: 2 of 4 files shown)


kratsg:
I think I like this and I don't see any obvious drawbacks. @kkrizka ?

kkrizka:
No complaints from me about this functionality.

However I want to point out that doing the histogram lookup will slow down your code somewhat, since you are calculating the string hash for every event in your selection.

kalderon:
Thanks. Yes, it is a bit slower - on 100k events, filling about 10 times per event, it takes 0.8-1.0s to fill this way and 0.15-0.25s to fill the original way, to be compared with a total of 15s for the execute time, i.e. a ~5% slowdown, which I think I can live with for development purposes but good to remember that switching back is an option to squeeze out some performance later.

kkrizka:
Did you measure this time with the changes that I recommended?

kalderon:
I'm not quite sure what it is that you're recommending, can you elaborate? Is there a way of getting around the lookup / a better way than hashing? I don't see any comments on the code

kkrizka:
Ah, first time I am using the code review function in github. Didn't realize I had to click a second button to actually submit my comments. Sorry about that!

kalderon:
@kkrizka I modified according to your suggestions, which speed it up a bit, thanks - the 0.9s I quoted before is reduced to 0.75s, vs 0.2s for "original" (ie the average time per event for "original" histogramming is 1.9e-6 s, and 7.5e-6 s for the new-with-changes). I think then that this is OK for merging?

kkrizka:
I think it is OK for merging.

kratsg:
> @kkrizka I modified according to your suggestions, which speed it up a bit, thanks - the 0.9s I quoted before is reduced to 0.75s, vs 0.2s for "original" (ie the average time per event for "original" histogramming is 1.9e-6 s, and 7.5e-6 s for the new-with-changes). I think then that this is OK for merging?

I still think this is too big of a slowdown. Can you run over 100k events and figure out how much slower it is overall? Is it noticeable enough? You're quoting basically a 500% slowdown.

kalderon:
It's a roughly five-fold slowdown in histogram filling, yes, but even for a job that only does some basic event and kinematic selection this is pretty negligible in overall terms. On 100k events, where 10k pass selection and are histogrammed, the timings running it three times are:
- 0.034, 0.025, 0.025 s for the "old-style" histogram filling
- 0.119, 0.125, 0.138 s for the "new-style" histogram filling
- 26.08, 27.21, 27.88 s from the second event (ie after all initialisation) to the end
i.e. 0.5% of the total time vs 0.1%, so a slowdown of 0.4%

kalderon:
Thanks :-)

PR #1362, opened 2019-10-05
MC Cleaning in JetSelector

marianatoscani:
Implementing MC Cleaning criteria wrt jets in JetSelector.
Comparing the average pT of leading and subleading jet with leading truth jet pT to remove events in which the leading truth jet has low pT but the leading reco one has high pT due to pile up overlay.

### DIFF
--- a/Root/JetSelector.cxx
+++ b/Root/JetSelector.cxx
@@ -442,7 +442,7 @@ EL::StatusCode JetSelector :: execute ()
     }
 
     // Check against pile-up only jets:
-    if( isMC() && m_doMCCleaning && m_haveTruthJets ){
+    if( isMC() && m_doMCCleaning && m_haveTruthJets && inJets->size()>1 ){
     	
         float pTAvg = 0.;
     	pTAvg = ( inJets->at(0)->pt() + inJets->at(1)->pt() ) / 2.0;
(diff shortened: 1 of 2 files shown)


Root/JetSelector.cxx
```diff
@@ -441,6 +441,17 @@ EL::StatusCode JetSelector :: execute ()
       }
     }
 
+    // Check against pile-up only jets:
+    if( isMC() && m_doMCCleaning && m_haveTruthJets ){
+    	
+        float pTAvg = 0.;
+    	pTAvg = ( inJets->at(0)->pt() + inJets->at(1)->pt() ) / 2.0;
```
kratsg: need protection against events with only one jet.

jbossios:
Sorry for not commenting on this on time (before merging). I would do instead something like the following:

// note: this is not the actual code, it is just to illustrate the idea.
float pTavg = jet_pt->at(0);
if(njet>1) pTavg = ( jet_pt->at(0) + jet_pt->at(1) ) / 2.0;
if( ntruthjets == 0 || (pTavg/truthjet_pt->at(0)>1.4) ){
  ANA_MSG_DEBUG("Failed MC cleaning, skipping event");
  wk()->skipEvent();
}

If you wish to have these changes in, I can prepare a PR.

kratsg:
> If you wish to have these changes in, I can prepare a PR.

Go ahead and submit a PR

PR #1365, opened 2019-10-15
Add eventClean_LooseBad flag to EventInfo output.

kkrizka:
Saves the jet event cleaning flag to ntuples. Added to the other event cleaning variables that are enable using the `eventCleaning` switch.


mattleblanc:
This was already here?

https://github.com/UCATLAS/xAODAnaHelpers/pull/1298

kratsg:
> This was already here?
> 
> #1298

That implemented it in BES. Wasn't in the TreeAlgo outputs.

PR #1367, opened 2019-10-15
DoPFlow in METMaker and METSignificance fix

jbossios:
- Initialize METSignificance tool only when requested

- Now possible to setup DoPFlow in METMaker tool


jbossios:
@kratsg How can I trigger a retry?

kratsg:
> @kratsg How can I trigger a retry?

done

PR #1380, opened 2019-11-12
Allow to set muon calibration tool properties

jbossios:
Add override options for StatComb, SagittaCorr, doSagittaMCDistortion,SagittaCorrPhaseSpace properties.

### DIFF
--- a/Root/MuonCalibrator.cxx
+++ b/Root/MuonCalibrator.cxx
@@ -124,6 +124,22 @@ EL::StatusCode MuonCalibrator :: initialize ()
   m_numObject     = 0;
 
   // Initialize the CP::MuonCalibrationPeriodTool
+  if ( m_overrideStatComb ) {
+    ANA_MSG_WARNING("Overriding muon calibration StatComb property to " << m_statComb);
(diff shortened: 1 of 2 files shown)


kratsg:
Use `int` instead of `bool` and only use 4 variables with something like

```
  if ( m_statComb >= 0 ) {
    ANA_MSG_WARNING("Overriding muon calibration StatComb property to " << m_statComb);
    ANA_CHECK(m_muonCalibrationTool_handle.setProperty("StatComb", static_cast<bool>(m_statComb)));
  }
```

jbossios:
> Use `int` instead of `bool` and only use 4 variables with something like
> 
> ```c++
>   if ( m_statComb >= 0 ) {
>     ANA_MSG_WARNING("Overriding muon calibration StatComb property to " << m_statComb);
>     ANA_CHECK(m_muonCalibrationTool_handle.setProperty("StatComb", static_cast<bool>(m_statComb)));
>   }
> ```

Done. I had to add more variables though, since there is one for each data-taking period (data15+16, data17 and data18).

PR #1381, opened 2019-11-14
Fix bug in MuonInFatJetCorrector when making muon-trackjet ElementLink.

kkrizka:
Replace `muon->index()` with a manually incremented index when making the muon to track jet ElementLink. The `index()` function returns the index in the original xAOD container and not the resorted calibrated muon container.

The `MuonInFatJetCorrector` now returns the same masses as the algorithm implementation in other frameworks. I would consider it fully validated now.


kratsg:
> The `index()` function returns the index in the original xAOD container and not the resorted calibrated muon container.

Interesting. I guess this is fixed if one uses a deep-copy instead of a shallow copy.

Issue #1371, opened 2019-10-30
Ordering of objectName and detailStr arguments in HelpTreeBase


taehyounpark:
Hi everyone,

This is a very minor point but I've noticed that the Add*() methods for some objects (TrackParts, TruthParts) in HelpTreeBase have different ordering of arguments from others.

Most:
`
AddSomething(const std::string detailStr = "", const std::string somethingName = "sth");
`
TrackParts & TruthParts:
`
AddSomething(const std::string somethingName, const std::string detailStr);
`

Personally, harmonizing everything to have arguments in the first style makes more sense since one is often processing one collection of objects at a time (i.e. have detailStr come first, supply default objectName)? But it is also possible that many users have custom ntuple setups that might be affected?

kratsg:
It's a good question. We should harmonize, and ask on the xAH mailing list to see if anyone is currently affected or rely on the ordering.

kkrizka:
I fully support this.

jbossios:
This would affect my xAH-based analysis code but I support this change. Would be possible to set a given advertised date for the merging to be done so I (and others) can prepare the needed changes so the transition is smooth?

kratsg:
> This would affect my xAH-based analysis code but I support this change. Would be possible to set a given advertised date for the merging to be done so I (and others) can prepare the needed changes so the transition is smooth?

sure. We generally advertise on the mailing list for any breaking-backward-compatibility API changes.

kratsg:
@taehyounpark don't close this until the PR is merged in...

PR #1384, opened 2019-11-16
Harmonize TreeAlgo Add & Fill methods argument ordering

taehyounpark:
This is a follow-up on #1371.

I've changed the ordering of the Track/TruthPart methods in HelpTreeBase & TreeAlgo to be:

Add*(detailStr,objectName)
Fill*(objectContainer,objectName)

As in accordance with the other container methods. I've tested that this does not change results for the track information in my personal TreeAlgo config. If this is merged into the master, there would have to be an announcement to analyzers with custom HelpTreeBase-based algorithms that the API has been changed.

Please let me know how this looks!

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -124,7 +124,7 @@ void HelpTreeBase::Fill() {
  *
  ********************/
 
-void HelpTreeBase::AddEvent( const std::string detailStr ) {
+void HelpTreeBase::AddEvent( const std::string& detailStr ) {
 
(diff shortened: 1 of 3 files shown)


kkrizka:
While you are changing the API, can you also change `const std::string` to `const std::string&`? There is no need to keep passing those strings as copies.

taehyounpark:
Hi @kkrizka,

Do you mean for all the detailStrs and objectNames of all the Containers in HelpTreeBase?

kkrizka:
Hi @taehyounpark ,

Yes, If you could do it for all containers that would be great. It's something that has been bugging me about `HelpTreeAlgo` for a while now.

--
Karol Krizka

PR #1385, opened 2019-11-16
Add jet MC cleaning cut-point option

taehyounpark:
I see that the standard procedure to apply an "cleaning" on multijet MC samples to remove pileup overlay, which requires that

leading dijet averge pT / leading truth jet pT < 1.4

Although this cut is loose enough to not be expected to bias any resolution effects, I think it would be helpful nonetheless to have the 1.4-cutoff point be a changeable default setting.

### DIFF
--- a/xAODAnaHelpers/JetSelector.h
+++ b/xAODAnaHelpers/JetSelector.h
@@ -80,6 +80,7 @@ public:
   std::string m_jetScale4Selection = "Final";
   /// @brief (MC-only) Kill event if avg(pT[0],pT[1])>1.4*truth_pT[0]
   bool m_doMCCleaning = false;
+  float m_mcCleaningCut = 1.4;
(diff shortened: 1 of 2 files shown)


xAODAnaHelpers/JetSelector.h
```diff
@@ -80,6 +80,7 @@ class JetSelector : public xAH::Algorithm
   std::string m_jetScale4Selection = "Final";
   /// @brief (MC-only) Kill event if avg(pT[0],pT[1])>1.4*truth_pT[0]
   bool m_doMCCleaning = false;
+  float m_mcCleaningCut = 1.4;
```
kkrizka: Update the documentation comment describing what this setting does.

kkrizka:
Looks good to me.

Out of curiosity, why limit the cut to >1.0?

PR #1388, opened 2019-11-25
TrackSelector:  mcEvtWeight, harmonization, more options

CoffeeTableShoe:
- More options for the TrackSelector,
- Now uses InDetTrackSelectionTool. 
- Some option names are modified to harmonize with other selectors and with the official tool. To restore previous behavior:
    - m_eta_max -> m_etaSigned_max (m_eta_max is now the max absoiute value)
    - m_eta_min -> m_etaSigned_min (m_eta_min is now the min absoiute value)
    - m_nSi_min -> m_nSiPhysical_min (m_Si_min is now the number of Si hits counting dead sensors. This is the convention used in the official tool and what goes into the recommendations)
- nBL is kept for downward compatibility, but is not recommended since it uses the deprecated summary value xAOD::numberOfBLayerHits, which are actually the number of IBL hits. m_nInnermostPixel and m_nNextToInnermostPixel are provided.
- fix mcEvtWeight which was always 1.0 before

- [x] The tool seems to be setup correctly now. No more errors.

### DIFF
--- a/Root/TrackSelector.cxx
+++ b/Root/TrackSelector.cxx
@@ -185,24 +185,18 @@ EL::StatusCode TrackSelector :: execute ()
 
   // MC event weight
   //
-  m_mcEvtWeight = 1.0;
-  static SG::AuxElement::Accessor< float > m_mcEvtWeightAcc("mcEventWeight");
-  if ( ! m_mcEvtWeightAcc.isAvailable( *eventInfo ) ) {
-    ANA_MSG_ERROR( "mcEventWeight is not available as decoration! Aborting" );
-    return EL::StatusCode::FAILURE;
-  }
-  m_mcEvtWeight = m_mcEvtWeightAcc( *eventInfo );
+  float mcEvtWeight = eventInfo->mcEventWeight();
 
   if(m_doTracksInJets){
     return executeTracksInJets();
   } else{
-    return executeTrackCollection();
+    return executeTrackCollection(mcEvtWeight);
   }
 
   return EL::StatusCode::SUCCESS;
 }
 
-EL::StatusCode TrackSelector :: executeTrackCollection ()
+EL::StatusCode TrackSelector :: executeTrackCollection (float mcEvtWeight)
 {
   m_numEvent++;
 
@@ -273,7 +267,7 @@ EL::StatusCode TrackSelector :: executeTrackCollection ()
   m_numEventPass++;
   if(m_useCutFlow) {
     m_cutflowHist ->Fill( m_cutflow_bin, 1 );
-    m_cutflowHistW->Fill( m_cutflow_bin, m_mcEvtWeight);
+    m_cutflowHistW->Fill( m_cutflow_bin, mcEvtWeight);
   }
 
   return EL::StatusCode::SUCCESS;
--- a/Root/TrackSelector.cxx
+++ b/Root/TrackSelector.cxx
@@ -4,6 +4,7 @@
 #include "AthContainers/ConstDataVector.h"
 #include "xAODAnaHelpers/HelperFunctions.h"
 #include "xAODAnaHelpers/TrackSelector.h"
+#include "InDetTrackSelectionTool/InDetTrackSelectionTool.h"
 
 // ROOT include(s):
 #include "TFile.h"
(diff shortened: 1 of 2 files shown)


Root/TrackSelector.cxx
```diff
@@ -132,7 +133,44 @@ EL::StatusCode TrackSelector :: initialize ()
...
+  if(m_nSctHits_min                     != 1e8)ANA_CHECK(m_trkSelTool->setProperty("minNSctHits",                     m_nSctHits_min));
+  if(m_nSctHitsPhysical_min             != 1e8)ANA_CHECK(m_trkSelTool->setProperty("minNSctHitsPhysical",             m_nSctHitsPhysical_min));
+  if(m_nSctSharedHits_max               != 1e8)ANA_CHECK(m_trkSelTool->setProperty("maxNSctSharedHits",               m_nSctSharedHits_max));
+  if(m_nSctHoles_max                    != 2e8)ANA_CHECK(m_trkSelTool->setProperty("maxNSctHoles",                    m_nSctHoles_max));
+  if(m_nSiSharedModules_max             != 1e8)ANA_CHECK(m_trkSelTool->setProperty("maxNSiSharedModules",             m_nSiSharedModules_max));
+  if(m_chi2Prob_min                     != 1e8)ANA_CHECK(m_trkSelTool->setProperty("minProb",                         m_chi2Prob_min));
+  if(m_chi2NdofCut_max                  != 1e8)ANA_CHECK(m_trkSelTool->setProperty("maxChiSqperNdf",                  m_chi2NdofCut_max));
+  if(m_cutLevelString                   != "" )ANA_CHECK(m_trkSelTool->setProperty("CutLevel",                        m_cutLevelString));
+
+  m_trkSelTool->initialize();
```
kratsg: missing an ANA_CHECK

Root/TrackSelector.cxx
```diff
@@ -144,6 +182,19 @@ EL::StatusCode TrackSelector :: execute ()
 
   ANA_MSG_DEBUG("Applying Track Selection... " << m_name);
 
+  const xAOD::EventInfo* eventInfo(nullptr);
+  ANA_CHECK( HelperFunctions::retrieve(eventInfo, m_eventInfoContainerName, m_event, m_store, msg()) );
+
+  // MC event weight
```
kratsg: you should be using `eventInfo->mcEventWeight()` not the decorator accessor

Root/TrackSelector.cxx
```diff
@@ -352,103 +401,64 @@ EL::StatusCode TrackSelector :: histFinalize ()
...
   float        ndof        = trk->numberDoF();
-  if( m_chi2NdofCut_max != 1e8){
-    float chi2NDoF     = (ndof > 0) ? chi2/ndof : -1;
-    if( chi2NDoF > m_chi2NdofCut_max ) {return 0;}
-  }
-
   if( m_chi2Prob_max != 1e8 ){
     float        chi2Prob    = TMath::Prob(chi2,ndof);
-    if( chi2Prob > m_chi2Prob_max) {return 0;}
+    if( chi2Prob > m_chi2Prob_max) { return 0; }
```
kratsg: are these checks needed if you have the trkselector tool passing or not?

xAODAnaHelpers/TrackSelector.h
```diff
@@ -68,9 +120,13 @@ class TrackSelector : public xAH::Algorithm
 
 private:
 
+  float m_mcEvtWeight = 1.0;
+
   std::vector<std::string> m_passKeys;
   std::vector<std::string> m_failKeys;
 
+  InDet::InDetTrackSelectionTool * m_trkSelTool;
```
kratsg: this should be a tool handle, not a tool.

xAODAnaHelpers/TrackSelector.h
```diff
@@ -68,9 +120,13 @@ class TrackSelector : public xAH::Algorithm
 
 private:
 
+  float m_mcEvtWeight = 1.0;
```
kratsg: this does not need to be a class member variable.

Root/TrackSelector.cxx
CoffeeTableShoe: Yes, these cuts are not available in the tool.

Root/TrackSelector.cxx
CoffeeTableShoe: I commited the change but I am surprised as it is done in this way in every other selector I looked at.

kratsg: This is likely for differences between AODs and DAODs, but I need to check.

kratsg:
Is this still a WIP?

Issue #1389, opened 2019-11-26
Updating Treatment of SimulationFlavour in MetaData


jdandoy:
We currently have 2 methods for manually setting SimulationFlavour of a sample (FullSim or AFII).   It can be done through a flag: https://github.com/UCATLAS/xAODAnaHelpers/blob/0d1a3af7be6867fe8e1b88189d49a1af860d86e8/xAODAnaHelpers/BasicEventSelection.h#L67

or through a private MetaData string in SampleHandler that requires some specialized c++ code to configure
https://github.com/UCATLAS/xAODAnaHelpers/blob/0d1a3af7be6867fe8e1b88189d49a1af860d86e8/xAODAnaHelpers/BasicEventSelection.h#L56-L64

We do not have a method for using the existing MetaData to set this.

I'd like to update this to have the option  to use existing MetaData:https://github.com/UCATLAS/xAODAnaHelpers/blob/4a6043dbe919cbd931f9937c20b702adb533c8ec/Root/BasicEventSelection.cxx#L1095-L1103

At the same time we can simplify things by only have one method to manually set this, rather than two.  I would prefer using the dedicated flags m_setAFII & m_setFS as these are more straightforward and can be set directly from the configuration file, rather than dedicated code.  This should be mostly backwards compatible because those now relying on the MetaData string in SampleHandler will now get the actual MetaData decision instead, which is probably what they wanted.

If people agree I can also introduce this into all the files that require AFII information for calibration or SFs.  This resolves #751

PR #1390, opened 2019-11-26
Simulation flavour from metadata

jdandoy:
Addresses #1389.  Properly gets FullSim vs FastSim information from MetaData unless otherwise specified.  isFastSim is a new function available to all algorithms, mirroring isMC.   Use m_forceData, m_forceFullSim, and m_forceFastSim flags to allow any algorithm to override the defaults & give xAH that flexibility we all love.  Includes backwards compatibility for older flags where needed.  Removes ability to force sample type using SampleHandler, but this should be unnecessary now that MetaData access properly works.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -1083,11 +1083,7 @@ StatusCode BasicEventSelection::autoconfigurePileupRWTool()
   ANA_CHECK( m_event->retrieve( eventInfo, "EventInfo" ) );
 
   // Determine simulation flavour
-  std::string SimulationFlavour;
-  if( isFastSim() )
-    SimulationFlavour = "AFII";
-  else
-    SimulationFlavour = "FS";
+  std::string SimulationFlavour = isFastSim() ? "AFII" : "FS";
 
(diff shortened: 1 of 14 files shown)


Root/BasicEventSelection.cxx
```diff
@@ -1087,12 +1084,10 @@ StatusCode BasicEventSelection::autoconfigurePileupRWTool()
...
   std::string SimulationFlavour;
-  if( m_setAFII )
-    SimulationFlavour="AFII";
+  if( isFastSim() )
+    SimulationFlavour = "AFII";
   else
-    SimulationFlavour = wk()->metaData()->castString("SimulationFlavour");
-  if(SimulationFlavour.empty())
-    SimulationFlavour="FS";
+    SimulationFlavour = "FS";
```
kratsg: Fix the above lines by replacing with a ternary (it's a bit cleaner to me)

```suggestion
    SimulationFlavour = isFastSim()?"AFII":"FS";
```

jdandoy: Done.

PR #1392, opened 2019-11-27
MuonInFatJetCorrector: Add support for different jet calibrations

kkrizka:
Make the calibration used to obtain the calibrated calorimeter and track assisted masses configurable. Different values might be needed depending on the latest available calibrations.

The default is now `JetInsituScaleMomentum` for data and `JetJMSScaleMomentum` for MC.

### DIFF
--- a/Root/MuonInFatJetCorrector.cxx
+++ b/Root/MuonInFatJetCorrector.cxx
@@ -65,8 +65,7 @@ EL::StatusCode MuonInFatJetCorrector :: initialize()
 
   //
   // Automatically determine calibrated mass decorators, if asked
-  if(m_calibratedMassDecorator.empty())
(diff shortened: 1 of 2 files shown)


kratsg:
Let's say `m_calibratedMassDecorator` is not empty, is it possible to automatically change it for MC/data? I guess it's too much to ask.

What's the twiki documentation here?

kkrizka:
I can separate `m_calibratedMassDecorator` into `m_calibratedMassDecoratorMC` and `m_calibratedMassDecoratorData`. Might be useful in case the calibrations name change again in the future.

These calibration names were obtained via private communication with the jet insitue calibration contacts.

kratsg:
> I can separate `m_calibratedMassDecorator` into `m_calibratedMassDecoratorMC` and `m_calibratedMassDecoratorData`. Might be useful in case the calibrations name change again in the future.

Let's go with that for now. I guess we can't make it magic enough...

PR #1394, opened 2019-12-02
Use recommended Config::algorithm API in docs and scripts

matthewfeickert:
xAH [warns users](https://github.com/UCATLAS/xAODAnaHelpers/blob/a4fe51ae81407d34c185ce67407cc81a71ab5924/python/config.py#L21) to use `Config::algorithm` instead of `Config::setalg` so show this in the docs. Also change the one time `setalg` is used in `xAH_run.py`.

### DIFF
--- a/docs/FAQs.rst
+++ b/docs/FAQs.rst
@@ -19,7 +19,7 @@ How do I...
 
         from xAH_config import xAH_config
         c = xAH_config()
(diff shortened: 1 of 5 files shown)


matthewfeickert:
I'll need to make a followup docs PR, as I now realize the alignment in the example looks dumb.

Issue #1372, opened 2019-10-31
Crash when using continuous b-tagging on MC


beojan:
Crashes if continuous b-tagging is used on MC. This seems to be because it tries to retrieve scale factors that do not exist.

See https://gitlab.cern.ch/hh4b/XhhCommon/-/jobs/5990676 for an example.

kratsg:
It's not clear to me that this is the case?

```
CalibrationDataInterfaceROOT::checkWeightScaleFactors: cross-checking scale factors for objects MV2c10/AntiKt4EMTopoJets/Continuous/B/default_SF and MV2c10/AntiKt4EMTopoJets/Continuous/B/default_Eff
----------------------------------------------------------------------------------------------------
CalibrationDataInterfaceROOT::checkWeightScaleFactors: cross-checking scale factors for objects MV2c10/AntiKt4EMTopoJets/Continuous/Light/default_SF and MV2c10/AntiKt4EMTopoJets/Continuous/Light/default_Eff
----------------------------------------------------------------------------------------------------
CalibrationDataInterfaceROOT::checkWeightScaleFactors: cross-checking scale factors for objects DL1/AntiKt4EMTopoJets/Continuous/B/default_SF and DL1/AntiKt4EMTopoJets/Continuous/B/default_Eff
----------------------------------------------------------------------------------------------------
CalibrationDataInterfaceROOT::checkWeightScaleFactors: cross-checking scale factors for objects DL1/AntiKt4EMTopoJets/Continuous/Light/default_SF and DL1/AntiKt4EMTopoJets/Continuous/Light/default_Eff
----------------------------------------------------------------------------------------------------
BJetEffCorCont_AntiKt4...ERROR   /build/atnight/localbuilds/nightlies/21.2/athena/PhysicsAnalysis/JetTagging/JetTagPerformanceCalibration/xAODBTaggingEfficiency/Root/BTaggingEfficiencyTool.cxx:571 (virtual CP::CorrectionCode BTaggingEfficiencyTool::getScaleFactor(int, const Analysis::CalibrationDataVariables&, float&) const): BTaggingEfficiencyTool::getScaleFactor call to underlying code returned a kError!
[...]

PR #1400, opened 2019-12-12
Save the Continuous b-tagging quantiles, SF, & ineffSFs to Ntuple

nhartman94:


Root/BJetEfficiencyCorrector.cxx
```diff
@@ -130,7 +133,9 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
   if (m_taggerName == "DL1rmu") { taggerOK = true; m_getScaleFactors =  false; }
 
   // Continuous
-  if (m_operatingPt == "Continuous") { opOK = true; taggerOK = true; m_getScaleFactors =  true; m_useContinuous = true; ANA_MSG_DEBUG(" Using continuous b-tagging");}
+  // if (m_operatingPt == "Continuous") { opOK = true; taggerOK = true; m_getScaleFactors =  true; m_useContinuous = true; ANA_MSG_DEBUG(" Using continuous b-tagging");}
+  //m_useContinuous = static_cast<bool>(m_operatingPt == "Continuous");
+  //if (m_useContinuous) ANA_MSG_DEBUG(" Using continuous b-tagging");
```
jbossios: Could you please remove the commented code?

nhartman94: Hi yes, ofc, one second.

nhartman94:
> Why are there so many changes in BJetEfficiencyCorrector when my understanding is that the only missing thing was to save what we already have into a TTree.

It looked like there was originally some logic where it was only looking for the SF if dec_isBTag() wasTrue, but this decoration never got in the continuous mode since we use the dec_quantile() instead. I tried to just shuffle a few lines to resolve this.

nhartman94:
> > Why are there so many changes in BJetEfficiencyCorrector when my understanding is that the only missing thing was to save what we already have into a TTree.
> 
> It looked like there was originally some logic where it was only looking for the SF if dec_isBTag() wasTrue, but this decoration never got in the continuous mode since we use the dec_quantile() instead. I tried to just shuffle a few lines to resolve this.

I was following your discussion in https://github.com/UCATLAS/xAODAnaHelpers/issues/1372 for starting these edits.

nhartman94:
Also tagging @mswiatlo and @rateixei since we want to use this for the HH->4b analysis

nhartman94:
> > Why are there so many changes in BJetEfficiencyCorrector when my understanding is that the only missing thing was to save what we already have into a TTree.
> 
> It looked like there was originally some logic where it was only looking for the SF if dec_isBTag() wasTrue, but this decoration never got in the continuous mode since we use the dec_quantile() instead. I tried to just shuffle a few lines to resolve this.

As in, this line was never true when I was running in the Continuous mode

jbossios:
You are right the filling of SF/InefficiencySFs was not right in the continuous case. Thanks for the fix!

kkrizka:
Can you remove changes that are only changing the tabulation?

Otherwise I think this looks good.

nhartman94:
> Can you remove changes that are only changing the tabulation?
> 
> Otherwise I think this looks good.

Ok, thank you for the comments. I'm working on undo-ing my indentation edits, so it'll take a few more commits to make sure I'm doing it all correctly, but I will comment again when I've completed this.

nhartman94:
Hi @kkrizka - I finished taking out the indentation edits, so hopefully it's easier for you to review my edits now. Please let me know if you need anything else from my side.

rateixei:
Hi, @jbossios and @kkrizka - is there anything missing to this MR for it to be merged? Thanks!

kkrizka:
Looks good to me, I think.

PR #1401, opened 2019-12-13
Muon container primary vertex check

robinnewhouse:
In MuonContainer.cxx, the primary vertex is accessed without checking if it is a null pointer. This caused infrequent but fatal failures when filling muon trackparams. A simple check is added so if the primary vertex is null, m_trkz0 is filled with -999.


kratsg:
I'm going to wait until dotnet/core#3988 is fixed/resolved before re-trying this PR.

PR #1402, opened 2019-12-18
added 'ANN_score' variable to the FatJet class

davide-melini:
The score of the adversarial NN used in ATL-PHYS-PUB-2018-014.
It was implemented recently in the BoostedJetTaggers package.

### DIFF
--- a/xAODAnaHelpers/FatJet.h
+++ b/xAODAnaHelpers/FatJet.h
@@ -47,6 +47,7 @@ namespace xAH {
       float  NTrimSubjets;
       int    NClusters;
       int    nTracks;
+      float ANN_score;
 
       // constituent
       int    numConstituents;


kratsg:
Is... are you planning on putting more changes in? This is just an update to a header file.

davide-melini:
Hi, no in principle, only this add. This would avoid making this change locally every time one sets-up our analysis code. I am sorry if this is not enough changes to make the merge request, or there is another way to do so which I missed.

kratsg:
> Hi, no in principle, only this add. This would avoid making this change locally every time one sets-up our analysis code. I am sorry if this is not enough changes to make the merge request, or there is another way to do so which I missed.

I'm not sure I understand really... this is adding a public member variable to the FatJet class. Why aren't you just creating your own FatJet class that inherits like

```
namespace YourCode {
  class FatJet : public xAH::FatJet
    {
      public:
        float ANN_score;
    }
};
```

and using that? These classes are fully public and meant to be extendable.

davide-melini:
it makes sense... Sorry for the very dumb MR then...

kratsg:
I just generally need to make sure xAH doesn't get cluttered with extra pieces of code that aren't used for anything.

kkrizka:
Actually I think it might be useful to add the ANN tagger tool, which now lives in ZprimeDM[1], to xAH itself. That way other users of xAH could benefit from it.

[1] https://gitlab.cern.ch/atlas-phys-exotics-dijetisr/ZprimeDM/blob/master/Root/ZprimeNtupler.cxx#L67

kratsg:
> Actually I think it might be useful to add the ANN tagger tool, which now lives in ZprimeDM[1], to xAH itself. That way other users of xAH could benefit from it.

Sure, that can certainly be done and anyone is welcome to add a PR for it.

PR #1404, opened 2019-12-19
Set jet uncertainties path

jbossios:
Fabrizio asked in the xAH mailing list to be able to set the "Path" property of JetUncertaintiesTool in JetCalibrator. This is now possible but the default remains the same.


fabnap:
Dear Jona, many thanks for this!
your solution works, it compiles & run and I could use the property "Path" at algorithm level.
Many thanks again

PR #1411, opened 2020-01-21
Updates to FatJetContainer

kkrizka:
A few updates to the `FatJetContainer` class responsible for ntupling fat jets.

* Move default values of track jet kinematic cuts into header for better visilibily.
* Save mass instead of energy for the fat jet four-vector. This is the more useful variable.
* Add matched truth jet kinematics. 
* Synchronize fat jet parent finding code between boson truth tagging and VR jet association.


kratsg:
@kkrizka , go ahead and fix the compilation errors?

PR #1412, opened 2020-01-24
Add configuration parameter for BTagSelectionTool

aemerman:
Makes the ErrorOnTagWeightFailure option of the BTaggingSelectionTool configurable in the BJetEfficiencyCorrector. The option defaults to true and the tool throws errors when attempting to tag jets without DL1 pu/pc/pb scores. Setting it to false makes the tool give a warning instead.
Needed for hh->4b analysis to be able to store low pt track jets (for VR overlap removal) which are occasionally missing DL1 scores.

### DIFF
--- a/xAODAnaHelpers/BJetEfficiencyCorrector.h
+++ b/xAODAnaHelpers/BJetEfficiencyCorrector.h
@@ -43,6 +43,7 @@ public:
   bool        m_useDevelopmentFile = true;
   bool        m_coneFlavourLabel = true;
   std::string m_systematicsStrategy = "SFEigen";
+  bool        m_errorOnTagWeightFailure = true;
 
   // allowed operating points:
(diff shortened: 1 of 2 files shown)


xAODAnaHelpers/BJetEfficiencyCorrector.h
```diff
@@ -43,6 +43,7 @@ class BJetEfficiencyCorrector : public xAH::Algorithm
   bool        m_useDevelopmentFile = true;
   bool        m_coneFlavourLabel = true;
   std::string m_systematicsStrategy = "SFEigen";
+  bool        m_errorOnTagWeightFailure = true;
```
kratsg: please add a comment for this option using `//!` or `///` to document it.

aemerman: I added a short description, but I can add more detail if you want

PR #1415, opened 2020-02-25
Fix large-R truth labeling for modern derivations

mswiatlo:
I'm not sure this is the right way to handle this.

But the truth decoration for large-R jets, in DAOD formats that have Truth3 type containers, requires a few more configurations than before, as described here: https://twiki.cern.ch/twiki/bin/view/AtlasProtected/JetUncertaintiesRel21Summer2019LargeR#Truth_labelling_via_BoostedJetTa

Probably this should be configurable, but I'm opening the MR with my current fix so we can discuss how to proceed

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -266,6 +266,10 @@ EL::StatusCode JetCalibrator :: initialize ()
     ANA_CHECK(m_SmoothedWZTagger_handle.setProperty("CalibArea" , "SmoothedWZTaggers/Rel21"));
     ANA_CHECK(m_SmoothedWZTagger_handle.setProperty("ConfigFile", "SmoothedContainedWTagger_AntiKt10LCTopoTrimmed_FixedSignalEfficiency50_MC16d_20190410.dat"));
     ANA_CHECK(m_SmoothedWZTagger_handle.setProperty("DSID", ei->mcChannelNumber()));
+    ANA_CHECK(m_SmoothedWZTagger_handle.setProperty( "TruthWBosonContainerName", "TruthBoson"));
+    ANA_CHECK(m_SmoothedWZTagger_handle.setProperty( "TruthZBosonContainerName", "TruthBoson"));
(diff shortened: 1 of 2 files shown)


jbossios:
Hi @mswiatlo 

It would probably be nice to introduce a boolean to decide to add those lines or not. I guess there are still out there derivations using TruthParticle container. Unless the addition of those lines has no impact when using TruthParticle container.

Best,
Jona

mswiatlo:
That's one good solution @jbossios . I'll see what others think, and switch to that if there's no objections.

Root/JetCalibrator.cxx
```diff
@@ -451,6 +455,9 @@ EL::StatusCode JetCalibrator :: execute ()
       if (this_TruthLabel == 5) isBjet = true;
       static SG::AuxElement::Decorator<char> accIsBjet("IsBjet"); // char due to limitations of ROOT I/O, still treat it as a bool
       accIsBjet(*jet_itr) = isBjet;
+      
+      SG::AuxElement::Accessor< int > accTruth("FatJetTruthLabel");
```
kratsg: what's this for?

kratsg:
> It would probably be nice to introduce a boolean to decide to add those lines or not. I guess there are still out there derivations using TruthParticle container. Unless the addition of those lines has no impact when using TruthParticle container.

I like this. As long as boolean is documented clearly enough, I don't see why this should be an issue.

kkrizka:
> It would probably be nice to introduce a boolean to decide to add those lines or not. I guess there are still out there derivations using TruthParticle container. Unless the addition of those lines has no impact when using TruthParticle container.
>

The addition of these lines does not have an impact if the derivation is using TRUTH1 style container. The [tagging tool checks which style truth container are available](https://gitlab.cern.ch/atlas/athena/blob/70405ec47024c6b9e60e87ce6fdfeeaeeb4cc22b/Reconstruction/Jet/BoostedJetTaggers/Root/JSSTaggerBase.cxx#L201) and uses the corresponding one. Thus I don't think an additional switch is necessary here. It might even complicate things if you are running over different versions of derivation.

Issue #1420, opened 2020-03-11
condor-only crashes in JetContainer::updateEntry


lhenkelm:
I recently started using the `xAH_run.py [...] condor [...]` driver to submit ntuple-->histos jobs to my institutes cluster. This largely works, but in every large submission, a small fraction of all jobs crashes.
The cause appears to be a std::out_of_range being thrown by some vector access somewhere inside of the `JetContainer` - specific virtual `JetContainer::updateParticle`. 

The curious aspect is that these crashes occur only when the job is distributed through condor.
I.e. when I log onto the same worker node and run the same job, over the same file, but using `xAH_run.py [...] direct [...]` instead, no crash occurs and the job processes the entire file successfully. 

Could these crashes somehow be due to how xAH_run wraps condor submission?

lhenkelm:
Some more detail:
* the exact command used to submit to condor is [here](https://gitlab.cern.ch/lhenkelm/histos_boosted_dijet_isr/blob/master/CondorHelpers/submit.sh#L32)
* this is how I run the local jobs that work fine:
`xAH_run.py --treeName outTree --files <ntuple_input.root> --config ../ZprimeDM/data/config_jetJet.py --submitDir test_1 direct`
* the files that the condor/out_of_range issue occurs with  ( as `<containername>/<filename>`) :

```user.lhenkelm.309712.e6584_s3126_r9364_p4004.NTUP_TAR_20200302_tree.root/user.lhenkelm.20713541._000001.tree.root
user.lhenkelm.309720.e6584_s3126_r10201_p4004.NTUP_TAR_20200302_tree.root/user.lhenkelm.20713591._000001.tree.root
user.lhenkelm.309720.e6584_s3126_r9364_p4004.NTUP_TAR_20200302_tree.root/user.lhenkelm.20713548._000001.tree.root
user.lhenkelm.309722.e6584_s3126_r10201_p4004.NTUP_TAR_20200302_tree.root/user.lhenkelm.20713592._000001.tree.root
user.lhenkelm.364375.e7048_s3126_r10724_p4004.NTUP_TAR_20200302_tree.root/user.lhenkelm.20713595._000001.tree.root
user.lhenkelm.364378.e7048_s3126_r10201_p4004.NTUP_TAR_20200302_tree.root/user.lhenkelm.20713557._000001.tree.root
[...]

PR #1422, opened 2020-03-13
Default `btagOpPoint::m_op` to None

lhenkelm:
This fixes the crashes in issue #1420 by giving the undefined `m_op` field in `xAH::JetContainer::btagOpPoint` the default value `xAH::Jet::BTaggerOP::None`.
It's difficult to test against this kind of thing, so I also added an exception being thrown 
if an unanticipated value is encountered in this field. 
Hopefully that will draw peoples attention if e.g. `btagOpPoint` gains a third c'tor  that forgets to set this default again. 

Finally, sorry for the indentation, but those files are a wild and inconsistent mix of tabs and spaces, so I just went with what my editor suggested.

### DIFF
--- a/Root/ClusterContainer.cxx
+++ b/Root/ClusterContainer.cxx
@@ -50,6 +50,5 @@ void ClusterContainer::FillCluster( const xAOD::CaloCluster* cluster ){
 void ClusterContainer::FillCluster( const xAOD::IParticle* particle )
 {
   ParticleContainer::FillParticle(particle);
-  const xAOD::CaloCluster* cluster=dynamic_cast<const xAOD::CaloCluster*>(particle);
   return;
 }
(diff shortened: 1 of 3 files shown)


kkrizka:
It might be better to do this initialization in the declaration of the class, instead of the construtor.

lhenkelm:
I don't disagree. I put it in the c'tors since the previously existing defaults for the other fields were set there too, and I wanted to stay consistent with the existing choices. But keeping all the defaults in the declaration would reduce repetition - a good thing.

kratsg:
@lhenkelm can you open a PR to fix this?

lhenkelm:
sorry for the slow response. Things are very busy at the moment. But yes, its not so much, I'll find some minutes eventually. Should I just move the default for m_op to the declaration, or for all fields where that is meaningfully possible?

kratsg:
> sorry for the slow response. Things are very busy at the moment. But yes, its not so much, I'll find some minutes eventually. Should I just move the default for m_op to the declaration, or for all fields where that is meaningfully possible?

all fields. we'd like to see the defaults in documentation :)

PR #1423, opened 2020-03-16
Jgeisen fat jet variables 2020 03 16

jgeisen:
Hi, I am performing an analysis on large-R jets with xAODAnaHelpers. I managed to add three additional variables that we have in our DAODs, but were previously not stored when writing those jets to a tree. Giordon asked me to merge these changes, as they might be interesting for other analysers.

### DIFF
--- a/Root/FatJetContainer.cxx
+++ b/Root/FatJetContainer.cxx
@@ -49,6 +49,9 @@ FatJetContainer::FatJetContainer(const std::string& name, const std::string& det
     m_NTrimSubjets      = new std::vector<float>();
     m_NClusters         = new std::vector<int>  ();
     m_nTracks           = new std::vector<int>  ();
+    m_ungrtrk500	= new std::vector<int>  ();
+    m_EMFrac		= new std::vector<float>();
+    m_nChargedParticles = new std::vector<int>();
 
   }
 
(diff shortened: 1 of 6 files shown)


jgeisen:
Hi, sorry I forgot to add a description of this MR. I am performing an analysis on large-R jets with xAODAnaHelpers. I managed to add three additional variables that we have in our DAODs, but were previously not stored when writing those jets to a tree. Giordon asked me to merge these changes, as they might be interesting for other analysers.

kratsg:
Will this crash if the variables are missing? Perhaps hiding it behind an extra flag is useful so people can turn on this extra stuff or not. I'm not sure here.

@jbossios ?

jgeisen:
Hi, when I run on a test sample that doesn't contain these variables, it does not crash, because I use the accessors for ungrtrk500 and nChargedParticles which thus fill -999 if the variable is not present/available for whatever reason, while the function to get the EMFrac returns either -1 or 0 by default if something is not working. I just reconfirmed with a local test run on an older sample that this works.

mswiatlo:
Hi,

Why did this add pvLocation without a default? This causes problems for users who now have to specify information they may not need to care about...

kratsg:
> Why did this add pvLocation without a default? This causes problems for users who now have to specify information they may not need to care about...

Oh, that's  a good point... sorry for the oversight. I'll add in a default.

jgeisen:
Hi,
indeed I forgot about this, sorry. Thanks for pointing it out. Also thanks Giordon for taking care of it!

PR #1424, opened 2020-03-17
Fix crash when running multiple instances of ElectronEfficiencyCorrector

jbossios:
It is probably not the best solution but it works, if you have a suggestion please let me know.


kratsg:
If it's in the ToolStore, you never need to delete it...

jbossios:
That is the issue, it is not in the ToolStore.

jbossios:
For example:
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/ElectronEfficiencyCorrector.cxx#L139

kratsg:
might I ask that we switch this to a toolhandle instead? We need toolhandles to migrate to AnaAlgorithms anyway.

jbossios:
OK, will see if I can find some time to do that his week.

jbossios:
Hi @kratsg 

I'm switching to tool handles but I need first to understand something. I don't know why inside executeSF() the tools are used only when they have not been used before (through isToolAlreadyUsed()). Do you know why is that needed? executeSF() is called for every systematic, does this mean that executeSF is effectively only used for the first systematic?

Best,
Jona

kratsg:
`isToolAlreadyUsed` is only true _if_ the tool name was used, not if the tool itself has been used. This was supposed to be the check that the tool name isn't already taken. If it is, likely because the tool was already made, you use it. It seems like there might be odd logic from what you're saying?

jbossios:
Thanks for the clarification on isToolAlreadyUsed. I found it weird it is used in execute() and not in initialize(). I guess the intention is not to get, for instance, reco SFs twice when running two instances of ElectronEfficiencyCorrector? Same thing for PID SFs, maybe someone (like me) needs SFs for two different combinations of PID+Iso for the same PID WP, in that case, we don't want to get the PID SFs twice. I'm not sure how to avoid that now using tool handles. Do you?

kratsg:
> I'm not sure how to avoid that now using tool handles. Do you?

Nope. The only way I can see getting around this is to perhaps check if the algorithm is already scheduled... we have similar checks on this level. e.g.

```
if ( this->numInstances() > 1 ) {
  ANA_MSG_INFO( "\t An algorithm of the same type has been already used " << numInstances() << " times" );
}```

jbossios:
But we want the possibility to run this algorithm several times. For instance, we need to be able to retrieve SFs for the following combinations:

ID Medium + Isolation FCLoose
ID Medium + Isolation FCTight
ID Truth + Isolation FCLoose
ID Truth + Isolation FCTight

Which means we need four instances of ElectronEfficiencyCorrection, but we don't need to retrieve the ID SFs more than once. So we need to check if the same algorithm with the same ID (or reco) configuration was used to not retrieve those SFs, but retrieve the ID + Isol (and + trig) combination.

kratsg:
> Which means we need four instances of ElectronEfficiencyCorrection, but we don't need to retrieve the ID SFs more than once. So we need to check if the same algorithm with the same ID (or reco) configuration was used to not retrieve those SFs, but retrieve the ID + Isol (and + trig) combination.

Ahh, good point. Then yes, you just need to check if the tool has been used then :) Which is probably why it was done this way.

kratsg:
Perhaps as this is likely easier solved by switching to AnaAlgorithms which allows tools to be scheduled in addition to algorithms -- this will be easier to deal with rather than right now. So how about we merge in the current fix?

jbossios:
Sounds good to me. 
I think it would also be good to make this algorithm to support multiple PID and Isolation WPs, in that way, every combination is handled by the algorithm and the user does not need to deal with this. For instance, at the moment, I have to set Reconstruction WP to "" to avoid a crash for other instance than the first one.

kratsg:
> I think it would also be good to make this algorithm to support multiple PID and Isolation WPs, in that way, every combination is handled by the algorithm and the user does not need to deal with this. For instance, at the moment, I have to set Reconstruction WP to "" to avoid a crash for other instance than the first one.

probably we should file an issue for it, but I won't have time to get to this at the moment.

PR #1427, opened 2020-03-24
Allow to reject events with muons/electrons

jbossios:
This would allow, for instance, to select events with muons (electrons) but no electrons (muons).


kkrizka:
Can this behaviour be made consistent among all particle selectors? I don't think it is good if half of `m_pass_max` mean one thing and the other half another.

Issue #1431, opened 2020-04-01
Filling branches crashes in multiple locations when a primary vertex cannot be found


robinnewhouse:
We've run into a problem when running over a newly processed set of data. There seem to be a couple events missing primary vertices, but that's a different issue.

The problem I've found here is that the ntuple maker crashes when trying to access a null pointer and it would be great if the xAH could keep on running.

So I've seen this happen in `ElectronContainer.cxx` and `VertexContainer.cxx`.

This can be reproduced by filling the vertices with `"primary"` in the `vertexDetailString` on the dataset: `data18_13TeV:DAOD_RPVLL.20361820._000157.pool.root.1`

I've prepared a merge request that I will attach to this issue.

#### Running details
* AnalysisBase 21.2.90
* Running on lxplus using `xAH_run.py`
* Waxing Crescent

kratsg:
> * Waxing Crescent

you know the names of the algorithms are randomly generated if you don't set one?

PR #1432, opened 2020-04-01
Prevent crashes when unable to find a primary vertex

robinnewhouse:
Addresses #1431

These changes keep the ntuple filler from crashing in the event that a primary vertex cannot be found.
This pull request only affects `VertexContainer.cxx` and `ElectronContainer.cxx` though there may be other sources of crashes that I have not yet tested.


kratsg:
Can you check through other containers and get them protected in this PR too?

kratsg:
Ping me when you're ready for this to get merged in.

robinnewhouse:
> Can you check through other containers and get them protected in this PR too?

Hi @kratsg , I've added a couple more checks and fill the relevant branches with `-999.0` placeholders when the PV is null. I've run a test on our framework using as many algorithms as we can and a data sample with a few missing PVs. These tests appear to pass.

PR #1433, opened 2020-04-01
Clean up available b-taggers.

kkrizka:
- The `BTaggingSelectionTool` now [requires](https://gitlab.cern.ch/atlas/athena/-/blob/44edce0e78b9d8901ff0722637edc8690c4ababe/PhysicsAnalysis/JetTagging/JetTagPerformanceCalibration/xAODBTaggingEfficiency/Root/BTaggingSelectionTool.cxx#L104) a `MinPt` property
- Removes b-tagger definitions that are no longer supported by the flavour tagging group. The 
currently supported (and calibrated) definitions are DL1, DL1r, DL1rmu and MV2c10 with fixed cut WPs.
- Remove deprecated `m_sfFTag*` info switch settings
- Remove deprecated `btagOpPoint::m_old` member
- Seeing `Jet::BTaggerOP::None` setting when reading an ntuple now throws an error


kkrizka:
This commit also does some of the changes in the old #1282 .

kkrizka:
Also moving a question from the closed !1420 here.

@lhenkelm, why is your code trying to load b-tagging information if none are defined? The crash you reported occurred in the [loop over m_btags](https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetContainer.cxx#L1373) which is [constructed from the info switch](https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/JetContainer.cxx#L360). Thus if you don't define any b-taggers, then you should never have this problem. Am I missing something? Are you maybe using an "unsupported" tagger definition?

It seems like something is wrong in the configuration. In this PR, I change the case of to throw an error on `BTaggerOP::None` since I don't believe this situation should occur and could be indicative of a misconfigured xAH. But it would be good to get your feedback in case I misunderstood your problem.

lhenkelm:
repeating the answer from #1420: 
I am not trying to load them. I am passing empty/meaningless strings into the info switch. Some context to explain why I would do such a thing:
I use xAH in the context of the ZprimeDM framework.
From this framework I inherit this python config (the link points to the version of the code I used when debugging this issue).
As you can see, the authors of the config intended it to be used such that specifying a False-ish object (None, empty container, etc.) in the switches controlling what btagging data to read
overrides that switch with the default setting (which in this case is some valid btagging working point, i.e. no good for my use case). So instead I pass lists containing an empty string,
which avoids the override, but still does not result in btags being (attempted to) read out.

lhenkelm:
So, if you want to throw in the switch, I think the ZprimeDM way configuring these things would need to be changed more substantially (though I am not the expert on that). But from a purely xAH perspective, I don't think there is a problem with throwing instead of skipping.

kkrizka:
@lhenkelm  Let's continue the conversation here.

It sounds like you are adding a bogus `jetBTag_*` to the detail string for the info switch, which I don't think is a good idea. It is an invalid definition.

Why not just update your config to skip b-tagging if the `doModes` list is empty?

lhenkelm:
well, the historic motivation for me, personally is that I wanted to keep things as close as possible to the way the config looks in ZprimeDM:master, and the invalid definition hack was the most convenient way to achieve that. But as I wrote above, I think if xAH starts rejecting invalid strings in infoswitches, then also the shared version of the config would be updated.

kratsg:
@kkrizka are you happy with this to get merged in? The discussion here seems not fully related to the changes in the PR.

kratsg:
ping @kkrizka

kkrizka:
Yes, fine with me.

PR #1441, opened 2020-05-04
Add support for more convenient JER pseudodata smearing

balunas:
This adds an option to JetCalibrator (`m_mcAndPseudoData`) which will run both nominal MC and pseudodata JER smearing in a single run. This eliminates the need to run the framework twice to compute JER systematics. Using this option will create all of the normal systematic-varied jet containers, plus additional ones for JER NPs with the pseudodata smearing applied in place of the MC smearing (the output names are the same, but with `_PD` appended). These additional "NPs" are also written to the output list for downstream use.

To avoid duplication of code, I've separated out some of the functionality into helper functions. This is purely for organization and readability; existing functionality should be unchanged.

**Obligatory warning:** These aren't normal nuisance parameters and shouldn't be blindly stuffed into some statistics package along with all the rest. Be sure to follow the approriate Jet/Etmiss recommendations carefully!

Tagging @chnzhangrui; this is based on some of his work.

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -285,6 +285,7 @@ EL::StatusCode JetCalibrator :: initialize ()
     if(m_pseudoData) {
       ANA_MSG_INFO("Input treated as pseudo-data");
       ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("IsData",true));
+      initializePDJER();
     }
     else {
       ANA_CHECK( m_JetUncertaintiesTool_handle.setProperty("IsData",!isMC()));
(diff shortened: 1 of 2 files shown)


kratsg:
> This adds an option to JetCalibrator (`m_mcAndPseudoData`) which will run both nominal MC and pseudodata JER smearing in a single run. This eliminates the need to run the framework twice to compute JER systematics.

Does this mean, previously, you were scheduling two algorithms of JetCalibrator or re-running all algorithms twice somehow? I guess you're just essentially doing an additional entry in a for-loop somewhere, which seems reasonable to me.

balunas:
Yeah, I believe that's what people have been having to do. In practice I think most analyses have been using the stronger reduction for the JER NPs which doesn't require this, but now there's more demand for this since we're doing more precision/combination stuff. If anything this is saving CPU time since it only runs twice over the JER NPs rather than all of them.

balunas:
Also, I had set an incorrect parameter for when you're doing only pseudodata, that last commit fixes that

PR #1452, opened 2020-08-19
feat: Add ChargedFraction to jet cleaning variables

### LINKED ISSUE #1445: Add fraction of jet pT from tracks to jet container?
For jet cleaning studies I'm interested in looking at a quantity ("the charge fraction") which would be the scalar sum of the pT of tracks associated with a jet divided by the jet pT. To get this into the `JetHistsAlgo` and the jet `TreeAlgo` this would need to be added to the jet container when the "clean" detail string is used. `m_SumPtTrkPt1000` and `m_SumPtTrkPt500` already exist (if the "trackAll" detail string is used)

https://github.com/UCATLAS/xAODAnaHelpers/blob/b3feeb1e34782478900bf060dd5a27eb8d8ed2d7/Root/JetContainer.cxx#L118-L121

however, I think that I should be instead checking to see if the jet is decorated with `TrackSumPt` (as is `JETM1`) and creating the charge fraction from this. Is there a better way to do what I'm interested in, and if not would there be any objections to a PR adding this in?

cc @mattleblanc @kkrizka @kratsg @jbossios

matthewfeickert:
Resolves #1445 

For jet cleaning studies, add the charge fraction (the scalar sum of the pT of tracks from the primary vertex associated with a jet divided by the jet pT) to the variables that are associated with the `clean` detail string. Given reasons brought up by @mattleblanc in discussion of Issue #1445 RE: the tracks of interest, `m_ChargedFraction` is calculated in a similar manner to how `m_SumPtTrkPt500PV` is instead of using the derivation dependent `DFCommonJets_TrackSumPt` decoration.

## Suggested squash and merge commit log

```
* Add ChargedFraction to the jet cleaning variables
   - Scalar sum of the pT of tracks from the primary vertex associated with a jet divided by the jet pT
* Add ChargedFraction to jet cleaning variables set of jet histograms
```

### DIFF
--- a/Root/JetContainer.cxx
+++ b/Root/JetContainer.cxx
@@ -2355,9 +2355,6 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
       static SG::AuxElement::ConstAccessor<int> LArBadHVNCell ("LArBadHVNCell");
       safeFill<int, int, xAOD::Jet>(jet, LArBadHVNCell, m_LArBadHVNCell, -999);
 
-      static SG::AuxElement::ConstAccessor<float> ChargedFraction ("ChargedFraction");
-      safeFill<float, float, xAOD::Jet>(jet, ChargedFraction, m_ChargedFraction, -999);
-
       static SG::AuxElement::ConstAccessor<float> OotFracClus5 ("OotFracClusters5");
       safeFill<float, float, xAOD::Jet>(jet, OotFracClus5, m_OotFracClusters5, -999);
 
@@ -2658,6 +2655,7 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
 
     } // trackPV || JVT
 
+    // Can't use safeFill as ChargedFraction is not derivation level decoration
     if ( m_infoSwitch.m_clean && pvLocation >= 0 ) {
       if ( sumPt500.isAvailable( *jet ) ) {
         m_ChargedFraction->push_back( sumPt500( *jet )[pvLocation] / jet->pt() ); // units cancel out
--- a/Root/JetContainer.cxx
+++ b/Root/JetContainer.cxx
@@ -1140,6 +1140,7 @@ void JetContainer::updateParticle(uint idx, Jet& jet)
         jet.N90Constituents           =m_N90Constituents           ->at(idx);
         jet.LArBadHVEFrac             =m_LArBadHVEnergyFrac        ->at(idx);
         jet.LArBadHVNCell             =m_LArBadHVNCell             ->at(idx);
+        jet.ChargeFrac                =m_ChargeFrac                ->at(idx);
         jet.OotFracClusters5          =m_OotFracClusters5          ->at(idx);
         jet.OotFracClusters10         =m_OotFracClusters10         ->at(idx);
         jet.LeadingClusterPt          =m_LeadingClusterPt          ->at(idx);
--- a/Root/JetContainer.cxx
+++ b/Root/JetContainer.cxx
@@ -2659,10 +2659,8 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
 
       static SG::AuxElement::ConstAccessor< float > ChargedFraction("ChargedFraction");
       static SG::AuxElement::Decorator< float > chargedFractionDecor("ChargedFraction");
-      if ( chargedFractionDecor.isAvailable( *jet ) ) {
-        safeFill<float, float, xAOD::Jet>(jet, ChargedFraction, m_ChargedFraction, -999);
-      } else { // calculate and decorate
-
+      if ( !chargedFractionDecor.isAvailable( *jet ) ) {
+        // calculate and decorate
         if ( sumPt500.isAvailable( *jet ) ) {
           m_ChargedFraction->push_back( sumPt500( *jet )[pvLocation] / jet->pt() ); // units cancel out
         } else {
@@ -2670,6 +2668,8 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
         }
 
         chargedFractionDecor( *jet ) = m_ChargedFraction->back();
+      } else {
+        safeFill<float, float, xAOD::Jet>(jet, ChargedFraction, m_ChargedFraction, -999);
       }
     } // clean
 
--- a/Root/JetContainer.cxx
+++ b/Root/JetContainer.cxx
@@ -2355,6 +2355,9 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
       static SG::AuxElement::ConstAccessor<int> LArBadHVNCell ("LArBadHVNCell");
       safeFill<int, int, xAOD::Jet>(jet, LArBadHVNCell, m_LArBadHVNCell, -999);
 
+      static SG::AuxElement::ConstAccessor<float> ChargeFrac ("ChargeFrac");
+      safeFill<float, float, xAOD::Jet>(jet, ChargeFrac, m_ChargeFrac, -999);
+
       static SG::AuxElement::ConstAccessor<float> OotFracClus5 ("OotFracClusters5");
       safeFill<float, float, xAOD::Jet>(jet, OotFracClus5, m_OotFracClusters5, -999);
 
--- a/Root/JetContainer.cxx
+++ b/Root/JetContainer.cxx
@@ -39,6 +39,7 @@ JetContainer::JetContainer(const std::string& name, const std::string& detailStr
       m_N90Constituents           =new std::vector<float>();
       m_LArBadHVEnergyFrac        =new std::vector<float>();
       m_LArBadHVNCell             =new std::vector<int>();
+      m_ChargeFrac                =new std::vector<float>();
       m_OotFracClusters5          =new std::vector<float>();
       m_OotFracClusters10         =new std::vector<float>();
       m_LeadingClusterPt          =new std::vector<float>();
@@ -466,6 +467,7 @@ JetContainer::~JetContainer()
       delete m_N90Constituents;
       delete m_LArBadHVEnergyFrac;
       delete m_LArBadHVNCell;
+      delete m_ChargeFrac;
       delete m_OotFracClusters5;
       delete m_OotFracClusters10;
       delete m_LeadingClusterPt;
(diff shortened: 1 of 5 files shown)


Root/JetContainer.cxx
```diff
@@ -1135,7 +1138,7 @@ void JetContainer::updateParticle(uint idx, Jet& jet)
         jet.AverageLArQF              =m_AverageLArQF              ->at(idx);
         jet.BchCorrCell               =m_BchCorrCell               ->at(idx);
         jet.N90Constituents           =m_N90Constituents           ->at(idx);
-        jet.LArBadHVEFrac             =m_LArBadHVEnergyFrac       ->at(idx);
+        jet.LArBadHVEFrac             =m_LArBadHVEnergyFrac        ->at(idx);
```
mattleblanc, line 1141: Is ChargeFrac missing here?

mattleblanc (commented):
Just the one point to clarify ...

Could you point at a plot of the ChargeFrac distribution somewhere made from your test file quickly, just so we know it looks ~as expected?

🍻 MLB

Root/JetContainer.cxx
matthewfeickert: Yup, but intentionally as I was pushing things before bed and wanted everything to compile as I hadn't added the necessary change to `Jet.h`. But good eye too. :)

matthewfeickert:
> Could you point at a plot of the ChargeFrac distribution somewhere made from your test file quickly, just so we know it looks ~as expected?

Yeah, I'll add that as part of the information before I request review.

xAODAnaHelpers/JetHists.h
```diff
@@ -53,6 +53,7 @@ class JetHists : public IParticleHists
     //TH1F* m_LArQmean;
     //TH1F* m_LArBadHVEFrac;
     //TH1F* m_LArBadHVNCell;
+    TH1F* m_ChargeFrac;             //!
```
matthewfeickert, line 56: While not directly relevant to this PR, I noticed that some jet cleaning variables like the LArQ Frac (`m_LArBadHVEFrac`) are commented out here. However, `git blame` shows that they were commented out when they were introduced first in PR #619 (4 years ago :slightly_smiling_face:).

I'm going to assume that there is no technical reason why they should remain commented (if so, they should be removed anyway) and after this PR goes in I can open up another to uncomment them and delete others that really aren't used. @kratsg @kkrizka Do you have any thoughts here?

kratsg: I'd go ahead and do that.

matthewfeickert:
A general comment for thought @mattleblanc: `ChargeFrac` is a bad naming choice IMO as it makes one immediatley think of something to do with the jet charge rather than the pT of the tracks. What are thoughts on something like `ChargeFracPt`?

Root/JetHists.cxx
```diff
@@ -39,6 +39,7 @@ StatusCode JetHists::initialize() {
 
     //m_LArBadHVEFrac              =book(m_name, "LArBadHVEFrac",              m_titlePrefix+" jet LAr Bad HV Energy Fraction", 120,   0,    1);
     //m_LArBadHVNCell              =book(m_name, "LArBadHVNCell",              m_titlePrefix+" jet LAr Bad HV N_{cells}",       120,  -0.5,499.5);
+    m_ChargeFrac  = book(m_name, "ChargeFrac",      m_titlePrefix+" Charge Fraction", 120, 0, 1);
```
matthewfeickert, line 42: Note to self: Use better axis labels
```suggestion
    m_ChargeFrac  = book(m_name, "ChargeFrac",      m_titlePrefix+" Fraction of P_{T} from charged tracks", 120, 0, 1);
```

Root/JetContainer.cxx
```diff
@@ -2649,6 +2658,12 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
 
     } // trackPV || JVT
 
+    if ( m_infoSwitch.m_clean && pvLocation >= 0 ) {
+      if ( sumPt500.isAvailable( *jet ) ) {
+        m_ChargeFrac->push_back( sumPt500( *jet )[pvLocation] / jet->pt() ); // units cancel out
```
matthewfeickert, line 2663: @mattleblanc I might have to follow up with you later on, as I'm doing something dumb in this PR and getting empty `ChargeFrac` histogram while the rest of the histograms in the `clean` infoSwitch are getting filled properly for a test run.

mattleblanc:
> A general comment for thought @mattleblanc: `ChargeFrac` is a bad naming choice IMO as it makes one immediatley think of something to do with the jet charge rather than the pT of the tracks. What are thoughts on something like `ChargeFracPt`?

I think it should be called `ChargedFraction` to be consistent with what is used in other JES/JER code (in the GSC & etc.).

Root/JetContainer.cxx
mattleblanc: Hm weird. And `sumPt500` is available?

matthewfeickert: > And `sumPt500` is available?

Good question that I had to verify for myself. The answer is yes as it is just a ConstAccessor for the `SumPtTrkPt500` decoration

https://github.com/UCATLAS/xAODAnaHelpers/blob/d3a39d971b1580ab8b23acd50ccbf225a174ca8b/Root/JetContainer.cxx#L2549-L2557

and setting up `AthDerivation` and then running

```
checkFile.py "${data_path}" | grep AntiKt4EMTopoJets | tee AntiKt4EMTopoJets_info.txt
```

where `"${data_path}"` is pointing to a file from `data16_13TeV:data16_13TeV.periodA.physics_Main.PhysCont.DAOD_JETM1.grp16_v01_p4061` shows

```
$ grep "SumPtTrkPt500" AntiKt4EMTopoJets_info.txt | awk -F " " '{print $NF}'
AntiKt4EMTopoJetsAuxDyn.SumPtTrkPt500
```

So I'm a bit confused as there has to be a primary vertex and so `pvLocation >= 0` should be `true`.

matthewfeickert: @kkrizka @kratsg Thoughts on what I'm messing up here?

matthewfeickert: If I do some cout debugging (sorry)

```c++
    std::cout << "* pvLocation: " << pvLocation << '\n';
     if ( m_infoSwitch.m_clean && pvLocation >= 0 ) {
       if ( sumPt500.isAvailable( *jet ) ) {
         m_ChargedFraction->push_back( sumPt500( *jet )[pvLocation] / jet->pt() ); // units cancel out
       } else { m_ChargedFraction->push_back( -999 ); }
      std::cout << "* m_ChargedFraction->back(): " << m_ChargedFraction->back() << '\n';
     } // clean
```

things are working as expected

```
* pvLocation: 0
* m_ChargedFraction->back(): 0.635272
* pvLocation: 0
* m_ChargedFraction->back(): 0.473161
* pvLocation: 0
* m_ChargedFraction->back(): 0.301372
* pvLocation: 0
* m_ChargedFraction->back(): 0.178198
* pvLocation: 0
* m_ChargedFraction->back(): 0.5505
* pvLocation: 0
* m_ChargedFraction->back(): 0.511196
* pvLocation: 0
* m_ChargedFraction->back(): 0.0188797
```

so I'm messing up the histogram filling apparently.

matthewfeickert:
@mattleblanc @kkrizka @kratsg: If I understand things correctly, I should remove the additions of `ChargedFraction` to JetHists as they are impossible to fill. This is because `ChargedFraction` is not a derivation level decoration (hence this PR exists) so it will never be available as a decoration already to simply fill. Additionally, as `ChargedFraction` relies on the ConstAccessor for the `SumPtTrkPt500` decoration

https://github.com/UCATLAS/xAODAnaHelpers/blob/d3a39d971b1580ab8b23acd50ccbf225a174ca8b/Root/JetContainer.cxx#L2549-L2555

along with information on the primary vertex location (`pvLocation`)

https://github.com/UCATLAS/xAODAnaHelpers/blob/d3a39d971b1580ab8b23acd50ccbf225a174ca8b/Root/JetContainer.cxx#L2282

and information on the primary vertex is not available in the JetHists `execute` methods.

So given the nature of these decorations unless they're included at the derivation level it isn't possible to plot them with the `JetHistsAlgo`, yes? If that's correct then I'll just drop the changes to `JetHists.{h, cxx}` in this PR.

@mattleblanc Given this, the current state of the PR is able to fill the tree fine though. However, with the definition we're using at the moment (scalar sum of the pT of tracks from the primary vertex associated with a jet divided by the jet pT) the tree has some tail values that go far beyond 1

ChargedFraction_full_range

ChargedFraction_tail_range

I was expecting that the Charged Fraction distributions would be more similar in bounding behavior to the LArQ Frac (where it actually truncates at 1). Do you have general thoughts on this? There may be obvious answers to these questions, but given the time of the night I'm writing this it is escaping me.

Root/JetContainer.cxx
matthewfeickert: This should be resolved by the discussion below: https://github.com/UCATLAS/xAODAnaHelpers/pull/1452#issuecomment-678061744

Root/JetHists.cxx
```diff
@@ -39,6 +39,7 @@ StatusCode JetHists::initialize() {
 
     //m_LArBadHVEFrac              =book(m_name, "LArBadHVEFrac",              m_titlePrefix+" jet LAr Bad HV Energy Fraction", 120,   0,    1);
     //m_LArBadHVNCell              =book(m_name, "LArBadHVNCell",              m_titlePrefix+" jet LAr Bad HV N_{cells}",       120,  -0.5,499.5);
+    m_ChargedFraction = book(m_name, "ChargedFraction", m_titlePrefix+" Sum of charged tracks p_{T}/jet p_{T}", 120, 0, 2);
```
matthewfeickert, line 42: The booking range for the histogram is set to `2` instead of `1` as there are some events in which the `ChargedFraction` is greater than `1`. This is already rare, and beyond `2` these number drop to single rare events.

Root/JetContainer.cxx
```diff
@@ -2649,6 +2655,24 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
...
+      static SG::AuxElement::Decorator< float > chargedFractionDecor("ChargedFraction");
+      if ( !chargedFractionDecor.isAvailable( *jet ) ) {
+        // calculate and decorate
+        if ( sumPt500.isAvailable( *jet ) ) {
+          m_ChargedFraction->push_back( sumPt500( *jet )[pvLocation] / jet->pt() ); // units cancel out
+        } else {
+          m_ChargedFraction->push_back( -999. );
+        }
+
+        chargedFractionDecor( *jet ) = m_ChargedFraction->back();
```
matthewfeickert, line 2670: Should this be moved to a section of `JetSelector.cxx` instead? Or it is okay here?

kratsg: fine here

kratsg: note it means that other frameworks who don't use jetcontainers won't have access to these variables. that's probably ok.

Root/JetHists.cxx
kratsg: maybe add that as a comment in the code.

matthewfeickert: Done.

PR #1459, opened 2020-10-27
Pass primaryVertex to FillMuonsUser() and FillElectronsUser()

jbossios:
I guess this probably is going to break those codes using FillMuonsUser() and FillElectronsUser() but I think we can not avoid that...


kratsg (approved):

jbossios:
Making WIP just for precautions until I fix I crash I see while testing locally....

PR #1467, opened 2020-12-01
Fix listTrigChains lepton branch and allow to turn ON ApplyRelPt in OveralRemover

jbossios:
The fix on listTrigChains lepton branch allows to use it to know which chains are matched to the offline leptons.


jbossios:
@kratsg 

Friendly reminder about this PR.

PR #1468, opened 2020-12-07
fix: pvLocation default to 0 for FatJet

kratsg:
This resolves a minor regression in functionality/behavior with the introduction of `pvLocation` in #1423.

/cc @mswiatlo


mswiatlo:
Thanks!

PR #1471, opened 2021-01-12
Enable event-level lepton trigger SFs

jbossios:
This PR adds the option of decorating each lepton with a global (event-level) trigger SF which is calculated using all leptons from the input container (so it is the same for each lepton). These changes are technically only needed for muons, since for the electron case we could calculate the global SF ourselves with the information already available but I added this option as well to electrons to keep things consistent. This option is disabled by default and per-lepton trigger SFs are used. Global trigger SFs can be enabled by setting m_usePerElectronTriggerSFs and m_usePerMuonTriggerSFs to false.

Adding also a missing OutputLevel setting in one of the tools.


kratsg (approved):
very nice

PR #1490, opened 2021-04-26
Support Sh_2210 when using m_setMapIndex

jbossios:
Support Sherpa 2.2.10 samples when using m_setMapIndex


jbossios:
@kratsg I'm not sure what is the issue with the crashing CI test.

kratsg:
It's crashing because it's from a fork. That's all.

PR #1495, opened 2021-06-14
Adding jet PFO information to tree

rbarrue:
Adding PFO-related information for PFlow jets - SumPtChargedPFOPt500PV and fCharged, as well as adding the necessary configuration string in TreeAlgo ("chargedPFOPV").

Tagging @jbossios.

### DIFF
--- a/Root/JetContainer.cxx
+++ b/Root/JetContainer.cxx
@@ -171,6 +171,11 @@ JetContainer::JetContainer(const std::string& name, const std::string& detailStr
     }
   }
 
+  // chargedPFOPV 
+  if ( m_infoSwitch.m_chargedPFOPV ) {
+    m_SumPtChargedPFOPt500PV = new std::vector<float> ();
+    m_fCharged = new std::vector<float> ();
+  }
 
   // allTrack
   // trackAll or trackPV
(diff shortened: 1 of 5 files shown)


rbarrue:
I followed the standards used to include other jet variables, as well as the fCharged definition in JetCalibTools: SumPtChargedPFOPt500PV/ pt @ constituent scale.

When I then go look at the actual fCharged distributions from this implementation, I get jets that have values of fCharged > 1. Before I start looking into the reconstruction-level reason this might happen, I want to determine if there's anything wrong with my implementation.

jbossios:
The implementation looks good and fCharged can be sometimes larger than 1.

Root/JetContainer.cxx
```diff
@@ -2646,6 +2686,22 @@ void JetContainer::FillJet( const xAOD::IParticle* particle, const xAOD::Vertex*
...
       } // trackPV
 
+      
+      if ( m_infoSwitch.m_chargedPFOPV && pvLocation >= 0) {
+
+        if ( sumPtChargedPFO500.isAvailable( *jet ) ) {
+          m_SumPtChargedPFOPt500PV->push_back( sumPtChargedPFO500( *jet)[pvLocation] );
+
+          xAOD::JetFourMom_t jetconstitP4 = jet->jetP4("JetConstitScaleMomentum");
+          m_fCharged->push_back( sumPtChargedPFO500( *jet)[pvLocation] / jetconstitP4.Pt());
```
jbossios, line 2696: @rbarrue could you please create a new PR adding a protection for cases where the jet pt is zero? if jetconstitP4.Pt() is zero we should simply fill -999

rbarrue: Done in https://github.com/UCATLAS/xAODAnaHelpers/pull/1498/

PR #1499, opened 2021-07-01
feat: useXAOD config in c.algorithm()

kratsg:
Add a way to disable xAOD if needed when creating an algorithm.

### DIFF
--- a/python/config.py
+++ b/python/config.py
@@ -92,10 +92,11 @@ class Config(object):
       for k,v in options.items():
         if k in ['m_msgLevel', 'm_name']: continue
         self._set_algo_attribute(alg_obj, k, v, className, algName)
+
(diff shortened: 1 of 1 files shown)


python/config.py
```diff
@@ -95,6 +95,7 @@ def algorithm(self, className, options):
     else:
       raise TypeError("Algorithm {0:s} is not an EL::Algorithm or EL::AnaAlgorithm. I do not know how to configure it. {1}".format(className, parents))
 
+    alg_obj.useXAOD(useXAOD)
```
mattleblanc, line 98: This should be 

```
    alg_obj.setUseXAODs(useXAOD)
 ``` 

🍻 MLB

mattleblanc (commented):
This seems to work once the change I noted is made.

PR #1512, opened 2021-12-03
throw if TTree::Fill() reports a failure to write

### LINKED ISSUE #1511: `HelpTreeBase::Fill` does not propagate failure to Fill/write to disk
this is troublesome e.g. for grid jobs, which will not fail (thus not allowing them to be re-tried and hiding the existence of a problem). 
but produce trees containing corrupted baskets that can not be read.

lhenkelm:
fixes #1511 by throwing an informative `std::runtime_error` if the `TTree::Fill()` call fails and reports it in its return value.
The reasons to throw rather than pass through the returrn value:
  * throwing leaves the signature of `void HelpTreeBase::Fill()` unchanged
  * the exception alerts users to the error unless they explicitly anticipate & handle it in a try/catch block. Conversely it is easy to forget to check the returned value for the special value `-1`

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -1,5 +1,7 @@
 // c++ include(s):
 #include <iostream>
+#include <exception>
+#include <sstream>
 
 // EDM include(s):
 #include "xAODBTagging/BTagging.h"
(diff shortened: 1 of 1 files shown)


lhenkelm:
@kratsg could you please give this a look in a spare minute? It is blocking me from producing ntuples for my analysis team.

lhenkelm:
Thanks!

PR #1517, opened 2022-01-21
Correct MetContainer, branches incorrectly created

flumeware:
The SigClus branches are being created with type tag `/F` which should be `F`. At present
```
TBranch::TBranch          WARNING Extra characters after type tag '/F' for branch 'metFinalClusOverSqrtSumEt'; must be one character.
TBranch::TLeaf            ERROR   Illegal data type for metFinalClusOverSqrtSumEt/metFinalClusOverSqrtSumEt//F
```
 is produced when you try to add the branches to the tree. 

Removes the `/` which also makes it consistent with the other lines in the file. With this change applied the branches are added.

### DIFF
--- a/Root/MetContainer.cxx
+++ b/Root/MetContainer.cxx
@@ -122,10 +122,10 @@ void MetContainer::setBranches(TTree *tree)
   }
 
   if ( m_infoSwitch.m_sigClus ) {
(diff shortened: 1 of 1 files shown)


kratsg (approved):
nice catch

jburzy01:
Can this fix also be propagated to r22/master?

PR #1518, opened 2022-01-25
Correct setBranch for significance branches, /F is incorrect should be F

flumeware:
Same change as #1517 but applied to `r22/master`


kratsg:
@jbossios you're the maintainer here.

jbossios:
Sorry for the delay, was too busy with other stuffs. Merging...

PR #1519, opened 2022-01-25
Darij/jet cleaning

Darij-Starko:
included LooseBadLLP cut in jet-cleaning


jbossios:
Do you mind changing this to add this WP to another (infoSwitch) option? Otherwise, you will be populating the cleanLight option for everyone, i.e. changing the current behaviour and increasing the output size for everybody while this is only needed for a very particular case.

PR #1524, opened 2022-02-16
Update Root/MuonCalibration.cxx: add momentum calibration mode property

hollypacey:
In line with the MCP precision recommendations from AAB 21.2.198 (https://twiki.cern.ch/twiki/bin/view/AtlasProtected/MCPAnalysisGuidelinesMC16#Details_on_using_the_MuonCalibra)
The `calibration mode' for the muon momentum calibration must be set explicitly by the user " via the calibrationMode property, that can be set to MuonCalibrationPeriodTool::correctData_CB (0), MuonCalibrationPeriodTool::correctData_IDMS (1) or MuonCalibrationPeriodTool::notCorrectData_IDMS (2). The current default option is "noOption", forcing all analyses are required to make an explicit choice."

I've defaulted it to set to 0 to avoid everyone's code crashing, and under the impression that most users of xAH don't use the highPt muon working point or are limited by muon systematics so should use 0, but details of which choice is best are in the twiki.

### DIFF
--- a/Root/MuonCalibrator.cxx
+++ b/Root/MuonCalibrator.cxx
@@ -188,6 +188,7 @@ EL::StatusCode MuonCalibrator :: initialize ()
     ANA_MSG_WARNING("Overriding muon 2018 saggita release to " << m_overrideSagittaRelease18);
     ANA_CHECK(m_muonCalibrationTool_handle.setProperty("SagittaRelease18", m_overrideSagittaRelease18));
   }
+  ANA_CHECK(m_muonCalibrationTool_handle.setProperty("calibrationMode", m_calibrationMode));
   ANA_CHECK(m_muonCalibrationTool_handle.retrieve());
   ANA_MSG_DEBUG("Retrieved tool: " << m_muonCalibrationTool_handle);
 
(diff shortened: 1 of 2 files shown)


jbossios:
I was about to create a PR to do exactly the same, although in my case I did two things differently:

- I didn't explicitly put a default, if Muon CP is forcing everyone to make a choice I think every analysis should make a conscious decision on the calibration mode.

- I might be wrong but my understanding from the twiki is that to recover previous functionality is not enough to set only calibrationMode, because of that I also switched to use MuonCalibrationPeriodTool instead of MuonCalibrationAndSmearingTool, this way, there is no need to manually setup things for each year/campaign. This is important because at least some of the options that you kept (which would be needed for <21.2.198) would not work on 21.2.198+.

hollypacey:
Hi @jbossios , 

I agree with the first bullet point that it's probably better to avoid a default, I was unsure about that myself but just followed that most other things were given a default. 

Reading the twiki more closely I think you are correct also about moving to the MuonCalibrationPeriodTool to ensure the right combination of settings are used. I guess if you have already updated the code and were ready to make a PR I can just cancel this one and you can do yours?

jbossios:
Sounds good to me. I will create a PR as soon as possible (I'm doing some cross-checks and should be done by today/tomorrow).

jbossios:
One remaining question is, do we want to keep master working on <21.2.198? Personally, I would create a tag for those wanting to stay on <21.2.198 and keep master for 21.2.198 such that we can simplify things and not having to support the setting of two different tools.

@kratsg any objection with that?

PR #1525, opened 2022-02-18
Update MuonCalibrator to follow MCP recommendations for AB>=21.2.198

jbossios:
MCP introduced in 21.2.198 a new property (calibrationMode) which by default is "noOption" which would make the tool crash. The same functionality is implemented here, forcing the user to pick up one of the supported ones ("correctData_CB", "correctData_IDMS" or "notCorrectData_IDMS") through m_calibrationMode.

I have also switched to using MuonCalibrationPeriodTool, instead of MuonCalibrationAndSmearingTool, which sets all properties automatically for the corresponding data year / MC campaign. All previous options were removed (some if not all of them are not supported in the PeriodTool), but if some of them are still needed (for expert usage) we could consider adding back the option to use the MuonCalibrationAndSmearingTool in a future PR (through a switch such that PeriodTool is the default).

IMPORTANT: This will make xAH::MuonCalibrator not to work on AB<21.2.198. I recommend making a tag for those wanting to stay on <21.2.198 and make 21.2.198 he minimum release for master. @kratsg: what do you think?

MCP twiki: https://twiki.cern.ch/twiki/bin/view/AtlasProtected/MCPAnalysisGuidelinesMC16

This should replace https://github.com/UCATLAS/xAODAnaHelpers/pull/1524


kratsg:
> IMPORTANT: This will make xAH::MuonCalibrator not to work on AB<21.2.198. I recommend making a tag for those wanting to stay on <21.2.198 and make 21.2.198 he minimum release for master. @kratsg: what do you think?

Let's make a tag, or at least refer to the commit. You'll also need to update the github actions workflow for the supported versions in this PR -- seems we're dropping all but 21.2.198+

kratsg:
> @kratsg I updated ci.yml to use +21.2.198. I don't seem to have permissions to create a tag.

I'll create a tag. Just give me the command I guess?

kratsg (approved):

kratsg:
https://github.com/UCATLAS/xAODAnaHelpers/releases/tag/v1.1.0 is drafted. Merging this now.

PR #1526, opened 2022-03-02
improve lepton efficiency systematics implementation

hollypacey:
- fix names of muon efficiency systematics in HelpTreeBase to match those in the MuonEfficiencyCorrector
- Make muon trigger efficiency systs reco-WP dependent in HelpTreeBase, to match MuonEfficiencyCorrector
- remove a couple of whitespaces for neatness
- add comments to the Muon/ElectronEfficiencyCorrector headers to note that the syst list needs to include "All" or "Nominal" explicitly to ensure the nominal SFs are stored in the ttree as well as the systematics varied list. 
- add an option to HelpTreeBase to specify the nominal tree name, which defaults to "nominal" for backward compatibility. Perhaps this is unnecessary but at least it draws attention to the fact it is set to nominal unless explicitly changed to avoid other people having the same problem I did with muon efficiency systs.
- make use of the m_nominalTree bool variable in HelpTreeBase::AddMuons  and HelpTreeBase::FillMuons rather than re-checking the name of the input tree and whether that matches a hard-coded string for the nominal tree name.

### DIFF
--- a/Root/HelpTreeBase.cxx
+++ b/Root/HelpTreeBase.cxx
@@ -29,7 +29,7 @@ using std::vector;
 #pragma link C++ class vector<float>+;
 #endif
 
-HelpTreeBase::HelpTreeBase(xAOD::TEvent* event, TTree* tree, TFile* file, const float units, bool debug, xAOD::TStore* store):
+HelpTreeBase::HelpTreeBase(xAOD::TEvent* event, TTree* tree, TFile* file, const float units, bool debug, xAOD::TStore* store, std::string nominalTreeName):
   m_trigInfoSwitch(nullptr),
   m_trigConfTool(nullptr),
   m_trigDecTool(nullptr),
@@ -40,7 +40,8 @@ HelpTreeBase::HelpTreeBase(xAOD::TEvent* event, TTree* tree, TFile* file, const 
   m_debug = debug;
   m_tree = tree;
   m_tree->SetDirectory( file );
-  m_nominalTree = strcmp(m_tree->GetName(), "nominal") == 0;
+  m_nominalTreeName = nominalTreeName;
+  m_nominalTree = strcmp(m_tree->GetName(), nominalTreeName.c_str()) == 0;
   m_event = event;
   m_store = store;
   Info("HelpTreeBase()", "HelpTreeBase setup");
@@ -109,8 +110,8 @@ HelpTreeBase::~HelpTreeBase() {
 }
 
 
-HelpTreeBase::HelpTreeBase(TTree* tree, TFile* file, xAOD::TEvent* event, xAOD::TStore* store, const float units, bool debug):
-  HelpTreeBase(event, tree, file, units, debug, store)
+HelpTreeBase::HelpTreeBase(TTree* tree, TFile* file, xAOD::TEvent* event, xAOD::TStore* store, const float units, bool debug, std::string nominalTreeName):
+  HelpTreeBase(event, tree, file, units, debug, store, nominalTreeName)
 {
   // use the other constructor for everything
 }
@@ -337,9 +338,7 @@ void HelpTreeBase::AddMuons(const std::string& detailStr, const std::string& muo
   xAH::MuonContainer* thisMuon = m_muons[muonName];
   HelperClasses::MuonInfoSwitch& muonInfoSwitch = thisMuon->m_infoSwitch;
 
-  std::string tname = m_tree->GetName();
-
-  if ( tname == "nominal" ) {
+  if (m_nominalTree) {
 
      if ( muonInfoSwitch.m_recoEff_sysNames && muonInfoSwitch.m_effSF && m_isMC ) {
        for (auto& reco : muonInfoSwitch.m_recoWPs) {
@@ -357,8 +356,10 @@ void HelpTreeBase::AddMuons(const std::string& detailStr, const std::string& muo
 
      if ( muonInfoSwitch.m_trigEff_sysNames && muonInfoSwitch.m_effSF && m_isMC ) {
        for (auto& trig : muonInfoSwitch.m_trigWPs) {
-         std::string trigEffSF_sysNames = "muon_TrigEff_SF_" + trig + "_sysNames";
-         m_tree->Branch( trigEffSF_sysNames.c_str() , & (m_MuonTrigEff_SF_sysNames)[ trig ] );
+         for (auto& reco : muonInfoSwitch.m_recoWPs) {
+           std::string trigEffSF_sysNames = "muon_TrigEff_SF_" + trig + "_" + reco + "_sysNames";
+           m_tree->Branch( trigEffSF_sysNames.c_str() , & (m_MuonTrigEff_SF_sysNames)[ trig ][ reco ] );
+         }
        }
      }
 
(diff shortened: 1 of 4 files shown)


Root/HelpTreeBase.cxx
```diff
@@ -378,14 +379,12 @@ void HelpTreeBase::FillMuons( const xAOD::MuonContainer* muons, const xAOD::Vert
   this->ClearMuons(muonName);
   HelperClasses::MuonInfoSwitch& muonInfoSwitch = m_muons[muonName]->m_infoSwitch;
 
-  std::string tname = m_tree->GetName();
-
-  if ( tname == "nominal" ) {
+  if (m_nominalTree) {
```
jbossios, line 382: Same as above

Root/HelpTreeBase.cxx
```diff
@@ -437,28 +438,29 @@ void HelpTreeBase::FillMuon( const xAOD::Muon* muon, const xAOD::Vertex* primary
 
 void HelpTreeBase::ClearMuons(const std::string& muonName) {
 
-  std::string tname = m_tree->GetName();
   xAH::MuonContainer* thisMuon = m_muons[muonName];
   HelperClasses::MuonInfoSwitch& muonInfoSwitch = thisMuon->m_infoSwitch;
 
-  if ( tname == "nominal" ) {
+  if (m_nominalTree) {
```
jbossios, line 444: same as above

Root/HelpTreeBase.cxx
```diff
@@ -337,9 +338,7 @@ void HelpTreeBase::AddMuons(const std::string& detailStr, const std::string& muo
   xAH::MuonContainer* thisMuon = m_muons[muonName];
   HelperClasses::MuonInfoSwitch& muonInfoSwitch = thisMuon->m_infoSwitch;
 
-  std::string tname = m_tree->GetName();
-
-  if ( tname == "nominal" ) {
+  if (m_nominalTree) {
```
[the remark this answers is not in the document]
hollypacey: maybe I misunderstood, isn't m_nominalTree a bool (originally set via m_nominalTree = strcmp(m_tree->GetName(), "nominal") == 0)  so shouldn't this be false if we don't have "nominal"?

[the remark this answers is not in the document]
hollypacey: just tested it locally and the syst SFs are only added to my nominal tree still

Root/HelpTreeBase.cxx
```diff
@@ -394,17 +393,19 @@ void HelpTreeBase::FillMuons( const xAOD::MuonContainer* muons, const xAOD::Vert
...
     }
 
     if ( muonInfoSwitch.m_trigEff_sysNames && muonInfoSwitch.m_effSF && m_isMC ) {
       for ( auto& trig : muonInfoSwitch.m_trigWPs ) {
-        std::vector< std::string >* tmp_trig_sys(nullptr);
-        if ( m_store->retrieve(tmp_trig_sys, "MuonEfficiencyCorrector_TrigSyst_" + trig).isSuccess() ) {
-          (m_MuonTrigEff_SF_sysNames)[ trig ] = *tmp_trig_sys;
+        for ( auto& reco : muonInfoSwitch.m_recoWPs ) {
+          std::vector< std::string >* tmp_trig_sys(nullptr);
+          if ( m_store->retrieve(tmp_trig_sys, "MuonEfficiencyCorrector_TrigSyst_" + trig + "_Reco" + reco).isSuccess() ) {
```
jbossios, line 406: I think this will break things for some people (like me), since the name doesn't always contains '_Reco' + reco. That only happens here:
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/MuonEfficiencyCorrector.cxx#L264

The same happens the two other modifications above.

So, I would fix this rare case w/o modifying the standard user case.

Root/HelpTreeBase.cxx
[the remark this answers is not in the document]
jbossios: Sorry, you are right, I thought it was the name of the TTree and not a boolean, I see now this:

https://github.com/UCATLAS/xAODAnaHelpers/blob/ea29ad9b0bb26137cf5d24f99f9b6e67d171231e/Root/HelpTreeBase.cxx#L44

resolving this...

Root/HelpTreeBase.cxx
jbossios: Exactly, I would check the same here to still support both cases.

Root/HelpTreeBase.cxx
```diff
@@ -357,8 +356,10 @@ void HelpTreeBase::AddMuons(const std::string& detailStr, const std::string& muo
 
      if ( muonInfoSwitch.m_trigEff_sysNames && muonInfoSwitch.m_effSF && m_isMC ) {
        for (auto& trig : muonInfoSwitch.m_trigWPs) {
-         std::string trigEffSF_sysNames = "muon_TrigEff_SF_" + trig + "_sysNames";
-         m_tree->Branch( trigEffSF_sysNames.c_str() , & (m_MuonTrigEff_SF_sysNames)[ trig ] );
+         for (auto& reco : muonInfoSwitch.m_recoWPs) {
+           std::string trigEffSF_sysNames = "muon_TrigEff_SF_" + trig + "_" + reco + "_sysNames";
+           m_tree->Branch( trigEffSF_sysNames.c_str() , & (m_MuonTrigEff_SF_sysNames)[ trig ][ reco ] );
```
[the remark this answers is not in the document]
hollypacey: I use a single reco WP and the other trigger Eff and SF branches them still get the reco WP appended to them so I thought this would be more consistent but you're right we should avoid breaking backward compatibility so I can change it.

Root/HelpTreeBase.cxx
jbossios: This is super confusing, but in my case, I don't need to modify any of the following:

m_outputSystNamesReco
m_outputSystNamesIso
m_outputSystNamesTrig
m_outputSystNamesTTVA

Which means, this line is executed:
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/MuonEfficiencyCorrector.cxx#L161

But this one not:
https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/MuonEfficiencyCorrector.cxx#L264

hollypacey: Okay I'm not sure how what I did locally is different to you but mine crashed, just focusing on the Reco for now..

If 161 is executed and you keep the default output name then you should get "MuonEfficiencyCorrector_RecoSyst_Reco"+yourworkingpoint after line 161

but without my change https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/HelpTreeBase.cxx#L388 is looking for "MuonEfficiencyCorrector_RecoSyst_"+yourworkingpoint
so does it work if you make sure your code calls that line?

jbossios: OK I have been investigating this. You are right, this doesn't work as it is but I'm not using it and likely many people are not using either (and hence it was never identified). What I do is to rely on MuonContainer and the safeSFVecFill function which fill nominal and systematic variations as well and there the names are correct, as you can see here:

https://github.com/UCATLAS/xAODAnaHelpers/blob/master/Root/MuonContainer.cxx#L957

Sorry for the confusion. So, in summary, the append of _Reco here is needed and doesn't affect the current (i.e. most common) workflow. Resolving.

hollypacey: ah interesting, I didn't spot there was alternative way to Fill the muons. Thank you for investigating!

PR #1522, opened 2022-02-07
Make Taus work in R22

HenryDayHall:
Solves https://github.com/UCATLAS/xAODAnaHelpers/issues/1521

Also, N.B. property EfficiencyCorrectionTypes (see the table on https://gitlab.cern.ch/atlas/athena/-/blob/release/22.2.60/PhysicsAnalysis/TauID/TauAnalysisTools/doc/README-TauEfficiencyCorrectionsTool.rst) has new values I haven't included because they are not obviously related to previous ones and I don't know what this code actually does.

PR #1527, opened 2022-03-04
QuickFix tau edits for R22

salderwe:
Quick edits to make this compatible with AB,22.2.63, mostly BDT-->RNN WPs.
Should clean up further, and add possible new WPs.


jbossios:
There seems to be conflicts, please resolve them. Thanks for the fix!

jbossios:
Closing it since it is no longer needed since fixes were merged in https://github.com/UCATLAS/xAODAnaHelpers/pull/1522

PR #1531, opened 2022-04-14
adding option to change MET working point

hollypacey:
As described here https://twiki.cern.ch/twiki/bin/view/AtlasProtected/EtmissRecommendationsFullRun2#Working_Points, the MET Run2 recommendations support several MET Working points that each place different selections on the Jets passed into METMaker. 

By default, the code and recommendation uses the Tight working point, but we should be able to alter this line https://gitlab.cern.ch/atlas/athena/-/blob/21.2/Reconstruction/MET/METUtilities/Root/METMaker.cxx#L111 when initiating the METMaker too so analyses (and mainly the METPerformance code) can change working points. 

The code change is just to add the option to set this property in METConstructor, with a variable called m_METWorkingPoint (I think this is a bit clearer than e.g. m_JetSelection to users).

### DIFF
--- a/Root/METConstructor.cxx
+++ b/Root/METConstructor.cxx
@@ -124,7 +124,9 @@ EL::StatusCode METConstructor :: initialize ()
   if ( m_doMuonPFlowBugfix ) {
     ANA_CHECK(m_metmaker_handle.setProperty("DoMuonPFlowBugfix", true));
   }
-  ANA_CHECK(m_metmaker_handle.setProperty("JetSelection", m_METWorkingPoint));
+  if ( !m_METWorkingPoint.empty() ){
+    ANA_CHECK(m_metmaker_handle.setProperty("JetSelection", m_METWorkingPoint));
+  }
   ANA_CHECK(m_metmaker_handle.retrieve());
   ANA_MSG_DEBUG("Retrieved tool: " << m_metmaker_handle);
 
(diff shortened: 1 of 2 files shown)


jbossios:
I'm wondering if we want to always set it to Tight or if it would be better to always rely on the tool's default... I would say we probably should do the latter (i.e. make it blank by default and set it only if string is not empty). What do you think?

hollypacey:
Hi, I think for R22 that's a good idea in case we change the default to something other than Tight. For now you could argue that it's nice to more easily see that you are using Tight, but it does print in the terminal anyway so okay have changed it. 

(one day I am determined to make a pull request that I don't need to edit again!)

kratsg:
> (one day I am determined to make a pull request that I don't need to edit again!)

everyone strives for that perfection ./cc @matthewfeickert

kratsg (approved):

hollypacey:
@jbossios no worries, I am a pretty beginner coder so it's all helping me learn good practice :)

matthewfeickert:
> > (one day I am determined to make a pull request that I don't need to edit again!)
> 
> everyone strives for that perfection

:stuck_out_tongue: Code review is a collaborative process and revision is a natural part of the discussion. I probably end up revising most of my PRs at least once before a reviewer ever gets to them! @hollypacey looks like you're doing fine. :+1:

PR #1540, opened 2022-05-19
Clean up btaggers and add DL1dv00

prbbing:
Hello,

I tried to clean up the outdated b-taggers for the rel22 version: all MV2 tagger and DL1, DL1rnn, DL1rmu are removed

Also, the f_c values are updated to the latest value 0.018. In the future we may just want to retrieve this value from the CDI directly. I will follow this up.

New DL1dv00 tagger is added, which is the baseline rel22 btagger. This would allow the users to test it.

I am not sure whether in rell22, there will still be users considering btagging at the HLT though. If so, we may want to bring those MV2 taggers at the HLT back. They are currently removed as well.

I tested it on one rel22 PHYS sample for both DL1r and DL1dv00 taggers and everything ran fine with reasonable outputs.

Suggestions are welcome!

Thanks for your time!

Bing

### DIFF
--- a/xAODAnaHelpers/HelperClasses.h
+++ b/xAODAnaHelpers/HelperClasses.h
@@ -479,10 +479,6 @@ namespace HelperClasses {
 
             ``jetBTag`` expects the format ``jetBTag_tagger_type_AABB..MM..YY.ZZ``. This will create a vector of working points (AA, BB, CC, ..., ZZ) associated with that tagger. Several entries can be given. For example::
 
-                m_configStr = "... jetBTag_MV2c10_FixedCutBEff_60707785 ..."
-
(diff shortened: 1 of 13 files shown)


xAODAnaHelpers/HelperClasses.h
```diff
@@ -479,10 +479,6 @@ namespace HelperClasses {
 
             ``jetBTag`` expects the format ``jetBTag_tagger_type_AABB..MM..YY.ZZ``. This will create a vector of working points (AA, BB, CC, ..., ZZ) associated with that tagger. Several entries can be given. For example::
 
-                m_configStr = "... jetBTag_MV2c10_FixedCutBEff_60707785 ..."
```
jbossios, line 482: I guess this is still useful but we just need to change MV2c10 by any of the R22 taggers, right?

PR #1544, opened 2022-05-31
Adding first implementation of access to Phase1 L1Calo jet containers

rbarrue:
Adding first implementation of access to Phase1 L1Calo jet containers. Separated functions for Legacy and Phase1 L1Calo containers, since the name of the ET variable is different.

Templated the function for Phase1 L1Calo jets, since the access to et, eta and phi is done the same way.

Harmonised the "ET" variable name at the output, mentioning that the variable names are different in the base EDMs.

Any further suggestions to clean up the code are welcome.

Tagging @jbossios.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
         release_type:
           - analysisbase
         release_version:
-          - 22.2.59
+          - 22.2.74
 
     steps:
     - uses: actions/checkout@master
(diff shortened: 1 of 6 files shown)


rbarrue:
CI is failing given that the EDM for the Phase1 L1 jet containers in the AnalysisBase version used for CI (22.2.59) does not contain the necessary members.

It worked locally for 22.2.74. 

@kratsg how would one go about updating the AnalysisBase version used in the CI ?

PR #1547, opened 2022-06-29
add option to copy b-tagging links to JetCalibrator

miholzbo:
This PR adds the possibility to copy b-tagging links to the "default" jet collection in the JetCalibrator (as this is the algorithm that should run first on the input jet collection). This allows to allow run the b-tagging tools also on these jet collections and one doesn't have to set up a new shallow copy of jets for the METMaker.


jbossios:
I minor suggestion to make m_btaggingContainerName emtpy by default such that the user has to decide which timestamp to use and also such that this never gets out of date (potentially w/o the user knowing).

miholzbo:
Good idea! I've pushed another commit that enforces the user to actually specify the container name from which the b-tagging links are copied from.

jbossios:
Perfect! Thanks for implementing this nice feature and the minor suggested change!

kratsg (approved):

PR #1550, opened 2022-07-11
Proper initialization of flag to include or not cleaning LLP information

rbarrue:
Ntuple production from AODs was including the jet cleaning information when not asking for it. This was leading to having files with different numbers of branches which was leading to merging problems on the Grid (smaller number of events than in the base sample).

This MR aims to fix that.


kratsg (approved):

kratsg:
whoops sorry @jbossios as I just noticed this is for R22. But it seems somewhat harmless...

PR #1556, opened 2022-07-29
Update FAQs.rst

shefmarkh:
The example grid submission command did not work for me. No dataset could be found..Adding the option "--inputRucio" solved that issue, so presumably the documentation should be updated to give a command that works.

Cheers,

Mark


kratsg:
Note: the example works fine if you had a local file / directory of that name. But we'll change it for consistency.

PR #1557, opened 2022-08-16
Change LLP stream and fix trigger matching for merged leptons

christian-appelt:
Hello,

The LLP DAOD stream is now called `StreamDAOD_LLP1`.

We use `CP::ElectronLRTMergingAlg` and `CP::MuonLRTMergingAlg` to merge LRT and std electrons and muons on analysis level, more details about the muon merger are here: [link](https://indico.cern.ch/event/1188523/contributions/4994850/attachments/2490167/4276213/AMG_5Aug2022.pdf).
In the merged container, the trigger matching does not work anymore bc of broken links. I've updated the links in the electron (and muon selector) and added a flag `m_merged_electrons` (`m_merged_muons`) to use the updated links whenever merged leptons are used. We also needed to be able to configure the Input Prefix property of the trigger decision tool. I've added `m_trigInputPrefix = "TrigMatch_"` in both electron and muon selector to do this (the value we need is `"LRTTrigMatch_"`).

Pls let me know if you have any comments and if this can be merged in r22/master.

Cheers,
Christian

### DIFF
--- a/Root/Algorithm.cxx
+++ b/Root/Algorithm.cxx
@@ -127,7 +127,7 @@ bool xAH::Algorithm::isPHYS(){
     TTree* metaData = dynamic_cast<TTree*>( wk()->inputFile()->Get("MetaData") );
     if(metaData){
       metaData->LoadTree(0);
-      return (metaData->GetBranch("StreamDAOD_PHYS") || metaData->GetBranch("StreamDAOD_PHYSLLP"));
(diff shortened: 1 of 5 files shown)


jbossios:
Hi,

If it is not too much to ask, in ElectronSelector and MuonSelector, I would make m_trigInputPrefix empty by default and set InputPrefix only when m_trigInputPrefix is not empy. That way, we are sure we always use the tool's default value and don't need to update it if at any point changes.

christian-appelt:
Hi @jbossios,

That's a very good point, thx. I've changed it.

PR #1564, opened 2022-09-06
cherry-pick commits from master branch

jbossios:
Screenshot from 2022-09-06 11-18-54

cherry-picked all commits from the master branch, except for the following ones because they would not work yet in r22 or because they might not be needed:

https://github.com/UCATLAS/xAODAnaHelpers/pull/1560
https://github.com/UCATLAS/xAODAnaHelpers/pull/1499
https://github.com/UCATLAS/xAODAnaHelpers/pull/1488
https://github.com/UCATLAS/xAODAnaHelpers/pull/1504
https://github.com/UCATLAS/xAODAnaHelpers/pull/1547

I also updated the list of supported releases which goes from 22.2.74 to 22.2.87


mattleblanc:
I think that #1560 should go in, is there a reason you think it won't work?

jbossios:
@mattleblanc We should eventually add this option that we will need but that option is currently missing in JetCalibTools in R22, so I prefer not adding something that will not work and we can add it once this is actually supported in R22.

mattleblanc:
Ah indeed! Good point, we should put it in athena first. :)

hollypacey:
Hi, Thanks for doing this! what was wrong with @miholzbo's #1547?

jbossios:
> Hi, Thanks for doing this! what was wrong with @miholzbo's #1547?

'class xAOD::Jet_v1' has no member named 'setBTaggingLink'

miholzbo:
Hi @hollypacey , @jbossios,

I've just learnt from the SUSYTools maintainers that copying the b-tagging links isn't needed anymore in R22 - this is already done during the derivation step (which is arguably also the much more intuitive thing). Hence #1547 isn't needed at all and we can just pass the normal AntiKt4EMPflowJets to the b-tagging tool. I've just tested this and it seems indeed to work :)

jbossios:
Yes, I should have commented that as well. I tested it too and it works w/o problems.

hollypacey:
Hi @jbossios @miholzbo , 

okay, thanks both!

PR #1567, opened 2022-09-09
Add isRun3Geo option to MuonSelector and MuonCalibrator

hollypacey:
Adds new member variable m_isRun3Geo to MuonSelector.cxx which sets the IsRun3Geo property in muonSelectionTool - required to be set to true to avoid a crash when running on Run3.

Defaults to false for Run2 setting.


jbossios:
Hi @hollypacey 

Thank you for implementing this. I think you need to be consistent and set to true the same property in MuonCalibTool (i.e. MuonCalibrator) [*]. Do you mind adding the same into MuonCalibrator?

Best,
Jona

[*] https://twiki.cern.ch/twiki/bin/view/AtlasProtected/MCPAnalysisGuidelinesR22#CP_MuonCalibTool_tool

PR #1569, opened 2022-09-20
autoconfigure PRW: add mc20 support, remove mc16 support

hollypacey:
- Just replaced mc16a/d/e/ with mc20a/d/e for the pileup reweighting autoconfiguration
- no user config changes needed, the actualMu and lumicalc files approach is unchanged.
- Also removed mc16c/mc16f as I don't think R22 will have these.

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -1124,14 +1124,13 @@ StatusCode BasicEventSelection::autoconfigurePileupRWTool()
   switch(runNum)
     {
     case 284500 :
-      mcCampaignMD="mc16a";
+      mcCampaignMD="mc20a";
       break;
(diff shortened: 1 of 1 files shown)


kratsg:
Silly question but do we want something that checks if someone passes in mc16 and warns or crashes? Or this is enough and nobody will make the mistake?

jbossios:
> Silly question but do we want something that checks if someone passes in mc16 and warns or crashes? Or this is enough and nobody will make the mistake?

I would say this check is enough:
https://github.com/UCATLAS/xAODAnaHelpers/blob/r22/master/Root/BasicEventSelection.cxx#L1167

Or am I missing something?

kratsg:
That works (if they use BES)!

PR #1573, opened 2022-10-22
Bugfix: LHCut not updated if multiple instances of ElectronSelector are called

jburzy01:
Fixes a bug in the electron selector which affects the LH decision cut if multiple instances of the selector are called with different WPs due to a static ConstAccessor.

For e.g. if you have

Presel -> OverlapRemoval -> FinalSel

with different LH WPs in Presel and FinalSel, the WP in Presel will be used for FinalSel. This should now be fixed.


jburzy01:
looks like master is affected too -- is sweeping implemented?

kratsg:
It should work now.

PR #1579, opened 2022-10-25
fix: Use PR target instead

kratsg:
Actions triggered by forked repositories only have read-only permissions (even for actions triggered by post-PR merging).

See `pull_request_target` information here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests


jbossios:
Is this still needed? @kratsg

PR #1581, opened 2022-11-23
Add support to electron LooseBL PID

jbossios:
News:
- Electron LooseBL ID working point is now saved in derivations (no need to use Loose instead)
- The only electron loose ID working point for which SFs are provided is LooseBLayer [1]
  - AFAICT: LooseBLayer corresponds to the LooseBL working point

NOTE: This MR breaks backward compatibility when:

- m_readIDFlagsFromDerivation and m_doLHPID and m_doLHPIDcut are all true in ElectronSelector: Loose -> LooseBL
- m_doLHPIDcut is false in ElectronSelector: new lowest working point: Loose -> LooseBL

I don't see the point of maintaining the current functionality since Loose do not have SFs, but I'm open for suggestions.

[1] https://atlas-groupdata.web.cern.ch/atlas-groupdata/ElectronEfficiencyCorrection/2015_2018/rel21.2/Precision_Summer2020_v1/map4.txt

### DIFF
--- a/Root/ElectronEfficiencyCorrector.cxx
+++ b/Root/ElectronEfficiencyCorrector.cxx
@@ -127,6 +127,10 @@ EL::StatusCode ElectronEfficiencyCorrector :: initialize ()
   //
   if ( !m_WorkingPointPID.empty() ) {
 
+    // Loose working point saved as DFCommonElectronsLHLooseBL in derivations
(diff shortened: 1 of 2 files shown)


jbossios:
@kratsg friendly reminder about this.

kratsg:
Sorry, this slipped through my radar, also because I'm on strike.

kratsg (approved):

PR #1583, opened 2022-11-30
additional options for TProfiles in HistogramManager

tofitsch:
adding additional methods for TProfiles in HistogramManager similar to those already available for TH1F.

### DIFF
--- a/Root/HistogramManager.cxx
+++ b/Root/HistogramManager.cxx
@@ -242,9 +242,3 @@ void HistogramManager::fillHist(const std::string& histName, double valueX, doub
   }
   histPointer->Fill(valueX, valueY, weight);
 }
-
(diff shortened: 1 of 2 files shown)


Root/HistogramManager.cxx
```diff
@@ -222,3 +242,9 @@ void HistogramManager::fillHist(const std::string& histName, double valueX, doub
   }
   histPointer->Fill(valueX, valueY, weight);
 }
+
+// void HistogramManager::fillProfile(const std::string& histName, double value, double bin) {
+//   TH1* profilePointer(NULL);
+//   histPointer = this->findHist(histName);
+//   histPointer->Fill(value, bin)
+// }
```
jbossios, line 250: Could you please remove these commented lines?

tofitsch: Oops. Thanks for pointing that out. Removed them now.

PR #1584, opened 2022-11-30
add timing cuts to JetSelector

tofitsch:
Add timing cuts to JetSelector tested in release 22.2.97. Modifications include support for cutflow histograms and adding an option to set an abs(timing) > X selection for the input jet collection. Returns from JetSelector::PassCuts when no "Timing" decoration is included on the jet.

### DIFF
--- a/Root/JetSelector.cxx
+++ b/Root/JetSelector.cxx
@@ -1304,9 +1304,6 @@ int JetSelector :: PassCuts( const xAOD::Jet* jet ) {
   }
   if ( m_useCutFlow ) m_jet_cutflowHist_1->Fill( m_jet_cutflow_jvt_cut, 1 );
 
-  //
-  //  Timing (same for offline and HLT for the time being)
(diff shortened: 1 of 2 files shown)


Root/JetSelector.cxx
```diff
@@ -1303,6 +1304,21 @@ int JetSelector :: PassCuts( const xAOD::Jet* jet ) {
   }
   if ( m_useCutFlow ) m_jet_cutflowHist_1->Fill( m_jet_cutflow_jvt_cut, 1 );
 
+  //
+  //  Timing (same for offline and HLT for the time being)
```
jbossios, line 1308: This comment is not correct/necessary, right? One could set two JetSelectors, one for offline and one for HLT and have a different timing cut on each.

tofitsch: Yes, that's right. Thanks. Removed that comment now.

PR #1586, opened 2022-12-07
Adding LRT muons information

gfrattar:
Hi, 

for dHNL analysis we would like to slightly modify the code such that in the case of LRT muons no ID cuts are checked. Additionally, the boolean used in this step (checking if the muon is LRT and if it passes ID cuts) are stored in the output ntuples.

In attachment one useful plot that shows that for LRT muons you can have both cases in which ID cuts are passed or not, while for prompt muons these are always satisfied since the ID cuts is intrinsic in the accept method.

[c1.pdf](https://github.com/UCATLAS/xAODAnaHelpers/files/10175616/c1.pdf)

Thank you, cheers.
Guglielmo & Sagar

### DIFF
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -250,8 +250,6 @@ namespace HelperClasses{
     m_passSel = has_exact("passSel");
     // passOR
     m_passOR = has_exact("passOR");
-
-    m_doLRT = has_exact("doLRTelectrons");
   }
 
   void PhotonInfoSwitch::initialize(){
@@ -289,6 +287,7 @@ namespace HelperClasses{
     m_truth         = has_exact("truth");
     m_truthDetails  = has_exact("truth_details");
     m_layer         = has_exact("layer");
+    m_fJVT          = has_exact("fJVT");
     m_trackPV       = has_exact("trackPV");
     m_trackAll      = has_exact("trackAll");
     m_chargedPFOPV  = has_exact("chargedPFOPV");
--- a/Root/HelperClasses.cxx
+++ b/Root/HelperClasses.cxx
@@ -201,6 +201,8 @@ namespace HelperClasses{
     m_trigEff_sysNames = has_exact("trigEff_sysNames");
     m_ttvaEff_sysNames = has_exact("ttvaEff_sysNames");
 
+    m_doLRT = has_exact("doLRTmuons");
+
   }
 
   void ElectronInfoSwitch::initialize(){
@@ -248,6 +250,8 @@ namespace HelperClasses{
     m_passSel = has_exact("passSel");
     // passOR
     m_passOR = has_exact("passOR");
+
+    m_doLRT = has_exact("doLRTelectrons");
   }
 
   void PhotonInfoSwitch::initialize(){
@@ -289,7 +293,6 @@ namespace HelperClasses{
     m_trackAll      = has_exact("trackAll");
     m_chargedPFOPV  = has_exact("chargedPFOPV");
     m_jvt           = has_exact("JVT");
-    m_fJvt          = has_exact("fJvt");
     m_NNJvt         = has_exact("NNJvt");
     m_allTrack      = has_exact("allTrack");
     m_allTrackPVSel = has_exact("allTrackPVSel");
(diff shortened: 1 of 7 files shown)


Root/MuonContainer.cxx
```diff
@@ -130,6 +130,9 @@ MuonContainer::MuonContainer(const std::string& name, const std::string& detailS
     m_PromptLeptonVeto                  = new std::vector<float> ();
   }
 
+  m_isLRT = new std::vector<char>();
```
jbossios, line 133: You need to add a new flag to the infoSwitch. Otherwise, everyone will get these branches (making unnecessarily larger TTrees for others).

Root/MuonSelector.cxx
```diff
@@ -780,9 +782,16 @@ int MuonSelector :: passCuts( const xAOD::Muon* muon, const xAOD::Vertex *primar
...
+
   ANA_MSG_DEBUG( "Doing muon quality" );
   // this will accept the muon based on the settings at initialization : eta, ID track info, muon quality
-  if ( ! m_muonSelectionTool_handle->accept( *muon ) ) {
+  // if muon is LRT do not check ID cuts
+  bool LRTdecorIsAvailable = isLRTmuon.isAvailable(*muon);
+  bool muonIsLRT = isLRTmuon(*muon) > 0.5;
+  bool AcceptPromptMuon = (bool)m_muonSelectionTool_handle->accept( *muon );
+
+  if ( (!LRTdecorIsAvailable && !AcceptPromptMuon) || (LRTdecorIsAvailable && !(muonIsLRT || AcceptPromptMuon)) ){
```
jbossios, line 794: I would prefer not to add more decorations to the muons for everyone, even if those are then not stored in the TTrees, since it increases the memory footprint for every single analysis. So, I would recommend to add all that is needed for this analysis under a flag, so the exact same behaviour is kept for any other analysis.

Also, if !muonIsLRT is true, then the if statement should be accessed only if AcceptPromptMuon is false. Which is not the case if the LRT decoration is available (it would enter for any value of AcceptPromptMuon). It would be more robust to keep the current behaviour on any case, i.e. regardless if that decoration is available or not, and only change the behaviour when requested by the user. This decoration might be added by an specific derivation, but if in the future such derivation is used for more analyses or the daod format is merged with another format making again that decoration available for others, then we risk to change the expected behaviour for those. So, I would separate this from normal usage and add an option to enable this check.

jbossios: I would add two options:
1) Turns ON the decoration of the muon with the passIDcuts (off by default)
2) Add dedicated treatment for LRT muons (disabled by default)

gfrattar: hi, ok we fixed the code as you suggested and added the flags. The logic of the if of line 794 here has been made more clear.

Root/HelperClasses.cxx
```diff
@@ -289,7 +293,6 @@ namespace HelperClasses{
     m_trackAll      = has_exact("trackAll");
     m_chargedPFOPV  = has_exact("chargedPFOPV");
     m_jvt           = has_exact("JVT");
-    m_fJvt          = has_exact("fJvt");
```
jbossios, line 292: Please add this line back.

Root/HelperClasses.cxx
```diff
@@ -248,6 +250,8 @@ namespace HelperClasses{
     m_passSel = has_exact("passSel");
     // passOR
     m_passOR = has_exact("passOR");
+
+    m_doLRT = has_exact("doLRTelectrons");
```
jbossios, line 254: This is not used anywhere, so I would remove it.

Root/MuonContainer.cxx
jbossios: Thanks

xAODAnaHelpers/HelperClasses.h
```diff
@@ -370,6 +372,7 @@ namespace HelperClasses {
     std::vector< std::string > m_trigWPs;
     bool m_passSel;
     bool m_passOR;
+    bool m_doLRT;
```
jbossios, line 375: Same, there, this is not used.

Root/MuonSelector.cxx
jbossios: Thanks

gfrattar:
hi @jbossios , all required modifications should be done

jbossios:
Hi @gfrattar ,

Thanks for implementing my suggestions. I will merge it once the CI tests are done.

BTW, for next PRs, please do not add spaces to empty lines (one like this was added) and do not change the order of the lines (m_fJVT was added back to a different line).

Best,
Jona

PR #1590, opened 2022-12-21
do not try to access btagging SFs for jets with eta > 2.5

gfrattar:
Hi,

this MR introduces a check on the eta of the jets when trying to access the b-tagging information, limiting to abs(eta) < 2.5 as per detector constraints. This is needed because of the recent modification to Athena, in which an error message is spammed if this action is tried, see https://gitlab.cern.ch/atlas/athena/-/blob/release/22.2.99/PhysicsAnalysis/JetTagging/JetTagPerformanceCalibration/CalibrationDataInterface/Root/CalibrationDataInterfaceROOT.cxx#L805-808 .

Cheers,
Guglielmo

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -408,6 +408,8 @@ EL::StatusCode BJetEfficiencyCorrector :: executeEfficiencyCorrection(const xAOD
   //
   for( const xAOD::Jet* jet_itr : *(inJets)){
 
+    if(abs(jet_itr->eta()) > 2.5) continue;
+
     if(!m_useContinuous){
       // get tagging decision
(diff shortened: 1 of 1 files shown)


hollypacey:
Hi @gfrattar and @jbossios,

This change means that the decision decorator isn't set for all jets in the collection which seems non-standard and means you have to add the eta check again if you try to access the corresponding accessor. Can you alter line 411 so that if the jet has |eta|>2.5 you set the btagging decisions to 0 rather than just continuing?

I think that would also be more backwards compatible with people's analysis code.

jbossios:
Hi @hollypacey ,

You are right! Nice catch! @gfrattar would you be able to take care of the fix?

Best,
Jona

PR #1598, opened 2023-01-17
Fixed btag decorator

gfrattar:
Hi, 

this is to fix the decorators as requested in 

https://github.com/UCATLAS/xAODAnaHelpers/pull/1590#issuecomment-1384653282

Cheers,
Guglielmo

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -408,6 +408,8 @@ EL::StatusCode BJetEfficiencyCorrector :: executeEfficiencyCorrection(const xAOD
   //
   for( const xAOD::Jet* jet_itr : *(inJets)){
 
+    if(abs(jet_itr->eta()) > 2.5) continue;
+
     if(!m_useContinuous){
       // get tagging decision
       ANA_MSG_DEBUG(" Getting tagging decision ");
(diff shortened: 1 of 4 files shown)


Root/BJetEfficiencyCorrector.cxx
```diff
@@ -408,7 +408,12 @@ EL::StatusCode BJetEfficiencyCorrector :: executeEfficiencyCorrection(const xAOD
   //
   for( const xAOD::Jet* jet_itr : *(inJets)){
 
-    if(abs(jet_itr->eta()) > 2.5) continue;
+    if(abs(jet_itr->eta()) > 2.5){
+      dec_isBTag( *jet_itr ) = 0;
+      dec_isBTagOR( *jet_itr ) = 0;
+      dec_Weight( *jet_itr) = 1;
```
jbossios, line 414: Hi @gfrattar,

Thanks for taking care of this.

Jets should be decorated with ```dec_isBTag``` and ```dec_isBTagOR``` only if this is satisfied:

```if(!m_useContinuous)```

while, jets should be decorated with ```dec_Quantile``` otherwise.

If the following is satisfied: ```if(m_useContinuous || m_alwaysGetTagWeight)```, then ```dec_Weight``` should also be set.

jbossios: A similar fix is also needed in L481, where ```dec_sfBTag``` and ```dec_ineffsfBTag``` needs to be set for jets outside the tracker.

gfrattar:
hi @jbossios ,

thanks for the suggestions, please let me know if now everything is fine.

Cheers,
Guglielmo

hollypacey:
Thanks for making this change! :)

PR #1599, opened 2023-01-17
Updating Tau calib/selection for R22 pre-recs

hollypacey:
Following https://twiki.cern.ch/twiki/bin/view/AtlasProtected/TauRecommendationsR22
- Update configurable arguments for TauSmearingTool in the TauCalibration, with updated member variables that can be set in config file and defaults to match twiki for "2022-prerec".
- Update TauSelection and TauContainers to replace BDT mentions with RNN, since BDT is obsolete.
Tested still in release 22.2.97

### DIFF
--- a/xAODAnaHelpers/TauCalibrator.h
+++ b/xAODAnaHelpers/TauCalibrator.h
@@ -20,9 +20,14 @@ public:
 
   std::string m_RecommendationTag = "";
   
-  bool m_applyMVATES = false;
-  bool m_applyCombinedTES = false;
+  //  apply a compatibility check between calo TES and MVA TES. For taus that do not pass the test calo based TES is used. Not recommended for 2022-prerec tag.
+  bool m_applyMVATESQualityCheck = false;
+  // For 2022-prerec, toggle between smearing/uncertainties calculated from PowhegPythia or Sherpa
+  std::string m_generator = "PoPy";
+  std::string m_campaign = "mc20";
   bool m_setAFII = false;
+
+  bool m_skipTruthMatchCheck = false;
 
   // sort after calibration
   bool    m_sort = true;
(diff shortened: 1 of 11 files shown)


xAODAnaHelpers/TauCalibrator.h
```diff
@@ -20,10 +20,15 @@ class TauCalibrator : public xAH::Algorithm
 
   std::string m_RecommendationTag = "";
   
-  bool m_applyMVATES = false;
-  bool m_applyCombinedTES = false;
+  //  apply a compatibility check between calo TES and MVA TES. For taus that do not pass the test calo based TES is used. Not recommended for 2022-prerec tag.
+  bool m_applyMVATESQualityCheck = false;
+  // For 2022-prerec, toggle between smearing/uncertainties calculated from PowhegPythia or Sherpa
+  std::string m_generator = "PoPy";
+  std::string m_campaign = "mc20";
```
jbossios, line 27: I have a few comments about this:

1. If we are going to rely on defaults, we should set coherent values for m_generator and m_campaign. "PoPy" goes together with "mc21" and not "mc20".

2. We could have synchronized default values for m_generator and m_campaign (i.e. PoPy and mc21), but personally, I would like to rely on the tool's defaults and set these properties only when the user specifically asked to do it. That way, if defaults are updated, we don't have to update this file. So, I propose settings these strings to empty and set those properties only when those strings are not empty.

jbossios (changes requested):

xAODAnaHelpers/TauCalibrator.h
hollypacey: Hi, 
Good points. I have followed your (2) suggestion to only set the strings if non-empty.

jbossios: thanks!

jbossios (approved):

PR #1605, opened 2023-02-08
Fix default MCType for FullSim in JetCalibrator

jbossios:
JetUncertainties has now two options in R22: MC20 or MC21. Hence, we shouldn't be using MC16 as the default FullSim MCType. Instead, I propose to force users to choose the MCType when using FullSim samples. The default is kept empty and it would have to be set by the user to MC20 or MC21.

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -158,7 +158,10 @@ EL::StatusCode JetCalibrator :: initialize ()
     }
   }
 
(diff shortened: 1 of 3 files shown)


mattleblanc:
Hi, just commenting to say that we (Jet/EtMiss) discussed this with Jona and think this is the best default option for the tool now in order to prevent problems from occuring. Thanks!

Issue #1572, opened 2022-09-30
Support Run 3 style navigation


jbossios:
The following line needs to be customizable:

https://github.com/UCATLAS/xAODAnaHelpers/blob/r22/master/Root/TrigMatcher.cxx#L103

We currently have ```Trig::MatchingTool``` which is for Run 2 style navigation but for Run 3 style navigation, we need ```Trig::R3MatchingTool``` instead.

More details here:
https://twiki.cern.ch/twiki/bin/viewauth/Atlas/R22TriggerAnalysis

jbossios:
Closing issue since this was implemented in #1606

PR #1606, opened 2023-02-08
Add isRun3 flag and configure TDT properly

longjon929:
Add a flag to control Run 3 vs Run 2 behavior and use it to configure the trigger decision tool properly for Run 3.  Addresses part of #1572

### DIFF
--- a/Root/Algorithm.cxx
+++ b/Root/Algorithm.cxx
@@ -134,6 +134,19 @@ bool xAH::Algorithm::isPHYS(){
     }
 }
 
+bool xAH::Algorithm::isRun3(){
+  /**
+   * Maybe there is a way to do this on-the-fly
+  */
+
+  if (m_isRun3 == -1) {
+    ANA_MSG_ERROR("isRun3 is can't be determined on-the-fly.  Please configure it.");
+    return 0;
+  } else {
+    return m_isRun3 == 1;
+  }
+}
+
 void xAH::Algorithm::registerInstance(){
     if(m_registered) return;
     m_instanceRegistry[m_className]++;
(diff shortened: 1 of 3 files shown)


Root/Algorithm.cxx
```diff
@@ -134,6 +134,19 @@ bool xAH::Algorithm::isPHYS(){
...
+   * Maybe there is a way to do this on-the-fly
+  */
+
+  if (m_isRun3 == -1) {
+    ANA_MSG_ERROR("isRun3 is can't be determined on-the-fly.  Please configure it.");
+    return 0;
+  } else {
+    return m_isRun3 == 1;
+  }
+}
```
jbossios, line 148: As it is, this will use Run 3 navigation even when ```m_isRun3 == 0```, which would be incorrect. I don't think there is a way to set this on the fly. So, I would make this completely configurable by the user.

I would change ```m_isRun3``` by ```m_useRun3navigation```, because there are Run 2 data and MC samples which were converted. Following my comment above, I would make it a boolean (since it seems it is not possible to know which navigation the file has when initializing the tools).

Similarly, I would also change ```isRun3()``` by ```useRun3navigation()``` which (following my comment above) will check ```m_useRun3navigation``` (which should be set to ```false``` or ```true```, to keep backwards compatibility we can make ```false``` the default).

Root/BasicEventSelection.cxx
```diff
@@ -521,6 +521,12 @@ EL::StatusCode BasicEventSelection :: initialize ()
     ANA_CHECK( m_trigDecTool_handle.setProperty( "ConfigTool", m_trigConfTool_handle ));
     ANA_CHECK( m_trigDecTool_handle.setProperty( "TrigDecisionKey", "xTrigDecision" ));
     ANA_CHECK( m_trigDecTool_handle.setProperty( "OutputLevel", msg().level() ));
+
+    if ( isRun3() )
+    {
+      ANA_CHECK( m_trigDecTool_handle.setProperty( "NavigationFormat", "TrigComposite") );
+      ANA_CHECK( m_trigDecTool_handle.setProperty( "HLTSummary", "HLTNav_Summary_DAODSlimmed") );
```
jbossios, line 528: Please add a new variable ```m_HLTSummary``` which default value can be ```"HLTNav_Summary_DAODSlimmed"```. We need this since for some samples we need to use ```"HLTNav_Summary_AODSlimmed"```.

Additionally, add please an else statement to set "NavigationFormat" to "TriggerElement". This is to follow the recommendation from this twiki: https://twiki.cern.ch/twiki/bin/viewauth/Atlas/R22TriggerAnalysis

jbossios (changes requested):
Thank you for implementing this. Please find some suggestions.

jbossios (approved):

PR #1620, opened 2023-02-24
Read LooseBL from Derivations

mdhank:
Read LooseBL from Derivations, as this seems to be available now and the alternative doesn't work with PHYSLITE

### DIFF
--- a/Root/ElectronSelector.cxx
+++ b/Root/ElectronSelector.cxx
@@ -280,12 +280,6 @@ EL::StatusCode ElectronSelector :: initialize ()
     }
 
     if ( m_readIDFlagsFromDerivation ) {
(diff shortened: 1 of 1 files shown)


kratsg:
Just to be sure, do we know when this got added to the derivation? Or was this during the time that the PHYSLITE format was still evolving in content?

mdhank:
I'm not sure when it changed, but the handling in xAODAnaHelpers is from 2016: https://github.com/UCATLAS/xAODAnaHelpers/pull/738

mdhank:
I'm not too familiar with the derivation setup, but I think it was added here: https://gitlab.cern.ch/atlas/athena/-/commit/6c65dbe53d3e9f01ab7c38b766e959c95be16ee0

PR #1623, opened 2023-02-28
Add extra AnalysisBase versions

mamerl:
Add additional AnalysisBase versions to the `ci.yml` configuration file to support AnalysisBase 22.2.108 and 22.2.109, which support the latest offline (`AntiKtEMPFlowJets`) jet calibration pre-recommendations.

Fixed previous yml formatting errors where tabs were used instead of spaces.


kratsg:
Hrmm (/cc @lukasheinrich @zlmarshall @matthewfeickert). Interesting that we don't have docker images for 108+ out still. I'll send an email on the mailing list.

Issue #1628, opened 2023-03-07
Explain how to change JVT tagger via the Python config


jbossios:
### Discussed in https://github.com/UCATLAS/xAODAnaHelpers/discussions/1626

<div type='discussions-op-text'>

<sup>Originally posted by **rbarrue** March  5, 2023</sup>
Currently, the C++ code in JetSelector uses the enum defined in the JVT namespace, but that enum can't be accessed in the Python config (only via the int directly). 

One way would be reimplement the JVT tagger enum in xAH, but that would require tracking the JVT enum and change in xAH anytime it changes in the ATLAS code.

Is there a smarter way to do this ?</div>

kratsg:
Is the `IJetJvtEfficiency` exposed at all through python (do we have a ROOT dict file generated for it for the python bindings)? If so, you could use the string names and use `getattr(CP.JvtTagger)` to get the integer value but...?

jbossios:
That might work. I will merge the above PR anyway since it works for sure but if @rbarrue wants to give that a try, then if it works I can add this to the docs/header.

PR #1629, opened 2023-03-07
Tell in JetSelector.h/docs how to change JVT alg

jbossios:
Implements: https://github.com/UCATLAS/xAODAnaHelpers/issues/1628

### DIFF
--- a/docs/Algorithms.rst
+++ b/docs/Algorithms.rst
@@ -94,10 +94,12 @@ with the decision of the cuts. A new xAOD::JetCollection can be made
 from the ones passing the cuts.
 
 The JVT selection is OFF by default, it can be turned ON with ```m_doJVT```.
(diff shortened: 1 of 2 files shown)


docs/Algorithms.rst
```diff
@@ -94,11 +94,13 @@ with the decision of the cuts. A new xAOD::JetCollection can be made
 from the ones passing the cuts.
 
 The JVT selection is OFF by default, it can be turned ON with ```m_doJVT```.
+The default algorithm is NNJvt. If another algorithm is needed, use corresponding index for the enum here: https://acode-browser1.usatlas.bnl.gov/lxr/source/athena/PhysicsAnalysis/Interfaces/JetAnalysisInterfaces/JetAnalysisInterfaces/IJetJvtEfficiency.h#0022.
```
kratsg, line 97: We might need to add details about making sure one gets the right version of the code corresponding to the release?

Issue #1631, opened 2023-03-10
"Could not find input string in enum!" and LooseBL updates


mdhank:
Dear Developers,

I've been seeing the line "Could not find input string in enum!" in my output and attempted to investigate.

I am using commit https://github.com/UCATLAS/xAODAnaHelpers/commit/ae64473c94194a705ea0cfd9eff7dd7641aaa95a of xAODAnaHelpers. I am running AnalysisBase 22.2.108 through a modified version of the BLstop framework (https://gitlab.cern.ch/atlassusy-bminusl/BLstopWorkspace). I am testing various samples using e8351_s3681_r13167_p5511 of DAOD_PHYS
. An example of my commands is

./source/xAODAnaHelpers/scripts/xAH_run.py -f --files ttbar_dilep_physlite_full.txt --nevents 10000 --config source/BLstopAnalysis/data/myconfig.py --inputRucio --inputList --isMC --submitDir tt_ll_test direct

I traced the output to https://github.com/UCATLAS/xAODAnaHelpers/blob/main/Root/ParticlePIDManager.cxx#L85 . This occurs because "VeryLooseNoPix" does not exist in https://gitlab.cern.ch/atlas/athena/-/blob/master/PhysicsAnalysis/ElectronPhotonID/ElectronPhotonSelectorTools/ElectronPhotonSelectorTools/LikelihoodEnums.h . The output no longer occurs if "VeryLooseNoPix" is removed from lines 17, 23, and 33. However, I don't understand the code well enough to know if this is desired- if "VeryLooseNoPix" can still be accessed in other ways, it might be better to manually add like in https://github.com/UCATLAS/xAODAnaHelpers/pull/1324.

While looking into this, it also made me realize https://github.com/UCATLAS/xAODAnaHelpers/pull/1620 isn't quite right. LooseBL is being overwritten on https://github.com/UCATLAS/xAODAnaHelpers/blob/main/Root/ParticlePIDManager.cxx#L26 and https://github.com/UCATLAS/xAODAnaHelpers/blob/main/Root/ParticlePIDManager.cxx#L44. If these are fixed and LooseBL is added to https://github.com/UCATLAS/xAODAnaHelpers/blob/main/Root/HelperClasses.cxx, it appears to work as desired. LooseBL was only added to LikelihoodEnums.h last July, not 3 years ago when it was added to the smart slimming list.

[...]

PR #1632, opened 2023-03-10
Remove VeryLooseNoPix and Fix LooseBL Electron WPs

mdhank:
Removing the VeryLooseNoPix electron WP as this no longer seems to be supported. Fixing hard-coded overwrites of LooseBL to Loose that were missed in https://github.com/UCATLAS/xAODAnaHelpers/pull/1620.


jbossios:
Solves: https://github.com/UCATLAS/xAODAnaHelpers/issues/1631

PR #1636, opened 2023-04-10
Adding AF3 support

hollypacey:
- Adds new function in Algorithm.cxx to tell from the metadata if isAF3() is true.
- AF3 will also set 'isFastSim()' to be true.
- "AF3" string will be set if isAF3() is true, rather than "AFII" or "FS" for the autoconfigured PRW filenames
- added variables to set AF3 in config akin to those for AFII in photon and tau calibrators. 
- At the moment assuming that any CP recommendations for AF3 are the same as AFII but this could be changed if needed in the future.
- Also added isAF3 override to python options, akin to current AFII override.

### DIFF
--- a/Root/Algorithm.cxx
+++ b/Root/Algorithm.cxx
@@ -7,6 +7,8 @@
 #include <xAODAnaHelpers/HelperFunctions.h>
 #include "xAODEventInfo/EventInfo.h"
 
+#include <boost/algorithm/string.hpp>
+
 std::map<std::string, int> xAH::Algorithm::m_instanceRegistry = {};
 
 // this is needed to distribute the algorithm to the workers
@@ -35,14 +37,13 @@ StatusCode xAH::Algorithm::algInitialize(){
 
 
     //Backwards compatibility
-    m_forceFastSim = m_forceFastSim || m_setAFII;
+    m_forceFastSim = m_forceFastSim || m_setAFII || m_setAF3;
 
     //Return a failure if there's contradictory flags (2 or more are true)
     if( m_forceData ? (m_forceFastSim || m_forceFullSim) : (m_forceFastSim && m_forceFullSim) ){
       ANA_MSG_ERROR("Multiple input-type flags are set, be sure only one of m_forceData(" << m_forceData << "), m_forceFastSim(" << m_forceFastSim << "), and m_forceFullSim(" << m_forceFullSim << ") are true.");
       return StatusCode::FAILURE;
     }
-
     return StatusCode::SUCCESS;
 }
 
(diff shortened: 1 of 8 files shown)


kratsg (approved):

mdhank:
Hi,

Should https://github.com/hollypacey/xAODAnaHelpers/blob/e373b601781f7a4cb4f50ca4d4fcd05a0ee4b19e/Root/Algorithm.cxx#L134 be `m_isAF3 = true;`?

Best,
Michael

kratsg:
> Should https://github.com/hollypacey/xAODAnaHelpers/blob/e373b601781f7a4cb4f50ca4d4fcd05a0ee4b19e/Root/Algorithm.cxx#L134 be `m_isAF3 = true;`?

good catch, yes. Can you submit a PR to quickly fix?

mdhank:
Unfortunately, it turns out the fix causes my tests to fail even earlier when setting AF3 (it fails either way). If I can figure out how to get it working, I'll submit a PR.

hollypacey:
Hi,

Yes well spotted, sorry about that! I should help you debug the errors since it was my MR, feel free to send me your error messages. My tests were working fine so I'm curious to see what else is wrong.

Cheers,
Holly

Dr Holly Pacey
ATLAS Group, Particle Physics
Schmidt AI in Science Fellow
ARF Reuben College, SCR Member Lady Margaret Hall.
University of Oxford
Skype: holly.pacey
(she/her)

Please do not feel obliged to reply to my emails outside of your working hours

mdhank:
Hi Holly,

Sorry for the late response. I suspect some of the problems might just be on my end. When I try adding 
` "m_setAF3"                : True,`
to my BasicEventSelection config, it doesn't seem to propagate through. If I add
`"m_isAF3"                 : 1,`
to the config and also --isAF3 to the command line, it uses AF3 but m_setAF3 becomes irrelevant. How do you configure to run this?

Even when this part works, though, it fails on the PhotonCalibrator with
` Property ForceDataType is set to 3, while it should be 1 for FullSim`

I haven't been able to find any R22 AF3 recommendations for the photon calibration. Did your test include the photon calibrator, and if so how did you handle this?

Thanks,
Michael

hollypacey:
Hi Michael, so for my tests on AF3 PHYS samples with p5556, I don't add any flags in my config, it automatically tells it's AF3 whenever isAF3() is called, if fastSim whenever isFastSim() is called (e.g in the PRW part of BasicEventSelection which is working fine). You're right that the algorithm and basicEventSel flags don't seem to be properly linked up but I just copied what is being done now for AFII. I didn't try running on the photons but I was assuming the for AF3, an AFII calibration would be more appropriate than a FullSim one so tried to make it do what is being done for AFII but I'll admit I didn't test the photonCalibrator.

From the error you're getting, it looks like the Photon tool in athena is being 'correctly' instructed to run a FastSim calibration, but there just isn't one available in R22 to set an appropriate Map file so it's complaining. I think it should also be complaining if it's AFII in R22 too based on the twiki though https://twiki.cern.ch/twiki/bin/view/AtlasProtected/PhotonEfficiencyRun2#Latest_Run_3_recommendations_wit

If I instead make photonCalibrator run as FullSim I get this error instead:

"Package.EventLoop        ERROR   /build1/atnight/localbuilds/nightlies/AnalysisBase/master/athena/PhysicsAnalysis/D3PDTools/EventLoop/Root/MessageCheck.cxx:37 (void EL::Detail::report_exception(std::__exception_ptr::exception_ptr)): caught exception: SG::ExcBadAuxVar: Attempt to retrieve nonexistent aux data item `::Rhad1' (832).
Package.EventLoop        ERROR   /build1/atnight/localbuilds/nightlies/AnalysisBase/master/athena/PhysicsAnalysis/D3PDTools/EventLoop/Root/Worker.cxx:669 (StatusCode EL::Worker::algsExecute()): while calling execute() on algorithm PhotonCalibrator
"

Have you got PhotonCalibrator working in R22 for any scenario yet?

mdhank:
Hi Holly,

Ah, I understand now why I saw issues. I'd been using the command line/config options because I'm running on the latest DAOD_PHYS production, p5631, which I believe has a bug where it's unfortunately unable to detect the simulation type automatically.

The PhotonCalibrator works in R22 when I run on FullSim files, but I haven't got it to work on AF3 files under any configuration. I get similar errors as you when I try to run it using the FastSim/FullSim configs. It might just not be possible to do AF3 photon calibrations yet.

Best,
Michael

mdhank:
Hi,

I just wanted to add for future reference that the PhotonCalibrator does work, if you use the Run3 pre-recommendations as for full-sim analysis and `m_readIDFlagsFromDerivation`.

Best,
Michael

PR #1640, opened 2023-05-23
Add back support for VeryLooseNoPix WP for electrons

gfrattar:
hi,

after the update in athena of the WP for electrons for LLPs searches (https://gitlab.cern.ch/atlas/athena/-/merge_requests/61580) we added back the support in xAH for the very loose no pixel ID WP used by the dHNL r22 analysis.

It works from release 24.2.1 onwards, i.e. from when the aforementioned MR got to an official release.

Kind regards,
Sagar & Guglielmo


kratsg:
Sorry, that was backported to R24 only and not R22? You'll need to upgrade the `.github/workflows/` file to specify the releases that should be tested.. As you can see, it's failing R22 right now.

gfrattar:
hi,

this was included in r22 by us https://github.com/UCATLAS/xAODAnaHelpers/pull/1597 and then removed for some reason later on in https://github.com/UCATLAS/xAODAnaHelpers/pull/1632 without even contacting us.

I will try to update the workflow as you suggested.

gfrattar:
hi @kratsg , the CI fails because of some error in the make file used in the test job. Do you know how to solve it? thanks

kratsg:
Sorry, I'm very confused now. xAH isn't for R24 but for R22. Are you saying we need to use 24.X from now on? The CMakeLists is generated from the top-level one in the `ci/` folder.

gfrattar:
hi @kratsg ,

my understanding is that 22.2.X release are not built anymore, as you can also see from here: https://twiki.cern.ch/twiki/bin/view/AtlasProtected/AnalysisBaseReleaseNotes22pt2 so one wants to do a "rel 22 analysis" with latest recommendations and so on one should move to the latest analysisBase release, which will be 24.2.X.

It is also true that from https://twiki.cern.ch/twiki/bin/view/AtlasProtected/AnalysisRelease one reads "24.2.X" is analysis release for rel23/24.

I am not an expert of software releases, but for example some recommendations for rel22 analyses are only avaialble from 24.2.0 (e.g. MET TST uncertainties - https://twiki.cern.ch/twiki/bin/view/AtlasProtected/JetEtmissRecommendationsR22#Missing_Transverse_Momentum_MET) which to me sounds like if you setup the latest analysisBase release you will get them. Given this, I would tend to say that in the end 24 is "just a number". 

If anybody has a clearer view of what's happening with the versioning, I am happy to learn it. If the maintainers want to keep xAH safe in 22.2.X then I would suggest to create a 24.2.X branch.

Cheers,
Guglielmo

jburzy01:
@gfrattar is correct -- the last `22.2.X` release is two months old. The latest analysis releases have been moved to `24.2.X`.

kratsg:
> @gfrattar is correct -- the last `22.2.X` release is two months old. The latest analysis releases have been moved to `24.2.X`.

yeah, I haven't been following the analysis releases here (this is what Jona used to do and we're still looking for a new maintainer for R22 xAH).

gfrattar:
hi @kratsg , any news on this PR?

PR #1642, opened 2023-06-01
Addition of support for close-by-correction to isolation variables

gfrattar:
hi,

this pull request is to add the support in xAH to the IsolationCloseByCorrectionTool that is used to correct the value of isolation variables evaluated at derivation level to take into account the possibility that close by leptons contribute to each other isolation variable value. This is a wrapper of the tool available in Athena - see https://gitlab.cern.ch/atlas/athena/-/blob/master/PhysicsAnalysis/AnalysisCommon/IsolationSelection/Root/IsolationCloseByCorrectionTool.cxx?ref_type=tags . 

Cheers,
Guglielmo & Sagar

### DIFF
--- a/Root/IsoCloseByCorr.cxx
+++ b/Root/IsoCloseByCorr.cxx
@@ -1,0 +1,209 @@
+// c++ include(s):
+#include <iostream>
+#include <sstream>
+
+// EL include(s):
+#include <EventLoop/Job.h>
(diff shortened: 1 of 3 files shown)


kratsg:
Can you provide a link to the twiki in the comments so it gets rendered as documentation for people to find what working points they can use, etc?

gfrattar:
done

PR #1652, opened 2023-08-03
Add the support for a lepton favored OR

SagarA17:
This MR adds the support for lepton favored overlap removal configuration. Within this, the electron-electron OR is enabled and uses cluster matching, and the leptons which overlap with jets are not removed (which is done by setting the outer cone DR to 0). This OR was developed and supported by HSG2.

Tagging @gfrattar.

Cheers,
Sagar


SagarA17:
Hi, could we request someone to check this MR? Maybe @kratsg?

Thanks,
Sagar

PR #1659, opened 2023-10-05
Update to support newer releases and fix isolation variables

SagarA17:
This MR aims to fix a few tool calls to support newer releases. The main changes however are in the lepton isolation variables - 
1. The newer PHYS and LLP1 derivations now contain the closeByCorrected variables already written, which should then be used to calculate the isolation working points. There are corresponding changes to support this.
2. Older isolation variables are removed which might not even be written to the derivations anymore.

We also made some safeFill changes to track parameters in case they are not available in the derivations (which is often).

Thanks,
Sagar and @gfrattar


SagarA17:
Hello,

Thanks for checking! I did not make any changes to the JetSelector, so I
would think it’s an independent error not induced by us.

Thanks,
Sagar

mdhank:
Hello,

I disabled all the usage of JVT, and was able to run, though I still get
the warning:
> METConstructor.METMake...WARNING No JetContainer set. This behaviour is
deprecated

I agree this error is not from your changes, but fixing it may be necessary
to fully support newer releases. I will continue looking into it.

It's possible this is related to the rewrite of the JVT in
https://gitlab.cern.ch/atlas/athena/-/merge_requests/64846 , which was
implemented in 24.2.19.

Best,
Michael

SagarA17:
Hello, @miholzbo. Thanks for the pointer! I have added the last commit to this PR as well. Hopefully, it solves the error/warning that @mdhank sees.

SagarA17:
Thanks for checking @mdhank! Let me know what else I should look at before you are ready to merge.

mdhank:
Hi @SagarA17 , I don't have merge permissions for xAH.

SagarA17:
Maybe @kratsg could help us merge? We would appreciate the help!

kratsg (approved):

PR #1660, opened 2023-10-11
Fix LRT muon selection and PRW

SagarA17:
Hi,
We make improvements to MuonSelector to act correctly on LRT muons. Earlier, they were just being ignored and the logic was convoluted and dirty.
After an update to PathResolver, there are problems looking for files if the path has a leading forward slash. Fixing it to get the PRW file automatically. This change was reverted in AB,24.2.26, but I still suggest removing the leading slash to not make it look like its an absolute path.
Tagging @kratsg to merge. Thanks!

### DIFF
--- a/Root/MuonSelector.cxx
+++ b/Root/MuonSelector.cxx
@@ -798,7 +798,8 @@ int MuonSelector :: passCuts( const xAOD::Muon* muon, const xAOD::Vertex *primar
 
   bool LRTdecorIsAvailable = false;
   bool muonIsLRT = false;
-  bool AcceptPromptMuon = (bool)m_muonSelectionTool_handle->accept( *muon );
+  bool AcceptStdMuon = (bool)m_muonSelectionTool_handle->accept( *muon );
+  bool AcceptLRTMuon = this_quality <= m_muonQuality; //LRT WP do not have the ID cuts applied so use getQuality()
 
   if ( m_doLRT ){
     static SG::AuxElement::Decorator< char > passIDcuts("passIDcuts");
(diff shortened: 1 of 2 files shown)


Root/MuonSelector.cxx
```diff
@@ -810,13 +811,16 @@ int MuonSelector :: passCuts( const xAOD::Muon* muon, const xAOD::Vertex *primar
...
       ANA_MSG_DEBUG( "Muon failed requirements of MuonSelectionTool.");
       return 0;
     }
   }
   else {
-    if ( (!LRTdecorIsAvailable && !AcceptPromptMuon) || (LRTdecorIsAvailable && !muonIsLRT && !AcceptPromptMuon) ) {
+    if ( (!LRTdecorIsAvailable && !AcceptStdMuon) || //If decor is not available, the muon is standard
+         (LRTdecorIsAvailable && !muonIsLRT && !AcceptStdMuon) || //If decor is available and is 0, the muon is standard. Use accept() method for standard muons.
+         (LRTdecorIsAvailable && muonIsLRT && !AcceptLRTMuon) //If decor is available and is 1, the muon is LRT. Use getQuality() method for LRT muons.
+       ) {
```
goblirsc, line 823:
```suggestion
    bool isLRT = LRTdecorIsAvailable && muonIsLRT; 
    if ( !(
               (isLRT && AcceptLRTMuon) || (!iLRT && AcceptStdMuon)
          )
      ) {
```

SagarA17:
Hi @kratsg. This is a reminder for this MR.

PR #1666, opened 2023-12-19
Adding common PRW and mc23 PRW

mdhank:
Adding mc23a and mc23c cases for PRW auto-configuration. Also adding the option to use common PRW files for each MC campaign instead of DSID-specific files.


Root/BasicEventSelection.cxx
```diff
@@ -1228,26 +1234,45 @@ StatusCode BasicEventSelection::autoconfigurePileupRWTool()
...
+	    prwConfigFiles.push_back(PathResolverFindCalibFile(m_prwActualMu2016File));
       if( !m_prwActualMu2017File.empty() && (mcCampaign == "mc20c" || mcCampaign=="mc20d") )
-	prwConfigFiles.push_back(PathResolverFindCalibFile(m_prwActualMu2017File));
+	    prwConfigFiles.push_back(PathResolverFindCalibFile(m_prwActualMu2017File));
       if( !m_prwActualMu2018File.empty() && (mcCampaign == "mc20e" || mcCampaign=="mc20f") )
-	prwConfigFiles.push_back(PathResolverFindCalibFile(m_prwActualMu2018File));
+	    prwConfigFiles.push_back(PathResolverFindCalibFile(m_prwActualMu2018File));
+      if( !m_prwActualMu2022File.empty() && mcCampaign == "mc23a" )
+        prwConfigFiles.push_back(PathResolverFindCalibFile(m_prwActualMu2022File));
+      if( !m_prwActualMu2023File.empty() && mcCampaign == "mc23c" || mcCampaign == "mc23d" )
```
miholzbo, line 1274: I think this needs to be changed to `(mcCampaign == "mc23c" || mcCampaign == "mc23d")` otherwise the code crashed it the user doesn't specify a actual PRW file for 2023

mdhank: Thanks for catching this! I've fixed this, and allowed metadata to detect the campaign directly (note this should give a warning but still work if you use an out-of-date PRW such as mc23c).

miholzbo: If I understand correctly, you can use the autoconfiguration of PRW files for mc23c if you manually specify the campaign as property via `m_mcCampaign`, otherwise you will run over mc23c but apply the PRW as it would be a mc23 file?

mdhank: My apologies, some of my commits didn't make it in - I've added them now.

PR #1670, opened 2024-01-15
Initialisation of down variation as well when running over limited list of systematics

gfrattar:
As of the title, this PR fixes the fact that when running over a given set of systematic uncertainties in the config, only the up variation was initialised. Now also the down variation is correctly initialised.


mdhank:
This looks straightforward to me! I'll merge this.

PR #1673, opened 2024-02-07
Update fast sim flag in EgammaCalibrationAndSmearingTool

SagarA17:
This MR updates a flag which is need to support newer AB releases, dictated by this change - https://gitlab.cern.ch/atlas/athena/-/merge_requests/67929#e97c3ea34df8cbbc908de52615597f69841c9960

It's a straightforward technical change with no functional differences.


mdhank:
Hi @SagarA17 ,

Should this also be changed in https://github.com/UCATLAS/xAODAnaHelpers/blob/main/Root/PhotonCalibrator.cxx#L160 ?

Best,
Michael

SagarA17:
Hi @mdhank,

Good catch! I have changed the second instance as well.

Sagar

mdhank:
Looks good, merging. One thing to watch out for with these releases- mc20e muon trigger scale factors are broken for AnalysisBase 24.2.29-24.2.39, and mc20a muon trigger scale factors are broken for 24.2.29+.

PR #1676, opened 2024-02-16
xAH updates for Trigger-Level Analyses

mamerl:
Adds various modifications including:
- changes to `JetCalibrator`  for correcting HLT jets (new options for setting up `JetCalibrationTool`, yet to be upstreamed in Athena but planned)
- more functionality for filling `TProfile` & `TH3` objects with `HistogramManager`
- changes to handling of photon containers in `OverlapRemover`, only for checking existence in TStore when deep/shallow copies exist
- changes to `BasicEventSelection` adding a TLA data flag to control how trigger decisions are checked (from some differences between trigger decisions stored in 2022 TLA data & the primary physics data stream)

NOTE: for `BasicEventSelection` trigger decision changes we may need a follow up PR after checking if the modified TLA data logic is required for 2023 TLA data (if not, then we will need a 2022 TLA data flag)

### DIFF
--- a/Root/BasicEventSelection.cxx
+++ b/Root/BasicEventSelection.cxx
@@ -962,10 +962,10 @@ EL::StatusCode BasicEventSelection :: execute ()
 
     if ( m_applyTriggerCut ) {
 
-      // additional DEBUG logging to validate conditional logicss
+      // additional DEBUG logging to validate conditional logic
       ANA_MSG_DEBUG("Applying trigger cut corresponding to chain group " << m_triggerSelection);
-      ANA_MSG_DEBUG("Trigger chain group is NOT passed && is TLA data = " << int(!triggerChainGroup->isPassed(TrigDefs::requireDecision) && m_isTLAData));
-      ANA_MSG_DEBUG("Trigger chain group is NOT passed (no requireDecision) && is NOT TLA data = " << int(!triggerChainGroup->isPassed() && !m_isTLAData));
+      ANA_MSG_DEBUG("Is Trigger-Level Analysis (TLA) data = " << int(m_isTLAData));
+      ANA_MSG_DEBUG("Trigger chain group is passed = " << int(m_isTLAData ? triggerChainGroup->isPassed(TrigDefs::requireDecision) : triggerChainGroup->isPassed()));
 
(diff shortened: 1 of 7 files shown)


tofitsch:
To add some context: These changes are needed in the dijet+ISR Trigger Level Analysis (TLA) which makes heavy use of xAH. They will also be necessary for other later trigger level analyses.

# JetCalibrator
Looks fine: While these new features are only needed for TLA, they are activated by the new option `m_recalibrateHLTJets` which is `false` by default. Therefore doesn't affect anyone that doesn't explicitly activate it.
# BasicEventSelection
 Similarly, everything new is guarded behind the new `m_isTLAData` option which is also false by default. However, I would change these debug messages to not check for an `&&` of 2 things at the same time and to remove the `NOT` which is confusing:
```
      ANA_MSG_DEBUG("Trigger chain group is NOT passed && is TLA data = " << int(!triggerChainGroup->isPassed(TrigDefs::requireDecision) && m_isTLAData));
      ANA_MSG_DEBUG("Trigger chain group is NOT passed (no requireDecision) && is NOT TLA data = " << int(!triggerChainGroup->isPassed() && !m_isTLAData));
```
to something like this:

```
      ANA_MSG_DEBUG("Is TLA data = " << int(m_isTLAData));
      ANA_MSG_DEBUG("Trigger chain group is passed = " << int(m_isTLAData ? triggerChainGroup->isPassed(TrigDefs::requireDecision) : triggerChainGroup->isPassed());
```
It's probably better to change `Is TLA data` into `Is Trigger Level Analysis (TLA) data` since not everyone can be expected to know this acronym.

# HistogramManager
Also fine: Just adding a few more overloaded options of functions that exist for TH2 now for TH3

# OverlapRemover
 Looks a bit strange only doing it for jets and photons. Should we not do the same (i.e. adding the conditional without `DataVector`) for e, mu, tau too then to be consistent?

mamerl:
Updates:

`BasicEventSelection`: changed logging statements as suggested and fixed indentation of a code block

`OverlapRemover`: made checking of ConstDataVector<xAOD container> & xAOD container presence in TStore consistent for all objects in overlap removal

PR #1681, opened 2024-03-26
Fix incorrect mc20d PRW filepath

balunas:
The hard-coded filepath for the common pileup reweighting config file for mc20d is incorrect in `BasicEventSelection`. This fixes it.


mdhank:
Thanks for catching this!

PR #1683, opened 2024-03-26
Configure GSCDepth option for JetCalibrator

mamerl:
Adding an additional option in `JetCalibrator` to allow the `GSCDepth` specified in a calibration config file to be overwritten with a user-specified value via `m_calibGSCDepth`. Implemented to only override `GSCDepth` when `m_calibGSCDepth` is not empty and `GSC` exists as a substring in the calibration sequence string.

### DIFF
--- a/Root/JetCalibrator.cxx
+++ b/Root/JetCalibrator.cxx
@@ -227,6 +227,10 @@ EL::StatusCode JetCalibrator :: initialize ()
   if ( m_jetCalibToolsDEV ) {
     ANA_CHECK( m_JetCalibrationTool_handle.setProperty("DEVmode", m_jetCalibToolsDEV));
   }
+  if (!m_calibGSCDepth.empty() && m_calibSequence.find("GSC") != std::string::npos) {
+    ANA_MSG_WARNING("Using modified GSCDepth property for jet calibration '" << m_calibGSCDepth << "' which will override config file value");
+    ANA_CHECK( m_JetCalibrationTool_handle.setProperty("GSCDepth", m_calibGSCDepth));
+  }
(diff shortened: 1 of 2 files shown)


mdhank:
Hi @mamerl ,

For my knowledge, what does the `GSCDepth` option do? And what sort of value(s) are allowed?

Thanks,
Michael

mamerl:
Hi @mdhank,

`GSCDepth` controls which of the GSC steps get applied e.g. normally the ordering is `chargedFraction->Tile0->EM3->nTrk->trackWIDTH->PunchThrough` but for HLT jets we want to only do up to `trackWIDTH` since the information for the `PunchThrough` correction is not available for HLT jets.

The allowed values of `GSCDepth` are strings that can be any of "chargedFraction", "Tile0", "EM3", "nTrk", "trackWIDTH", "PunchThrough", or "Full" (same as "PunchThrough").

In Athena there is a check for the `GSCDepth` parameter to ensure it is valid at [https://gitlab.cern.ch/atlas/athena/-/blob/main/Reconstruction/Jet/JetCalibTools/Root/GlobalSequentialCorrection.cxx?ref_type=heads#L109](https://gitlab.cern.ch/atlas/athena/-/blob/main/Reconstruction/Jet/JetCalibTools/Root/GlobalSequentialCorrection.cxx?ref_type=heads#L109). For now we just implemented the flag to pass the value to JetCalibrationTool and let the Athena tools handle the checks. If needed we can add some error checking though. What do you think is best?

Thanks,
Max

mdhank:
Thanks for the information! I think it's fine as is.

PR #1686, opened 2024-04-24
Fixing FTAG MCMC SFs

mdhank:
Updating MC-MC SFs to match https://ftag.docs.cern.ch/algorithms/activities/mcmc/#release-22-supported-generators and the CDI files (with the CDI files given precedence). Note this requires a new option in BJetEfficiencyCorrector, isRun3, which is set to False by default.

### DIFF
--- a/Root/HelperFunctions.cxx
+++ b/Root/HelperFunctions.cxx
@@ -538,7 +538,7 @@ HelperFunctions::ShowerType HelperFunctions::getMCShowerType(const std::string& 
   else if(tmp_name.Contains("AMCATNLOH")) return AmcH7;
   else if(tmp_name.Contains("PYTHIA8EVTGEN")) return Pythia8;
   else if(tmp_name.Contains("HERWIG")) return Herwig7p1;
-  else if(tmp_name.Contains("PHH7EG_H7UE")) return Herwig7p2;
+  else if(tmp_name.Contains("PHH7EG")) return Herwig7p2;
   else if(tmp_name.Contains("SHERPA_221_")) return Sherpa221;
   else if(tmp_name.Contains("SH_221_")) return Sherpa221;
   else if(tmp_name.Contains("SH_2210")) return Sherpa2210;
--- a/Root/HelperFunctions.cxx
+++ b/Root/HelperFunctions.cxx
@@ -534,14 +534,15 @@ HelperFunctions::ShowerType HelperFunctions::getMCShowerType(const std::string& 
 
   //
   // Determine shower type by looking for keywords in name
-  if(tmp_name.Contains("PYTHIA8EVTGEN")) return Pythia8;
+  if(tmp_name.Contains("AMCATNLOPY")) return AmcPy8;
+  else if(tmp_name.Contains("AMCATNLOH")) return AmcH7;
+  else if(tmp_name.Contains("PYTHIA8EVTGEN")) return Pythia8;
   else if(tmp_name.Contains("HERWIG")) return Herwig7p1;
-  else if(tmp_name.Contains("PhH7EG_H7UE")) return Herwig7p2;
+  else if(tmp_name.Contains("PHH7EG_H7UE")) return Herwig7p2;
   else if(tmp_name.Contains("SHERPA_221_")) return Sherpa221;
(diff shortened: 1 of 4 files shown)


mdhank:
Dear @tofitsch , 

Could you take a quick look at this? I've tested that it behaves as I expect, but it's good to have a reviewer separate from the author.

Thanks,
Michael

mdhank:
> Looks generally good to me. I trust you with all the numbers and will not check them :D only worry I have is regarding backwards compatibility. E.g. before any file sample name containing `H7` got assigned `Herwig7`, now only if it contains `PHH7EG` it is `Herwig7p2` but if it contains `H7` but not `PHH7EG` it will be assigned `Unknown`. Not sure if that's ok? Are there no possible sample names that contain `H7` but not `PHH7EG`?

Good idea, I'll check for h7s. There are some such as
`mc23_13p6TeV.525891.MGH7EG_LO_lj_tchannel_el_150to200.deriv.DAOD_PHYS.e8544_s4162_r14622_p5855`,
`mc23_13p6TeV.522027.aMCH7EG_NNPDF30NLO_H723UE_ttmumu_run3.deriv.DAOD_PHYS.e8558_a910_r14932_p6026`, 
`mc20_13TeV.346526.PowhegHerwig7EvtGen_H7UE_NNPDF30ME_ttH125_gamgam.deriv.DAOD_PHYS.e7488_s3681_r13167_p5631`, or
`mc23_13p6TeV.830148.H7EG_H72NNPDF30NLO_jetjet_Cluster_JZ6.deriv.DAOD_PHYS.e8551_s4159_r14799_p5855`.

The amc definitely shouldn't fall under this case as it has it's own MC-MC SFs, and I don't think MG should either as the MC-MC SFs (https://ftag.docs.cern.ch/algorithms/activities/mcmc/#run-2) specify they are for Powheg+Herwig7. I believe the one with PowhegHerwig in the name should fall under Herwig7.1. I'm less clear on the dijet samples.
There are also datasets like
`mc23_13p6TeV.902037.QBHPy8EG_QBH_photonjet_n6_Mth7000.deriv.DAOD_PHYS.e8557_s4162_r14622_p6026` and
`mc23_13p6TeV.603317.PhPy8EG_ggH700W20_tautaulh.deriv.DAOD_PHYS.e8566_s4162_r14622_p6026`
which have h7 but should definitely not be included. So I'd lean towards leaving as is to avoid false postives.

It could be worth adding a warning in general that it's good to double-check what values the parsing uses, as it's very easy to miss edge cases.

Root/HelperFunctions.cxx
```diff
@@ -527,21 +527,22 @@ HelperFunctions::ShowerType HelperFunctions::getMCShowerType(const std::string&
...
+  else if(tmp_name.Contains("AMCATNLOH")) return AmcH7;
+  else if(tmp_name.Contains("PYTHIA8EVTGEN")) return Pythia8;
+  else if(tmp_name.Contains("HERWIG")) return Herwig7p1;
+  else if(tmp_name.Contains("PHH7EG")) return Herwig7p2;
+  else if(tmp_name.Contains("SHERPA_221_")) return Sherpa221;
+  else if(tmp_name.Contains("SH_221_")) return Sherpa221;
   else if(tmp_name.Contains("SH_2210")) return Sherpa2210;
   else if(tmp_name.Contains("SH_2211")) return Sherpa2210;
-  else if(tmp_name.Contains("SH_2212")) return Sherpa2210;
-  else if(tmp_name.Contains("SH_2214")) return Sherpa2210;
```
dbuchin, line 544: Hi, I realise I'm a little late now, but would it be possible to declare SH_2214 also Sherpa2212 so that the BJetEffCorr doesn't crash on the Run3 V+jets samples?
I can start a new pull request for this line if needed

mdhank: Hi @dbuchin ,

You should be able to manually set the calibration by setting m_EfficiencyCalibration to the desired value in your configuration ("700660" to use the Sherpa 2.2.12 SFs). Would this work for you? If possible, I would prefer not to automatically declare any unless they are officially supported (see https://ftag.docs.cern.ch/algorithms/activities/mcmc/#release-22-supported-generators for the official list).

PR #1689, opened 2024-05-17
Option to fallback to default calibration in automatic configuration of MC-to-MC b-tagging SFs

miholzbo:
The "auto" option to configure the b-tagging SFs currently causes failures if a sample isn't covered by the list. This PR gives an option to make this a bit more forgiven by potentially falling back to the default calibration. The default is of course to not use this.

### DIFF
--- a/Root/BJetEfficiencyCorrector.cxx
+++ b/Root/BJetEfficiencyCorrector.cxx
@@ -247,9 +247,16 @@ EL::StatusCode BJetEfficiencyCorrector :: initialize ()
                     calibration="700660";
                     break;
                 default:
-                    ANA_MSG_ERROR("Cannot determine MC shower type for sample " << gridName << ".");
-                    return EL::StatusCode::FAILURE;
-                    break;
+                    if (m_allowCalibrationFallback) {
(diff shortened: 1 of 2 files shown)


mdhank:
Hi @miholzbo ,

You should be able to manually set the MC-MC configuration if you don't use the "auto" option. Would this work for you? I can merge this if necessary, but I think it would be safer if the auto option only works with recommended samples. That way, if it fails, users can decide on a case-by-case basis what is appropriate (for example, if you are using a sample with an unsupported Sherpa version, you may decide to use the closest Sherpa rather than the default, which I believe is Pythia).

miholzbo:
From the point of view of our analysis, we'd highly prefer having this merged. When submitting to the grid, we don't think we can properly access the name of the sample name  in the xAH config file to process and deduce what calibration to apply (which would be mostly anyway the default one if not dedicated is available).

So this would quite strongly reduce the amount of bookeeping on our part, as otherwise we have to submit some samples with dedicated commandline arguments or dedicated configs.

miholzbo:
That's a fair suggestion; I've added another line of warning when using the fallback option to clarify this.

PR #1700, opened 2024-08-05
Add distinctions for muon trigger SFs for 2023 and 2024

miholzbo:
This pull request adds some distinctions for 2023 and 2024 when evaluating muon triggers SFs. The logic in this block is a bit un-intuitive to me, so a cross check is appreciated.

To my understanding these distinctions are not required but mostly avoid to create unnecessary decorators.

I've slightly changes some of the existing run numbers to be in line with
https://gitlab.cern.ch/atlas/athena/-/blob/main/PhysicsAnalysis/MuonID/MuonIDAnalysis/MuonEfficiencyCorrections/Root/MuonTriggerScaleFactors.cxx
where the values for 2023 originate from this ongoing MR
https://gitlab.cern.ch/atlas/athena/-/merge_requests/73513/diffs


mdhank:
Hi @miholzbo ,

Thanks for the PR! The logic looks right to me.

Best,
Michael

PR #1702, opened 2024-08-13
Move to new set of refactored Jvt-related tools

miholzbo:
From the jet pile-up tagging side the old and overloaded `JetJvtEfficiency` tool has been deprecated now since nearly a year in favor of a new set of refactored tools (where each does more or less only one things and are hence simpler to maintain).

This PR now finally also makes this migration within xAH.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,7 @@ jobs:
         release_type:
           - analysisbase
         release_version:
-          - 25.2.17
-          - 25.2.18
-          - 25.2.19
-          - 25.2.20
-          - 25.2.21
+          - 25.2.24
 
     steps:
     - uses: actions/checkout@master
--- a/Root/JetSelector.cxx
+++ b/Root/JetSelector.cxx
@@ -498,22 +498,6 @@ EL::StatusCode JetSelector :: execute ()
     // this will be the collection processed - no matter what!!
     ANA_CHECK( HelperFunctions::retrieve(inJets, m_inContainerName, m_event, m_store, msg()) );
 
-    // decorate inJets with truth info
-    if ( isMC() && m_doJVT && m_haveTruthJets ) {
-      static SG::AuxElement::Decorator<char>  isHS("isJvtHS");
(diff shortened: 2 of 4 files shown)


mdhank:
Hi @miholzbo ,

Thanks for making this PR! I have just a couple questions/comments before merging.

1. There's a few times you use `#ifndef XAOD_STANDALONE`. I think this should work, but I don't think xAH usually makes this distinction (as far as I'm aware, it's only run in AnalysisBase, though I could be wrong).

2. For my own understanding, what is the purpose of `m_doPileuplabel`? When is this needed?

Thanks,
Michael

mdhank:
Hi @miholzbo ,

The commit "remove manual evaluation of isJvtHS in xAH, now done by JetPileupLabelingTool" does slightly change the JVT and fJVT scale factors in my test. Is this expected behavior? It might be that the new behavior is more correct, but I wanted to verify this.

Best,
Michael

miholzbo:
Turns out there is a bug in the `JetPileupLabelingTool` which should do the jet truth matching in the first place now. This will be fixed in athena via this MR here:
https://gitlab.cern.ch/atlas/athena/-/merge_requests/73729

After this fix, I get now the same Jvt weights as before. As the `isJvtHS` decorations are slightly buggy in the existing decorations, I'll changed to default to actually run this tool now (as in new AB releases it will be correct).

mdhank:
HI @miholzbo ,

I checked with the AnalysisBase 25.2.24 and now see agreement. There's two small changes that I think would be good before merging this:

1. Could you change the supported releases in `.github/workflows/ci.yml` to just AnalysisBase 25.2.24? (If I understand this PR correctly, it will make prior AB incorrect).

2. In `Root/TauCalibrator.cxx` line 134, could you change `ApplyMVATESQualityCheck` to `MVATESQualityCheck` due to https://gitlab.cern.ch/atlas/athena/-/merge_requests/73708? As this PR will make xAH incompatible with AB<25.2.24, it would be good to add this now as otherwise xAH will crash if taus are used.

Thanks,
Michael

miholzbo:
Hi @mdhank,

No problem, I've included the changes you suggested.

Best,
Michael

PR #1704, opened 2024-08-19
Update METConstructor.cxx adding object resolution

alplebani:
I'm adding the MET resolution branches to the Ntuples, needed to be compatible with the latest METPerformance version


hollypacey:
Just to clarify, this is just adding decorators to the MET significance object with the resolution terms, it shouldn't add anything to poeple's NTuples generally, it just means we can access these in our METPerformance code to put into our NTuples. This is needed for METNetSignificance development.

mdhank:
Looks straightforward, merged!

PR #1706, opened 2024-10-29
Draft: Introduce option to run AsgDeadHVCellRemovalTool for electrons

miholzbo:
Following the egamma twiki here, we should to an additional veto for electrons in data 2016:
https://twiki.cern.ch/twiki/bin/view/AtlasProtected/EGammaIdentificationRun2#Removal_of_Electron_Photon_clust

It's also done in the CP algs here
https://gitlab.cern.ch/atlas/athena/-/blob/main/PhysicsAnalysis/Algorithms/EgammaAnalysisAlgorithms/Root/EgammaIsGoodOQSelectionTool.cxx
so I more or less just copied the implementation.

The only question I'd have is if we even may want to run this veto as default because it seems to be recommended?

### DIFF
--- a/Root/ElectronSelector.cxx
+++ b/Root/ElectronSelector.cxx
@@ -151,6 +151,7 @@ EL::StatusCode ElectronSelector :: initialize ()
     m_el_cutflow_all             = m_el_cutflowHist_1->GetXaxis()->FindBin("all");
     m_el_cutflow_author_cut      = m_el_cutflowHist_1->GetXaxis()->FindBin("author_cut");
     m_el_cutflow_OQ_cut          = m_el_cutflowHist_1->GetXaxis()->FindBin("OQ_cut");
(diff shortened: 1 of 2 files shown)


mdhank:
Hi @miholzbo ,

Thanks for the PR! I think if it's recommended then it should probably be applied by default (unless there's some common use-case where it shouldn't be applied).

Thanks,
Michael

miholzbo:
Thanks for the quick review. I've processed a data16 file locally and 68552 from 68714 electrons (around 0.2%) didn't pass the veto. That's of course just picking one file, no idea how large the fraction over the full data16 is.

Given that observation, I didn't set the option to true for the moment but just print a warning if it isn't true which seems to be a compromise to let users know something should be double-checked. But setting it per default to true is of course a trivial change. Let me know if you prefer that.

mdhank:
Looks good to me! I looked and we don't apply similar flags (such as event and jet cleaning) by default, so I think this is the best solution.

PR #1717, opened 2025-02-17
Adding functionality for AnaReentrantAlgorithms

mdhank:
Adding functionality for AnaReentrantAlgorithms using PythonConfig


mdhank:
Hi @SagarA17 ,

Could you test if this works for you with AB 24.2.20? I don't have a full AnaRentrantAlgorithms setup so can't test this easily, though I've managed to make my test branch crash in the same way as in 24.2.39 (before these changes it would crash differently).

Best,
Michael

SagarA17:
Hi @mdhank. Thanks a lot for this change! I was able to run my tests with these changes and they worked successfully. Appreciate the prompt help! Might I suggest adding AB versions in CI up to `24.2.43` to ensure that the up to date compatibility is accurately reflected? Thanks again!

SagarA17:
Btw one small suggestion from my side would be to look into the `AnaAlgorithm.DualUseConfig` as was suggested in the MR https://gitlab.cern.ch/atlas/athena/-/merge_requests/77352#note_9056993, since they seem to hint that the other methods will be obsoleted in due time, but perhaps this is not urgent.

